PR #16, opened 2020-04-25
WIP: Add Docker support

### LINKED ISSUE #13: Docker deployment
It would be nice to have docker deployment which is now top solution for multi-platform managing apps on server or localhost in quick time. Do you agree? Can I send pull request with implementation?

Skn0tt:
Closes #13.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,44 @@
+.git
+
+# Compiled class file
+*.class
+
+# Log file
+*.log
+
+# BlueJ files
+*.ctxt
+
+# Mobile Tools for Java (J2ME)
+.mtj.tmp/
+
+# Package Files #
+*.war
+*.nar
+*.ear
+*.rar
+
+# editor files
+.vscode
+
+# virtual machine crash logs, see http://www.java.com/en/download/help/error_hotspot.xml
+hs_err_pid*
+
+# web
+/sources/web/node_modules/
+dist/
+
+## IDEA configs
+
+*.iml
+/idea/
+.idea/
+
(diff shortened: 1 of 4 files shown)


serge-rider:
Thanks for PR.
It gives some extra clues which I wasn't think about.

We already fixed Maven build though (in fact it changes quite frequently now as we still establishing right deployment and CI/CD).

We still not decided what should be included in Docker image and how it should be assembled.
Not sure that we should build entire product for that (it may be a prerequisite).
Perhaps we should add some kind of process supervisor (as Java process fail). Or maybe leave this to users.

Another question - what database configuration should be included in image. For now default deploy script adds sample SQLite database. Probably it should be removed.

tchibatron:
Is it possible to upload all standalone DBeaver artifacts to maven repo? If it was so, it would be able to make fully independent maven project, which would allow to use already created maven plugins for building frontend and docker as well.

dkrupenya:
Hi @Skn0tt 
What was the reason to include source code and build system (mvn and yarn) to the image?
I expected you just build locally and add final artifacts to the docker image

Skn0tt:
Making the Dockerfile self-reliant enables utilization of the Docker Hub builds.

I'm not firm on Docker builds in the Java world, but building locally and adding artifacts to the image seems fine as well. You'll loose the benefit of Docker Hub Autobuilds, but that's not that big of a deal.
Maybe [jib](https://github.com/GoogleContainerTools/jib)'s interesting to this? It makes building pure JVM-based images very easy, but I don't know how easy it is to integrate the frontend.

PR #35, opened 2020-05-07
Feat administration

Wroud:


webapp/packages/core/src/root/PermissionsService.ts
```diff
@@ -0,0 +1,56 @@
...
+
+  constructor(
+    private graphQLService: GraphQLService,
+  ) { }
+
+  has(id: string): boolean {
+    return this.permissions.data.has(id);
+  }
+
+  async hasAsync(id: string): Promise<boolean> {
```
dkrupenya: never required. Permissions will be loaded on application bootstrap. there will be no cases in normal workflow when there are no permissions

webapp/packages/core/src/app/TopNavBar/SettingsMenu/SettingsMenuService.ts
```diff
@@ -23,21 +23,21 @@ export class SettingsMenuService {
...
     this.addLocales();
   }
 
   getMenu() {
-    return this.menu.getMenu(this.settingsMenuToken);
+    return this.menu.getMenu(SettingsMenuService.settingsMenuToken);
   }
 
-  addMenuItem(panelId: string, options: IComputedMenuItemOptions) {
+  addMenuItem(options: IComputedMenuItemOptions, panelId: string = SettingsMenuService.settingsMenuToken) {
```
dkrupenya: revert back.
all other `addMenuItem` has signature "panel, option"

dkrupenya (changes requested):

webapp/packages/core/src/root/PermissionsService.ts
Wroud: fixed

webapp/packages/core/src/app/TopNavBar/SettingsMenu/SettingsMenuService.ts
Wroud: fixed

dkrupenya (approved):

PR #40, opened 2020-05-16
Fix sdk

Wroud:


server/bundles/io.cloudbeaver.server/schema/service.core.graphqls
```diff
@@ -199,30 +199,30 @@ type LogEntry {
...
     # Returns session state ( initialize if not )
-    sessionState: SessionInfo
+    sessionState: SessionInfo!
 
     # Session permissions
     sessionPermissions: [ID]!
 
     # Get driver info
-    driverList( id: ID ): [ DriverInfo! ]
+    driverList( id: ID ): [ DriverInfo ]!
```
dkrupenya: revert al such changes back
or write in the way `driverList( id: ID ): [ DriverInfo! ]!` if you sure that array exists

dkrupenya: you code will generate `Array<Maybe<DriverInfo>`. it is obiously crap because we have no sparse arrays on backend

dkrupenya (changes requested):

server/bundles/io.cloudbeaver.server/schema/service.core.graphqls
Wroud: fixed

dkrupenya (approved):

PR #54, opened 2020-05-25
Cb 92 builder package

dkrupenya:


webapp/packages/core/src/localization/LocalizationService.ts
```diff
@@ -44,7 +43,7 @@ export class LocalizationService {
               private settingsService: SettingsService) {
 
     // note that the default locale is always embedded in the build
-    this.localeMap.set(DEFAULT_LOCALE_NAME, defaultLocale);
+    this.loadLocaleAsync(DEFAULT_LOCALE_NAME);
```
Wroud: async operation in synchronous constructor is bad practice, I think better to move this code part to `init`

Wroud (changes requested):

webapp/packages/core/src/localization/LocalizationService.ts
dkrupenya: fixed

Wroud (approved):

PR #63, opened 2020-05-28
Cb 43 refresh data

dkrupenya:


webapp/packages/core/src/app/NavigationTree/NavigationTreeContextMenuService.ts
```diff
@@ -44,12 +44,26 @@ export class NavigationTreeContextMenuService {
...
       },
     };
     this.contextMenuService.addMenuItem<NodeWithParent>(this.contextMenuService.getRootMenuToken(), openNodeTab);
+
+    const refreshNode: IContextMenuItem<NodeWithParent> = {
+      id: 'refreshNode',
+      isPresent(context) {
+        return context.contextType === NavigationTreeContextMenuService.nodeContextType;
+      },
+      order: 1000,
```
Wroud: better to use `Number.MAX_SAFE_INTEGER`

webapp/packages/core/src/app/shared/NodesManager/NodesManagerService.ts
```diff
@@ -80,6 +80,13 @@ export class NodesManagerService {
     });
   }
 
+  async refreshNode(nodeId: string) {
+    console.log('refresh', nodeId);
```
Wroud: remove?

webapp/packages/core/src/app/shared/NodesManager/NodesManagerService.ts
```diff
@@ -80,6 +80,13 @@ export class NodesManagerService {
...
+    await this.purgeServerCache(nodeId);
+    this.nodesStore.removeNode(nodeId);
+    await this.updateNodeInfo(nodeId);
```
Wroud, line 86: remove can cause many bugs in case of an update, so we probably must check the result of request inside `updateNodeInfo` and remove node if it does not exist anymore, on other cases we probably just update node related data

Wroud (changes requested):

webapp/packages/core/src/app/NavigationTree/NavigationTreeContextMenuService.ts
dkrupenya: fixed

webapp/packages/core/src/app/shared/NodesManager/NodesManagerService.ts
dkrupenya: fixed

webapp/packages/core/src/app/shared/NodesManager/NodesManagerService.ts
dkrupenya: to be done later

Wroud (approved):

PR #69, opened 2020-05-29
feat: set focus on first form element CB-124

dkrupenya:


webapp/packages/core/src/blocks/InFocus.tsx
```diff
@@ -0,0 +1,31 @@
...
+
+type InFocusProps = {
+  children: ReactElement;
+}
+
+export function InFocus({ children }: InFocusProps) {
+  const isFirstRender: MutableRefObject<boolean> = useRef(true);
+  const childRef: MutableRefObject<HTMLElement | null> = useRef(null);
+
+  useEffect(() => {
```
Wroud: useLayoutEffect better for purpose like this

webapp/packages/core/src/blocks/InFocus.tsx
```diff
@@ -0,0 +1,31 @@
...
+        }
+
+      }
+    }
+  });
+  const childElement = React.Children.only(children);
+  return React.cloneElement(
+    childElement,
+    { ref: (el: HTMLElement) => childRef.current = el }
+  );
```
Wroud, line 30: first child can be not HTML element, so maybe better to wrap `<div {...props} ref={childRef}>{children}</div>`

webapp/packages/core/src/blocks/InFocus.tsx
dkrupenya: fixed

webapp/packages/core/src/blocks/InFocus.tsx
dkrupenya: just don't use InFocus around non-html components
The Idea is to have inFocus as HOC that dosn't append something to DOM

Wroud: yep I understand that idea, in this case, may be better to replace InFocus with hook that returns ref, that we can pass to HTML element

PR #165, opened 2020-09-30
fix(core-localization): some localization improvements

devnaumov:
CB-227


webapp/packages/core-blocks/src/locales/en.ts
```diff
@@ -0,0 +1 @@
+export default [['block_properties_table_name', 'Name'], ['block_properties_table_value', 'Value'], ['block_properties_table_add', 'Add']];
```
Wroud: please format, so each token will be on the new line

webapp/packages/core-blocks/src/locales/ru.ts
```diff
@@ -0,0 +1 @@
+export default [['block_properties_table_name', 'Название'], ['block_properties_table_value', 'Значение'], ['block_properties_table_add', 'Добавить']];
```
Wroud: same

Wroud (changes requested):

Wroud (approved):

PR #174, opened 2020-10-08
fix(core-connections): CB-390 driver properties order

devnaumov:
CB-390


webapp/packages/core-connections/src/DriverPropertiesService.ts
```diff
@@ -26,6 +26,9 @@ export class DriverPropertiesService {
       throw new Error('Driver properties loading failed');
     }
 
-    return response.driver[0].driverProperties! as ObjectPropertyInfo[];
+    const propertiesSortedByAlphabet = response.driver[0].driverProperties?.sort(
+      (a,b) => (a?.displayName ?? '').localeCompare(b?.displayName ?? ''));
+
+    return propertiesSortedByAlphabet as ObjectPropertyInfo[];
```
Wroud: this is the wrong place for that, we have `PropertiesTable` component that displays properties tables for some data, like in custom connection driver properties or same in connections administration, so we should sort properties in that component, also keep in mind that custom properties should appear in the beginning

Wroud (approved):

PR #234, opened 2020-12-02
fix(plugin-data-export): CB-580 add "white-space: pre-wrap" to source…

devnaumov:
… name


webapp/packages/plugin-data-export/src/ExportNotification/ExportNotification.tsx
```diff
@@ -25,9 +25,9 @@ const styles = css`
     composes: theme-typography--body2 from global;
     padding-top: 16px;
     max-height: 50px;
-    overflow: hidden;
-
+    overflow: auto;
     & pre {
+      white-space: pre-wrap;
```
Wroud, line 30: you should add this style in to `SNACKBAR_BODY_STYLES`

Wroud (changes requested):

PR #250, opened 2020-12-21
feat(core-connections): CB-140 add connection details visualization

devnaumov:


webapp/packages/core-connections/src/Administration/Connections/ConnectionsTable/ConnectionDetailsInfo.tsx
```diff
@@ -0,0 +1,56 @@
...
+  }
+
+  return styled(styles)(
+    <details-container as='div'>
+      {detailsToDisplay.map(({ icon, title }, idx) => (
+        <StaticImage key={idx} icon={icon} title={title} />
+      ))}
+    </details-container>
+  );
+};
```
Wroud: seems better to split into three components, one for icon displayment (accepts icon and title), and two:
1. connection type (local/origin)
2. template
last two components should use the first component for icon displayment

also, you can't use a container with multiple icons inside the single component, all icons should be in the same container (you can wrap `Placeholder` in `Connection` with container, or use self-contained icons (without any common container, except `TableColumnValue`))

webapp/packages/core-connections/src/Administration/Connections/ConnectionsTable/ConnectionsTable.tsx
```diff
@@ -58,7 +58,7 @@ export const ConnectionsTable = observer(function ConnectionsTable({
         <TableColumnHeader min />
         <TableColumnHeader>{translate('connections_connection_name')}</TableColumnHeader>
         <TableColumnHeader>{translate('connections_connection_address')}</TableColumnHeader>
-        <TableColumnHeader>{translate('connections_connection_template')}</TableColumnHeader>
+        <TableColumnHeader />
```
Wroud, line 61: you sure that column doesn't need a title?

Wroud (changes requested):

Wroud (approved):

Wroud (approved):

PR #257, opened 2020-12-24
improve(core-utils): CB-595 add getUniqueConnectionName func and chan…

devnaumov:
…ged sdk

### DIFF
--- a/webapp/packages/core-utils/src/getUniqueConnectionName.ts
+++ b/webapp/packages/core-utils/src/getUniqueConnectionName.ts
@@ -1,0 +1,33 @@
+/*
+ * cloudbeaver - Cloud Database Manager
+ * Copyright (C) 2020 DBeaver Corp and others
+ *
+ * Licensed under the Apache License, Version 2.0.
+ * you may not use this file except in compliance with the License.
+ */
+
+/**
+ * Help to create unique name for connection
+ * @param  {string} baseName
+ * @param  {string[]} connectionNames
+ * @returns string
+ */
+export function getUniqueConnectionName(baseName: string, connectionNames: string[]): string {
+  let index = 1;
+  let name = baseName;
+
+  const names = new Set();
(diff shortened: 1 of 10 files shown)


webapp/packages/core-utils/src/getUniqueConnectionName.ts
```diff
@@ -0,0 +1,33 @@
...
+  while (true) {
+    if (!names.has(name)) {
+      break;
+    }
+    name = `${baseName} (${index})`;
+    index++;
+  }
+
+  return name;
+}
```
Wroud: it should be in `core-connections` package

webapp/packages/core-utils/src/getUniqueConnectionName.ts
```diff
@@ -0,0 +1,33 @@
...
+ * @returns string
+ */
+export function getUniqueConnectionName(baseName: string, connectionNames: string[]): string {
+  let index = 1;
+  let name = baseName;
+
+  const names = new Set();
+  for (const connectionName of connectionNames) {
+    names.add(connectionName);
+  }
```
Wroud: extra step, remove it

webapp/packages/core-utils/src/getUniqueConnectionName.ts
```diff
@@ -0,0 +1,33 @@
...
+
+  while (true) {
+    if (!names.has(name)) {
```
Wroud: `!connectionNames.includes(name)`

Wroud (changes requested):

webapp/packages/core-utils/src/getUniqueConnectionName.ts
Wroud: also, you can do `const names = new Set(connectionNames);` instead, but I don't see the real reason for that

Wroud (approved):

PR #264, opened 2021-01-14
fix(core-blocks): CB-605 add opacity active state for IconButton

devnaumov:


Wroud:
looks strange, I think it better to set default color for button primary and `:not(:active) { color: inherit; }`, so we get an effect like the item was active (currently it's more like disabled)

PR #269, opened 2021-01-20
fix(plugin-data-viewer): CB-631 async fetch cancel button now always …

devnaumov:
…available

### DIFF
--- a/webapp/packages/plugin-data-viewer/src/ContainerDataSource.ts
+++ b/webapp/packages/plugin-data-viewer/src/ContainerDataSource.ts
@@ -67,12 +67,12 @@ export class ContainerDataSource extends DatabaseDataSource<IDataContainerOption
       throw new Error('containerNodePath must be provided for table');
     }
 
+    this.currentFetchTableProcess = new FetchTableDataAsyncProcess(this.graphQLService, this.notificationService);
     const executionContext = await this.ensureContextCreated();
+
     const limit = this.count;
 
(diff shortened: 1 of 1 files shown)


webapp/packages/plugin-data-viewer/src/ContainerDataSource.ts
```diff
@@ -170,6 +169,7 @@ export class ContainerDataSource extends DatabaseDataSource<IDataContainerOption
   private async ensureContextCreated(): Promise<IExecutionContext> {
     if (!this.executionContext) {
       if (!this.options) {
+        this.currentFetchTableProcess = null;
```
Wroud, line 172: seems like the wrong place for that, maybe you should do it in `finally` after the process finished

Wroud (changes requested):

webapp/packages/plugin-data-viewer/src/ContainerDataSource.ts
```diff
@@ -67,12 +67,12 @@ export class ContainerDataSource extends DatabaseDataSource<IDataContainerOption
       throw new Error('containerNodePath must be provided for table');
     }
 
+    this.currentFetchTableProcess = new FetchTableDataAsyncProcess(this.graphQLService, this.notificationService);
     const executionContext = await this.ensureContextCreated();
```
Wroud, line 71: you can't start the process without context

Wroud (changes requested):

PR #314, opened 2021-03-16
fix(plugin-data-spreadsheet-new): CB-757 column selection mac support

devnaumov:


Wroud:
please correct commit message

PR #333, opened 2021-04-01
fix(plugin-connection-template): CB-910 unique connection name for te…

devnaumov:
…mplate connection

### DIFF
--- a/webapp/packages/plugin-connection-template/src/ConnectionDialog/ConnectionController.ts
+++ b/webapp/packages/plugin-connection-template/src/ConnectionDialog/ConnectionController.ts
@@ -114,8 +114,7 @@ export class ConnectionController
     this.isConnecting = true;
     this.clearError();
     try {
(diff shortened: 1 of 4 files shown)


webapp/packages/plugin-connection-template/src/ConnectionDialog/ConnectionController.ts
```diff
@@ -114,7 +114,10 @@ export class ConnectionController
     this.isConnecting = true;
     this.clearError();
     try {
-      const connection = await this.connectionInfoResource.createFromTemplate(this.template.id);
+      const connectionNames = Array.from(this.connectionInfoResource.data.values())
+        .map(connection => connection.name);
```
Wroud:
```
const connectionNames = Array.from(this.connectionInfoResource.values)
        .map(connection => connection.name);
```

Wroud (changes requested):

Wroud (approved):

PR #335, opened 2021-04-01
fix(core-blocks): CB-451 add missing tooltips

devnaumov:


webapp/packages/core-blocks/src/FormControls/InputField.tsx
```diff
@@ -84,7 +84,7 @@ export const InputField: InputFieldType = observer(function InputField({
 
   return styled(styles)(
     <field as="div" className={className} {...use({ long, short })}>
-      <field-label as='label'>{children} {required && '*'}</field-label>
+      <field-label as='label' title={rest.title}>{children} {required && '*'}</field-label>
       <input
         {...rest}
         name={name}
```
Wroud, line 90: so, in that case, we have the same title on input and label, is it expected?

PR #336, opened 2021-04-02
db2_i feature working and tested with current dbeaver build only 3 files affected in cloudbeaver

jjsilvan:
db2_i feature working and tested with current dbeaver build only 3 files affected


jjsilvan:
only cloudbeaver workink folder for AS400 db2-jt400 using current link from dbeaver

serge-rider:
Thanks!
We'll test it and include in the following CB version.

PR #359, opened 2021-04-29
dbeaver/dbeaver#11439 Use getLargeUpdateCount for rows update by default

ShadelessFox:


server/bundles/io.cloudbeaver.server/src/io/cloudbeaver/service/sql/WebSQLProcessor.java
```diff
@@ -490,7 +490,7 @@ private void fillQueryResults(
                     results.setResultSet(dataReceiver.getResultSet());
                 }
             } else {
-                int updateRowCount = dbStat.getUpdateRowCount();
+                long updateRowCount = dbStat.getUpdateRowCount();
                 if (updateRowCount >= 0) {
                     results.setUpdateRowCount(updateRowCount);
                 } else {
```
Wroud, line 496: is this value sent to the frontend? (js supports numbers in the range -(2^53 - 1) to 2^53 - 1)

ShadelessFox: I believe so. 

The model already accepts `long` which is 64-bit integer (-2<sup>63</sup> to 2<sup>63</sup>-1). Perhaps you should handle this case then.

PR #435, opened 2021-08-05
fix(plugin-data-viewer): CB-1129 handle readonly boolean representati…

devnaumov:
…on state

### DIFF
--- a/webapp/packages/core-theming/src/styles/_form-controls.scss
+++ b/webapp/packages/core-theming/src/styles/_form-controls.scss
@@ -33,9 +33,11 @@
       width: 16px;
     }
 
-    &:global([disabled]),
-    &:global([readonly]) {
-      opacity: 0.8;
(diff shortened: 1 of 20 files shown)


webapp/packages/core-theming/src/styles/_form-controls.scss
```diff
@@ -33,9 +33,11 @@
...
     }
 
-    &:global([disabled]),
-    &:global([readonly]) {
-      opacity: 0.8;
+    &[type="text"] {
+      &:global([disabled]),
+      &:global([readonly]) {
+        opacity: 0.8;
+      }
```
Wroud: probably affects too many places, better to revert and make a fix for a specific place
```
some-selector {
  opacity: 1 !important;
}
```

Wroud (changes requested):

Wroud (approved):

PR #460, opened 2021-08-29
fix(plugin-authentication): CB-1063 add disabled property to auth con…

devnaumov:
…figuration

### DIFF
--- a/webapp/packages/core-authentication/src/AuthConfigurationsResource.ts
+++ b/webapp/packages/core-authentication/src/AuthConfigurationsResource.ts
@@ -18,9 +18,7 @@ import { MetadataMap } from '@cloudbeaver/core-utils';
 
 const NEW_CONFIGURATION_SYMBOL = Symbol('new-configuration');
 
-type NewConfiguration = AdminAuthProviderConfiguration & {
-  [NEW_CONFIGURATION_SYMBOL]: boolean; timestamp: number;
-};
+type NewConfiguration = AdminAuthProviderConfiguration & { [NEW_CONFIGURATION_SYMBOL]: boolean; timestamp: number };
 
 @injectable()
 export class AuthConfigurationsResource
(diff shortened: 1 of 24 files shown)


webapp/packages/core-blocks/src/Table/TableItemSelect.tsx
```diff
@@ -52,8 +52,8 @@ export const TableItemSelect: React.FC<Props> = observer(function TableItemSelec
   return styled(styles, checkboxStyles)(
     <Checkbox
       className={className}
-      disabled={context.selectDisabled || disabled}
-      checked={checked || context.isSelected()}
+      disabled={disabled ?? context.selectDisabled}
+      checked={checked ?? context.isSelected()}
```
Wroud: revert, the disabled property can't override the context, so logic is: if any of the two rules is true, then the checkbox is disabled or checked

webapp/packages/core-authentication/src/AuthConfigurationsResource.ts
```diff
@@ -10,11 +10,18 @@ import { EAdminPermission } from '@cloudbeaver/core-administration';
...
+  AdminAuthProviderConfiguration, AuthProviderConfiguration,
+  CachedMapResource, GetAuthProviderConfigurationsQueryVariables,
   GraphQLService, ResourceKey, ResourceKeyList, resourceKeyList, ResourceKeyUtils
 } from '@cloudbeaver/core-sdk';
 import { MetadataMap } from '@cloudbeaver/core-utils';
 
+const NEW_CONFIGURATION_SYMBOL = Symbol('new-configuration');
+
+type NewConfiguration = AdminAuthProviderConfiguration & {
+  [NEW_CONFIGURATION_SYMBOL]: boolean; timestamp: number;
```
Wroud: please don't inline properties in not inline type declaration

webapp/packages/core-authentication/src/AuthConfigurationsResource.ts
```diff
@@ -87,7 +94,17 @@ export class AuthConfigurationsResource
 
   async saveConfiguration(config: AdminAuthProviderConfiguration): Promise<AdminAuthProviderConfiguration> {
```
Wroud, line 79: please place all private methods after public

Wroud (changes requested):

Wroud (approved):

PR #492, opened 2021-09-27
fix(core-dialogs): remove onMouseEnter from MenuButton

devnaumov:


webapp/packages/core-dialogs/src/Menu/MenuPanel/menuPanelStyles.ts
```diff
@@ -29,6 +29,9 @@ export const menuPanelStyles = composes(
     }
   `,
   css`
+    button-wrapper MenuButton {
+      width: 100%;
+    }
```
Wroud: replace with:
```
menu-box, menu-panel-wrapper {
  display: flex;
  flex-direction: column;
}
```

and remove these styles from `menu-box`

Wroud (changes requested):

Wroud (approved):

PR #495, opened 2021-09-29
fix(plugin-connections): CB-783 ask for reconnect if connection was u…

devnaumov:
…pdated


webapp/packages/plugin-connections/src/PublicConnectionForm/PublicConnectionFormService.ts
```diff
@@ -123,6 +135,25 @@ export class PublicConnectionFormService {
...
+      confirmActionText: 'ui_yes',
+    });
+
+    if (result === DialogueStateResult.Rejected) {
+      return;
+    }
+
+    try {
+      await this.connectionInfoResource.close(id);
+      await this.connectionInfoResource.init({ id });
```
Wroud: use `ConnectionAuthService` `auth` method instead of `init`

Wroud (changes requested):

Wroud (approved):

PR #623, opened 2022-01-19
CB-825 fix duplicated tabs in Firebird

alexander-skoblikov:


server/bundles/io.cloudbeaver.server/src/io/cloudbeaver/service/sql/WebSQLProcessor.java
```diff
@@ -766,5 +767,10 @@ public void close() {
 
     ///////////////////////////////////////////////////////
     // Utils
-
+    private static int resolveMaxResultsCount(@Nullable DBPDataSource dataSource) {
+        if (dataSource == null) {
+            return MAX_RESULTS_COUNT;
+        }
+        return dataSource.getInfo().supportsMultipleResults() ? MAX_RESULTS_COUNT : 1;
```
alexander-skoblikov, line 774: similar check in the dbeaver
https://github.com/dbeaver/dbeaver/blob/0b039dafb3fb2c42fd51c2fa244c0f0c789ec42d/plugins/org.jkiss.dbeaver.ui.editors.sql/src/org/jkiss/dbeaver/ui/editors/sql/execute/SQLQueryJob.java#L573

PR #754, opened 2022-04-07
CB-1891 display build script execute errors

alexander-skoblikov:


deploy/build.sh
```diff
@@ -1,5 +1,6 @@
 #!/bin/bash
 set -e
```
mayerro: This row does not needed anymore

mayerro (approved):

PR #759, opened 2022-04-11
fix(plugin-connections): CB-1888 hide auth models that need local con…

devnaumov:
…figuration

### DIFF
--- a/webapp/packages/core-connections/src/DatabaseAuthModelsResource.ts
+++ b/webapp/packages/core-connections/src/DatabaseAuthModelsResource.ts
@@ -6,7 +6,9 @@
  * you may not use this file except in compliance with the License.
  */
 
+import { EAdminPermission } from '@cloudbeaver/core-administration';
 import { injectable } from '@cloudbeaver/core-di';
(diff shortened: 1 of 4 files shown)


webapp/packages/core-connections/src/DatabaseAuthModelsResource.ts
```diff
@@ -16,8 +18,21 @@ import {
...
+    this.isModelAvailable = this.isModelAvailable.bind(this);
+  }
+
+  isModelAvailable(model: DatabaseAuthModel) {
+    if (this.permissionsService.has(EAdminPermission.admin)) {
+      return true;
+    }
+
+    return !model.requiresLocalConfiguration;
   }
```
Wroud: better to split to 
`const admin = usePermission(EAdminPermission.admin)` 
...
`const availableAuthModels = applicableAuthModels.filter(model => !!model && (admin || !model.requiresLocalConfiguration));`

or modify `isModelAvailable` to asynchronous function to guarantee that permissions are loaded and updated before filtering auth models

Wroud (changes requested):

Wroud (approved):

PR #847, opened 2022-05-24
fix(plugin-sql-editor-navigation-tab): CB-2056 save script on tab close

devnaumov:


webapp/packages/plugin-sql-editor-navigation-tab-resource/src/SqlEditorTabResourceService.ts
```diff
@@ -138,7 +146,29 @@ export class SqlEditorTabResourceService {
...
-  private onTabCloseHandler(tab: ITab | undefined) {
+  private async canCloseTabHandler(
+    tab: ITab<ISqlEditorTabState>,
+    contexts: IExecutionContextProvider<any>
+  ) {
+    if (!this.state.has(tab.id)) {
+      return;
+    }
+
+    if (!await this.updateResource(tab, tab.handlerState.query)) {
```
Wroud: condition is always false

Wroud (changes requested):

Wroud (approved):

PR #843, opened 2022-05-22
Enhance build script

pan3793:
Currently, user must run `cd deploy` before building the project.

This PR enhances the build script to make build script can run from any folder.

PS: I don't touch `bat` because I am not familiar with that language and don't have Windows machine to test.

PR #861, opened 2022-05-26
Update zh locales

pan3793:


Wroud (approved):

pan3793:
Thanks @Wroud for review, and free feel to ping me if there is anything need to translate to Chinese.
BTW, would you please take a look at #843, do you think the change make sense?

PR #906, opened 2022-06-19
fix(plugin-sql-editor-navigation-tab): CB-1675 select connection if t…

devnaumov:
…here are no schemas

### DIFF
--- a/webapp/packages/plugin-sql-editor-navigation-tab/src/SqlEditorTabService.ts
+++ b/webapp/packages/plugin-sql-editor-navigation-tab/src/SqlEditorTabService.ts
@@ -131,11 +131,17 @@ export class SqlEditorTabService extends Bootstrap {
     }
   }
 
-  private getNavNode() {
+  private getNavNode(tab: ITab<ISqlEditorTabState>) {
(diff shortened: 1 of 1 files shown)


webapp/packages/plugin-sql-editor-navigation-tab/src/SqlEditorTabService.ts
```diff
@@ -131,11 +131,17 @@ export class SqlEditorTabService extends Bootstrap {
...
+  private getNavNode(tab: ITab<ISqlEditorTabState>) {
     const objectCatalog = this.connectionSchemaManagerService.currentObjectCatalog;
     const objectSchema = this.connectionSchemaManagerService.currentObjectSchema;
+    const connectionId = this.getConnectionId(tab);
+
+    let nodeId = objectSchema?.id ?? objectCatalog?.id;
+
+    if (!nodeId && connectionId) {
+      nodeId = NodeManagerUtils.connectionIdToConnectionNodeId(connectionId);
+    }
```
Wroud, line 143: we can't use data from `connectionSchemaManagerService`

Wroud (changes requested):

PR #907, opened 2022-06-20
feat(plugin-sql-editor): CB-1498 ask result-set save confirmation on …

devnaumov:
…certain actions


webapp/packages/plugin-sql-editor/src/SqlEditorBootstrap.ts
```diff
@@ -0,0 +1,86 @@
...
+    super();
+  }
+
+  register(): void {
+    this.sqlEditorMenuService.register();
+
+    this.authenticationService.onLogout.addHandler(this.logoutHandler.bind(this));
+    this.connectionsManagerService.onDisconnect.addHandler(this.connectionHandler.bind(this));
+    this.connectionsManagerService.onDelete.addHandler(this.connectionHandler.bind(this));
+    this.serverConfigurationService.saveTask.addHandler(this.saveHandler.bind(this));
```
Wroud: `serverConfigurationService` can't be used here
probably we don't want to check something from public side while we in administration

Wroud: also we can't use `sqlEditorTabService` here (because we don't want to know about tabs from `plugin-sql-editor`)

Wroud (changes requested):

Wroud (approved):

PR #911, opened 2022-06-20
feat(plugin-sql-editor): CB-1498 ask confirmation for the ObjectViewe…

devnaumov:
…r result-set tab


webapp/packages/plugin-object-viewer/src/ObjectViewerBootstrap.ts
```diff
@@ -6,24 +6,53 @@
...
+      for (const tab of tabs) {
+        const canClose = await this.dataViewerTabService.canClose(tab);
+
+        if (!canClose) {
+          ExecutorInterrupter.interrupt(contexts);
+          return;
+        }
+      }
+    }
+  }
```
Wroud: this logic should be in data viewer plugin

Wroud (changes requested):

Wroud (approved):

PR #918, opened 2022-06-22
CB-1912 change cb scripts for running tests

yagudin10:


serge-rider:
No need to run tests during deploy script

PR #1049, opened 2022-08-15
CB-2396 validate anonymous access in app configuration

yagudin10:


alexander-skoblikov (commented):
move fields associated with anonymous user from BaseWebAppConfiguration to CBAppConfig, then we will not need additional validation in QM/RM/DC services

PR #1191, opened 2022-09-30
fix(plugin-projects): CB-2677 submit form only if the state is valid

devnaumov:


webapp/packages/plugin-projects/src/FolderDialog.tsx
```diff
@@ -120,7 +120,9 @@ export const FolderDialog: DialogComponent<FolderDialogPayload, IFolderDialogRes
   });
 
   function resolveHandler() {
-    resolveDialog({ folder: state.value, projectId: state.projectId });
+    if (state.valid) {
+      resolveDialog({ folder: state.value, projectId: state.projectId });
+    }
```
Wroud, line 126: please execute `await state.validate()` before to ensure that form is validate (usually it validated, but it's better to ensure it)

Wroud (changes requested):

Wroud (approved):

PR #1204, opened 2022-10-07
CB-2077 save pattern variables

yagudin10:


server/bundles/io.cloudbeaver.server/src/io/cloudbeaver/server/CBApplication.java
```diff
@@ -675,6 +672,15 @@ protected Map<String, Object> readConfiguration(File configFile) throws DBExcept
         return configProps;
     }
 
+    private void readConfigurationFile(File configFile, Map<String, Object> configProps) throws DBException {
```
alexander-skoblikov: return Map<String, Object>

PR #1252, opened 2022-10-31
CB-2638 Oracle wallet authentication

oksana-cherniavskaia:
FE part


serge-rider:
Why there are 2 PRs (identical?)

PR #1256, opened 2022-11-02
CB-2797: TE. Support deleting connection folders

oksana-cherniavskaia:


server/bundles/io.cloudbeaver.model/src/io/cloudbeaver/model/rm/local/LocalResourceController.java
```diff
@@ -332,6 +333,25 @@ public void deleteProjectDataSources(@NotNull String projectId, @NotNull String[
         registry.checkForErrors();
     }
 
+    @Override
+    public void deleteProjectConnectionFolders(
+        @NotNull String projectId,
+        @NotNull String[] folderPaths,
+        boolean dropContents
+    ) throws DBException {
+        final DBPProject project = getProjectMetadata(projectId);
```
alexander-skoblikov: we don't use `final` according to our code style

oksana-cherniavskaia: ((that is unfortunate , btw )

PR #1267, opened 2022-11-08
CB-2858 update in memory projects

yagudin10:


server/bundles/io.cloudbeaver.model/src/io/cloudbeaver/model/session/WebSession.java
```diff
@@ -295,7 +296,11 @@ public synchronized void refreshUserData() {
      */
     public synchronized void updateProjectConnection(DBPProject project, List<String> connectionIds, CBEventConstants.EventType type) {
         DBPDataSourceRegistry registry = project.getDataSourceRegistry();
+        boolean isInMemory = project.isInMemory();
+        ((BaseProjectImpl) project).setInMemory(false);
```
alexander-skoblikov: Looks strange

PR #1353, opened 2022-12-07
CB-2838 exclude non secured properties from auth dialog if they are s…

devnaumov:
…aved

Propably we want to pass the filterProperties function to the ConnectionAuthenticationForm or to the ObjectPropertyInfoForm


webapp/packages/plugin-connections/src/DatabaseAuthDialog/DatabaseAuthDialog.tsx
```diff
@@ -94,7 +98,7 @@ export const DatabaseAuthDialog: DialogComponent<Payload> = observer(function Da
             <ConnectionAuthenticationForm
```
Wroud: please add to `ConnectionAuthenticationForm` property: `hideFeatures` which is a list of features that's shuldn't be displayed

Wroud (changes requested):

Wroud (approved):

PR #1362, opened 2022-12-14
CB-2968 preload possible drivers

devnaumov:


webapp/packages/plugin-connection-search/src/Search/SearchDatabase.tsx
```diff
@@ -8,19 +8,22 @@
...
 
 import { ConnectionSearchService } from './ConnectionSearchService';
 import { DatabaseList } from './DatabaseList';
 
 export const SearchDatabase: React.FC = observer(function SearchDatabase() {
   const connectionSearchService = useService(ConnectionSearchService);
+  const possibleDriversList = connectionSearchService.databases.map(db => db.possibleDrivers).flat();
+
   useMapResource(SearchDatabase, ProjectInfoResource, CachedMapAllKey);
+  useMapResource(SearchDatabase, DBDriverResource, resourceKeyList(possibleDriversList));
```
Wroud: maybe it's better to load all drivers at once (because in ee it can be up to 4 different drivers for same database, it will cause 4 requests)

or we can dynamically determine if it's 1 driver, then we can load it, if it's more, we will load all

Wroud (changes requested):

Wroud (approved):

PR #1392, opened 2022-12-29
CB-2305 open empty folders when in the folder view

devnaumov:
image


Wroud (approved):
Partially fix issue with folders, but for other objects it still can happen

PR #1422, opened 2023-01-17
CB-3029 add information about recursive deletion

devnaumov:


webapp/packages/plugin-resource-manager/src/locales/en.ts
```diff
@@ -20,4 +20,5 @@ export default [
   ['plugin_resource_manager_script_not_found_message', 'The script was probably removed or renamed. The last known value will be used'],
   ['plugin_resource_manager_script_name_invalid_characters_message', 'Script\'s name may contain the following symbols "_-$.()@" and can\'t start with a dot'],
   ['plugin_resource_manager_save_script_project_restriction_descripion', "The script's project can't be different from the connection's project"],
+  ['plugin_resource_manager_folder_delete_confirmation', 'You\'re going to delete "{arg:name}". All related scripts will be deleted. Are you sure?'],
```
Wroud: `You\'re going to delete "{arg:name}" and all its contents. Are you sure?`

Wroud (changes requested):

webapp/packages/plugin-resource-manager/src/locales/en.ts
Wroud: or
`Are you sure you want to delete "{arg:name}" and all its contents?`

Wroud (approved):

PR #1452, opened 2023-02-03
Replace bash process with java process instead of creating a sub process

### LINKED ISSUE #2454: Replace bash process with Java process
Replace the bash process with the Java process instead of creating a sub-process.

To avoid incorrect container closing (as an example).

MarcelCoding:
This enables docker to do a graceful shutdown of the container, because the java process gets notified that it should be terminated.


LonwoLonwo (approved):

LonwoLonwo:
Thanks for your contribution! I'm very sorry for the late response.
This enhancement will be available in the next release, 24.0.3, on 22 April 2024.

PR #1457, opened 2023-02-07
CB-3105 fix: websocket retry / repeat delay

Wroud:


webapp/packages/core-root/src/SessionEventSource.ts
```diff
@@ -37,11 +37,7 @@ export interface ITopicSubEvent extends ISessionEvent {
   topicId: SessionEventTopic;
 }
 
-const retryInterval = 5000;
-
-const retryConfig: RetryConfig = {
-  delay: retryInterval,
-};
+const retryInterval = 30 * 1000;
```
devnaumov: Please name variables outside of classes and components in upper case. RETRY_INTERVAL

devnaumov (approved):

webapp/packages/core-root/src/SessionEventSource.ts
Wroud: fixed

PR #1467, opened 2023-02-10
CB-3008 TE: limit access to private project

EmYar:
Creates an "isPrivateProjectsAccessible" flag in the RM config with default == false, which excludes private projects from the getProjects request on the TE-only backend.


server/bundles/io.cloudbeaver.model/src/io/cloudbeaver/model/app/WebAppConfiguration.java
```diff
@@ -41,4 +41,8 @@ public interface WebAppConfiguration {
     boolean isFeaturesEnabled(String[] requiredFeatures);
 
     boolean isFeatureEnabled(String id);
+
+    default boolean isPrivateProjectsAccessible() {
```
alexander-skoblikov:
```suggestion
    default boolean isSupportsCustomConnections() {
```
we can use already existing method

PR #1486, opened 2023-02-20
CB-2669 fix dnd to the project

devnaumov:


Wroud:
we need to display only related projects (like the same project of dragging item)

Wroud (approved):

PR #1487, opened 2023-02-20
CB-2577 validate host only in the manual mode

devnaumov:


webapp/packages/plugin-connections/src/ConnectionForm/Options/ConnectionOptionsTabService.ts
```diff
@@ -157,7 +157,7 @@ export class ConnectionOptionsTabService extends Bootstrap {
   ) {
     const validation = contexts.getContext(this.connectionFormService.connectionValidationContext);
 
-    if (!state.config.host?.length) {
+    if (state.config.configurationType === DriverConfigurationType.Manual && state.config.host?.length === 0) {
```
Wroud: you also need to check that connection's driver is not embeded `driver?.embedded`

Wroud (changes requested):

Wroud (approved):

PR #1504, opened 2023-02-27
CB-3144 Fix styles for sql editor tab

tfokina:
`:has` not supported by default in FF: https://caniuse.com/?search=has


Wroud (changes requested):
image
this style are not equal, now we have free space from the right side

Wroud (approved):

PR #1505, opened 2023-02-27
CB-3129 add filter script

devnaumov:
execute it on specific package as ```yarn filter``` or execute it on all packages as ```lerna run filter```

### DIFF
--- a/webapp/packages/core-cli/bin/filter-deps.js
+++ b/webapp/packages/core-cli/bin/filter-deps.js
@@ -1,0 +1,102 @@
+#!/usr/bin/env node
+
+'use strict';
+process.title = 'core-filter-deps';
+
+const fs = require('fs');
+const path = require('path');
+
+// Resolve current root
+const root = path.resolve('.');
+
+// Replace with the correct path to the `package.json` file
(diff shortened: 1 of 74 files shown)


webapp/packages/core-cli/bin/filter-deps.js
```diff
@@ -0,0 +1,102 @@
...
+      const filePath = path.join(directory, file);
+      const fileStat = fs.statSync(filePath);
+
+      if (fileStat.isDirectory()) {
+        // If it is, recursively process the subdirectory
+        processDirectory(filePath);
+      } else {
+        // Otherwise, check if the file is a .ts/.tsx file
+        const extensions = ['.ts', '.tsx'];
+        if (extensions.some(ext => file.endsWith(ext)) && !file.split('.').includes('test')) {
```
Wroud:
```
const extensions = /(?<!\.test).tsx?$/i
extensions.some(ext => file.endsWith(ext))
```
```suggestion
        const extensions = /(?<!\.test).tsx?$/i;
        if (extensions.test(file)) {
```

Wroud (changes requested):

Wroud (approved):

PR #1506, opened 2023-02-27
Cb 3115 add new create delete project events

EmYar:


server/bundles/io.cloudbeaver.model/src/io/cloudbeaver/model/session/WebSession.java
```diff
@@ -1048,7 +1064,7 @@ public void subTask(String name) {
         }
     }
 
-    private class TaskProgressMonitor extends ProxyProgressMonitor {
+    private static class TaskProgressMonitor extends ProxyProgressMonitor {
```
alexander-skoblikov, line 1067: redundant change

server/bundles/io.cloudbeaver.model/src/io/cloudbeaver/model/session/WebSession.java
```diff
@@ -1070,7 +1086,7 @@ public void subTask(String name) {
         }
     }
 
-    private class PersistentAttribute {
+    private static class PersistentAttribute {
```
alexander-skoblikov, line 1089: redundant change

server/bundles/io.cloudbeaver.server/src/io/cloudbeaver/server/events/WsProjectLifeCycleEventHandler.java
```diff
@@ -0,0 +1,49 @@
...
+import io.cloudbeaver.DBWConstants;
+import io.cloudbeaver.model.session.BaseWebSession;
+import org.jkiss.code.NotNull;
+import org.jkiss.dbeaver.DBException;
+import org.jkiss.dbeaver.Log;
+import org.jkiss.dbeaver.model.websocket.event.WSEventTopic;
+import org.jkiss.dbeaver.model.websocket.event.WSEventType;
+import org.jkiss.dbeaver.model.websocket.event.WSProjectUpdateEvent;
+
+public class WsProjectLifeCycleEventHandler extends WSProjectUpdatedEventHandler<WSProjectUpdateEvent> {
```
alexander-skoblikov, line 12: WsProjectLifeCycleEventHandler rename to WSProjectUpdatedEventHandler

old WSProjectUpdatedEventHandler rename to WSAbstractProjectEventHandler

server/bundles/io.cloudbeaver.model/src/io/cloudbeaver/model/session/WebSession.java
EmYar: Static inner classes is more memory efficient. (https://stackoverflow.com/questions/20380600/gc-performance-hit-for-inner-class-vs-static-nested-class)

The IDEA has an inspection that reports the possibility of making the inner class static. Here is description
> A static inner class does not keep an implicit reference to its enclosing instance. This prevents a common cause of memory leaks and uses less memory per instance of the class.

Are there specific reasons for leaving a non-static declaration?

EmYar:
merged into CB-2888-project-sync in commit https://github.com/dbeaver/cloudbeaver/commit/d37a7f840c65e062c4723c64af501d5ee332d9bb

PR #1538, opened 2023-03-10
CB-1577-H2-driver-update

EmYar:
Implements migrator


server/bundles/io.cloudbeaver.service.security/src/io/cloudbeaver/service/security/db/H2Migrator.java
```diff
@@ -0,0 +1,219 @@
+package io.cloudbeaver.service.security.db;
```
serge-rider, line 17: Copyright missng

server/bundles/io.cloudbeaver.service.security/src/io/cloudbeaver/service/security/db/CBDatabase.java
```diff
@@ -110,18 +110,19 @@ public void initialize() throws DBException {
         if (CommonUtils.isEmpty(databaseConfiguration.getDriver())) {
             throw new DBException("No database driver configured for CloudBeaver database");
         }
-        DBPDriver driver = DataSourceProviderRegistry.getInstance().findDriver(databaseConfiguration.getDriver());
-        if (driver == null) {
+        var dataSourceProviderRegistry = DataSourceProviderRegistry.getInstance();
+        DBPDriver driverProvider = dataSourceProviderRegistry.findDriver(databaseConfiguration.getDriver());
+        if (driverProvider == null) {
```
serge-rider: Let's keep variable name 'driver'. driverProvider is confusing.

server/bundles/io.cloudbeaver.service.security/src/io/cloudbeaver/service/security/db/H2Migrator.java
```diff
@@ -0,0 +1,219 @@
...
+
+public class H2Migrator {
+
+    private static final Log log = Log.getLog(H2Migrator.class);
+
+    private static final String H2_DRIVER_NAME = "h2_embedded";
+    private static final String H2_V2_DRIVER_NAME = "h2_embedded_v2";
+    private static final String BACKUP_V1_FILENAME = "h2db_backup";
+    private static final String EXPORT_FILE_NAME = "H2v1ExportScript";
+    @SuppressWarnings("SqlNoDataSourceInspection")
```
serge-rider: Please don't use IDEA-specific annotations in code

server/bundles/io.cloudbeaver.service.security/src/io/cloudbeaver/service/security/db/H2Migrator.java
```diff
@@ -0,0 +1,219 @@
...
+            log.error("H2 to v2 migration failed and reverted", e);
+        }
+    }
+
+    @Nullable
+    private Driver getV2DriverIfMigrationNeeded() {
+        if (!H2_DRIVER_NAME.equals(databaseConfiguration.getDriver()) &&
+            !H2_V2_DRIVER_NAME.equals(databaseConfiguration.getDriver())
+        ) {
+            log.debug("Non-H2 database detected. No migration needed");
```
serge-rider: No need to write debug logs about that. It will be in the log on every app start.

server/bundles/io.cloudbeaver.service.security/src/io/cloudbeaver/service/security/db/H2Migrator.java
```diff
@@ -0,0 +1,219 @@
...
+            return null;
+        }
+
+        if (dbUrl.startsWith("jdbc:h2:mem:")) {
+            log.debug("In-memory database detected. No migration needed");
```
serge-rider: No need to write debug logs about that. It will be in the log on every app start.

server/bundles/io.cloudbeaver.service.security/src/io/cloudbeaver/service/security/db/H2Migrator.java
```diff
@@ -0,0 +1,219 @@
...
+                return null;
+            }
+            v2Driver = driverProvider.getDriverInstance(monitor);
+        } catch (DBException e) {
+            log.error("Failed to initialize H2 v2 driver", e);
+            return null;
+        }
+
+        try (var ignored = v2Driver.connect(dbUrl, dbProperties)) {
+            log.debug("Current H2 database is v2. No migration needed");
```
serge-rider: No need to write debug logs about that. It will be in the log on every app start.

server/bundles/io.cloudbeaver.service.security/src/io/cloudbeaver/service/security/db/H2Migrator.java
```diff
@@ -0,0 +1,219 @@
...
+            migrateDb(workPaths, v2Driver);
+            updateConfig();
+        } catch (Exception e) {
+            rollback(workPaths);
+            log.error("H2 to v2 migration failed and reverted", e);
+        }
+    }
+
+    @Nullable
+    private Driver getV2DriverIfMigrationNeeded() {
```
serge-rider: Id don't like the idea to connect to old database with new driver and visa-versa.
Why can;'t we just use different name for V2 database and check file existence?
- if h2_v2 database doesn't exist and h2_v1 exists then do migration (and drop v1 after it)
- if h2_v2 database doesn't exist and h2_v1 doesn't exist then skip migraton
- if h2_v2 database exists then skip migraton

serge-rider (changes requested):

server/bundles/io.cloudbeaver.service.security/src/io/cloudbeaver/service/security/db/CBDatabase.java
EmYar: It is not a driver: DriverDescriptor does not implement the Driver interface. 
"driver" is a confusing name in combination with "driverInstance" from the old version.

What about "driverDescriptor" instead of "driverProvider"?

server/bundles/io.cloudbeaver.service.security/src/io/cloudbeaver/service/security/db/H2Migrator.java
EmYar: It is a debug log with debug information.

> It will be in the log on every app start.

Only if the logging level is set to DEBUG, which should not be the case during typical use of the application

server/bundles/io.cloudbeaver.service.security/src/io/cloudbeaver/service/security/db/H2Migrator.java
EmYar: We cannot use the filename to check that the database is v1, because we offer the possibility to customize the filename in the config:
```
server: {
  database: {
    url: "jdbc:h2:${workspace}/.data/cb.h2.dat"
```
At the same time, when trying to connect with an inappropriate driver, we get a well-known error message that only H2 contributors can change.

On the other hand, it may be worth shortening the text with which we determine the connection error:
```
General error: "The write format 1 is smaller than the supported format
```
instead of
```
General error: "The write format 1 is smaller than the supported format 2
```

server/bundles/io.cloudbeaver.service.security/src/io/cloudbeaver/service/security/db/H2Migrator.java
EmYar: Fixed

server/bundles/io.cloudbeaver.service.security/src/io/cloudbeaver/service/security/db/CBDatabase.java
serge-rider: DriverDescriptor  implements DBPDriver. It is not a JDBC driver but DBeaver driver.

driver != driver instance.

server/bundles/io.cloudbeaver.service.security/src/io/cloudbeaver/service/security/db/H2Migrator.java
serge-rider: We have debug level tuned ON by default for all applications.

server/bundles/io.cloudbeaver.service.security/src/io/cloudbeaver/service/security/db/H2Migrator.java
serge-rider: If users customize their database configuration in any maner then we don't need to do any migration, it is on users. Otherwise it becomes too unpredictable and too complicated.

Migration must upgrade only our preconfigured out-of-the-box database.

server/bundles/io.cloudbeaver.service.security/src/io/cloudbeaver/service/security/db/H2Migrator.java
EmYar: Fixed

server/bundles/io.cloudbeaver.service.security/src/io/cloudbeaver/service/security/db/H2Migrator.java
EmYar: Ok, implemented

server/bundles/io.cloudbeaver.service.security/src/io/cloudbeaver/service/security/db/CBDatabase.java
EmYar: Name reverted

server/bundles/io.cloudbeaver.service.security/src/io/cloudbeaver/service/security/db/H2Migrator.java
EmYar: The logging level of "No migration needed" message changed to TRACE

serge-rider (approved):

server/bundles/io.cloudbeaver.service.security/src/io/cloudbeaver/service/security/db/H2Migrator.java
```diff
@@ -0,0 +1,269 @@
...
+            log.error("Unable to restore old database file '" + workspacePaths.v1Paths.dbDataFile + "'");
+        }
+        databaseConfiguration.setUrl(oldUrl);
+        databaseConfiguration.setDriver(oldDriver);
+    }
+
+
+    private static class WorkspacePaths {
+
+        private static final String V1_DB_NAME = "cb.h2.dat";
```
alexander-skoblikov, line 215: the database name is set in the config and can be different

server/bundles/io.cloudbeaver.service.security/src/io/cloudbeaver/service/security/db/H2Migrator.java
```diff
@@ -0,0 +1,269 @@
...
+            statement.setString(1, filePath);
+            statement.setString(2, dbProperties.getProperty(DBConstants.DATA_SOURCE_PROPERTY_PASSWORD));
+            statement.execute();
+        }
+    }
+
+    private void updateConfig(@NotNull WorkspacePaths workspacePaths) {
+        if (!V2_DRIVER_NAME.equals(databaseConfiguration.getDriver())) {
+            log.info("Using database driver '" + V2_DRIVER_NAME + "' instead of '" + V1_DRIVER_NAME + "' from config");
+            databaseConfiguration.setDriver(V2_DRIVER_NAME);
```
alexander-skoblikov, line 173: do we store the new value in the server config?

PR #1542, opened 2023-03-13
CB-3240 if the navigator is disabled, do not mention it in placeholder

devnaumov:


Wroud (changes requested):
We need to change localization in TE only. We can't rely on panels emptiness in this case

PR #1544, opened 2023-03-14
CB-3187 close all dialogs on nav to settings

devnaumov:


Wroud (approved):
Please add jsdoc to `rejectAll` "you must avoid using this function, it can lead to unpredictable behaviour"

Add comment to `rejectAll` call place "We want to close dialog that triggered additional authentication dialog in case we going to settings"

PR #1556, opened 2023-03-20
CB-2885-TE-web-session-state-synchronisation-between-several-nodes

EmYar:


server/bundles/io.cloudbeaver.server/src/io/cloudbeaver/service/WebServiceBindingBase.java
```diff
@@ -121,14 +120,20 @@ protected static WebConnectionInfo getWebConnection(DataFetchingEnvironment env)
         return getWebConnection(getWebSession(env), getProjectReference(env), env.getArgument("connectionId"));
     }
 
+    /**
+     * Returns WebSession or null if the session was not found in the cache, the session could not be restored, or an error occurred
+     */
+    @Nullable
     public static WebSession findWebSession(DataFetchingEnvironment env) {
```
serge-rider, line 127: Are we sure we want to change this behavior?
getOrRestoreSession creates a new session while findWebSession mustn't do it (it was added specifically to avoid new session creation).

serge-rider: We have getWebSession instead. It may create new session.

server/bundles/io.cloudbeaver.server/src/io/cloudbeaver/service/session/WebSessionManager.java
```diff
@@ -139,19 +153,66 @@ public WebSession getWebSession(@NotNull HttpServletRequest request, @NotNull Ht
         return webSession;
     }
 
-    private void tryToRestorePreviousUserSession(WebSession webSession) {
-        try {
-            SMAuthInfo oldAuthInfo = webSession.getSecurityController().restoreUserSession(webSession.getSessionId());
-            if (oldAuthInfo == null) {
-                return;
+    /**
```
serge-rider, line 162: Please describe what is the puyrpose of session restoration,

serge-rider (changes requested):

server/bundles/io.cloudbeaver.server/src/io/cloudbeaver/service/WebServiceBindingBase.java
EmYar: `getOrRestoreSession` does not create a new session.
It takes the session from the cache, if not there, it tries to restore it. If the restore failed, null is returned.
When restoring a session, CBEmbeddedSecurityController#findTokenByAppSession checks that the session being restored has not expired, so we will not get an invalid session

server/bundles/io.cloudbeaver.server/src/io/cloudbeaver/service/session/WebSessionManager.java
EmYar: In case of replacing the backend node, we lose active valid sessions in the cache. With the next frontend request, we try to restore the lost session in order to continue working with it

webapp/packages/core-root/src/ServerNodeService.ts
```diff
@@ -0,0 +1,50 @@
...
+    this.sessionEventSource.onEvent<WsSocketConnectedEvent>(ServerEventId.CbSessionWebsocketConnected, data => {
+      this.applicationRunIdChanged(data.applicationRunId);
+    });
+  }
+
+  register(): void { }
+
+  load(): void { }
+
+  applicationRunIdChanged(applicationRunId: string): void {
```
devnaumov: I recommend marking any methods and fields that are not intended to be used outside of the class as private.

webapp/packages/core-root/src/ServerNodeService.ts
```diff
@@ -0,0 +1,50 @@
...
+    if (this.applicationRunId === '') {
```
devnaumov: I would suggest using "null" as the default value instead of an empty string or any other value if you want to clearly indicate that no value has been set. This will help avoid any confusion or unexpected behavior that might arise if an empty string or other default value is interpreted as a valid input.

webapp/packages/core-root/src/ServerNodeService.ts
```diff
@@ -0,0 +1,50 @@
...
+      this.applicationRunId = applicationRunId;
+      return;
+    }
+
+    if (this.applicationRunId === applicationRunId) {
+      return;
+    }
+
+    this.applicationRunId = applicationRunId;
+    this.graphQLService.blockRequests(new ServerNodeError('Server node changed'));
```
devnaumov, line 43: Better to translate if an error will be shown to the user in the UI

webapp/packages/core-root/src/ServerNodeService.ts
```diff
@@ -0,0 +1,50 @@
...
+
+import { Bootstrap, injectable } from '@cloudbeaver/core-di';
+import { Executor, IExecutor } from '@cloudbeaver/core-executor';
+import { GraphQLService, WsSocketConnectedEvent } from '@cloudbeaver/core-sdk';
+
+import { ServerNodeError } from './ServerNodeError';
+import { ServerEventId, SessionEventSource } from './SessionEventSource';
+
+@injectable()
+export class ServerNodeService extends Bootstrap {
```
devnaumov: As a suggestion, we could potentially refactor the code to move all logic related to the ServerNode into a separate plugin. This would allow for easier extension and modification of the logic in the future. In this case class should be renamed to the ServerNodeBootstrap

devnaumov (approved):

webapp/packages/core-localization/src/locales/en.ts
```diff
@@ -101,6 +101,10 @@ export default [
   ['app_root_session_expired_title', 'Session expired'],
   ['app_root_session_expired_message', 'Session has expired. Do you want to reload?'],
   ['app_root_session_expired_reload', 'Reload'],
+  ['app_root_server_node_changed_title', 'Server node changed'],
+  ['app_root_server_node_changed_message', 'Server node changed. You need to reload the application.'],
+  ['app_root_server_node_changed_reload', 'Reload'],
```
devnaumov: Reload and Cancel language tokens are propably already exists in the codebase.

devnaumov (approved):

webapp/packages/core-root/src/ServerNodeService.ts
```diff
@@ -0,0 +1,50 @@
...
+  applicationRunId = '';
```
Wroud: please set initial values in the constructor

webapp/packages/core-root/src/ServerNodeService.ts
```diff
@@ -0,0 +1,50 @@
...
+    super();
+    this.onApplicationRunIdChange = new Executor();
+    this.sessionEventSource.onEvent<WsSocketConnectedEvent>(ServerEventId.CbSessionWebsocketConnected, data => {
+      this.applicationRunIdChanged(data.applicationRunId);
+    });
+  }
+
+  register(): void { }
+
+  load(): void { }
```
Wroud: its better to use `extends Dependency` instead of `extends Bootstrap`
Dependency will be initialized on start

webapp/packages/core-root/src/ServerNodeService.ts
Wroud, line 43: we need to call new created method in `SessionEventSource` let's call it `disconnect`
`disconnect` method must set private property `disconnected` to `true`

Use this property in the `retryTimer`
```
this.retryTimer = interval(RETRY_INTERVAL)
      .pipe(
        filter(() => !this.sessionExpireService.expired && networkStateService.state && !this.disconnected)
      );
```

Wroud: please also check that front-end will not be trying to open websocket

For example on topic subscriptions:
Open / close "Log Viewer"

Wroud (changes requested):

server/bundles/io.cloudbeaver.server/src/io/cloudbeaver/service/WebServiceBindingBase.java
serge-rider: findWebSession mustn't restore session. It is used only in a few places specifically to avoid any session refresh/restore/etc. E.g. when we read error log. 

Changing its behaviour may break applicationm workflow. Why do we need in this ticker, how is it related to web sockets?

serge-rider (changes requested):

server/bundles/io.cloudbeaver.server/src/io/cloudbeaver/service/WebServiceBindingBase.java
EmYar: This change ensured that the session was restored when a `sessionState` request was made after a server restart.
It is not necessary to implement the ticket. Change rolled back

webapp/packages/core-root/src/ServerNodeService.ts
tfokina: I agree, fixed this

webapp/packages/core-root/src/ServerNodeService.ts
tfokina: Fixed this

webapp/packages/core-root/src/ServerNodeService.ts
tfokina: This message isn't displayed to the user

webapp/packages/core-root/src/ServerNodeService.ts
tfokina: Seems like it isn't necessary

webapp/packages/core-localization/src/locales/en.ts
tfokina: Updated this

webapp/packages/core-root/src/ServerNodeService.ts
tfokina: Fixed this

webapp/packages/core-root/src/ServerNodeService.ts
tfokina: Fixed this

webapp/packages/core-root/src/ServerNodeService.ts
tfokina: Thank you, applied changes

Wroud (approved):

server/bundles/io.cloudbeaver.server/src/io/cloudbeaver/service/session/WebSessionManager.java
```diff
@@ -144,19 +158,66 @@ public WebSession getWebSession(@NotNull HttpServletRequest request, @NotNull Ht
...
+            if (sessionMap.containsKey(sessionId)) {
+                var cachedWebSession = sessionMap.get(sessionId);
+                if (!(cachedWebSession instanceof WebSession)) {
+                    log.warn("Unexpected session type: " + cachedWebSession.getClass().getName());
+                    return null;
+                }
+                return (WebSession) cachedWebSession;
+            } else {
+                try {
+                    webSession = createWebSessionImpl(httpSession);
```
alexander-skoblikov: creating a web session is quite a fat entity, 
we can use only `sessionId` to know can we restore session or not, 
and if we can do it - only then create a web session

alexander-skoblikov (changes requested):

alexander-skoblikov (approved):

serge-rider (approved):

PR #1576, opened 2023-03-29
CB-3235-Send-event-about-session-expire

EmYar:


server/bundles/io.cloudbeaver.model/src/io/cloudbeaver/model/session/BaseWebSession.java
```diff
@@ -148,7 +149,13 @@ public synchronized WebUserContext getUserContext() {
     public void close() {
         super.close();
         synchronized (sessionEventHandlers) {
+            var sessionExpiredEvent = new WSSessionExpiredEvent();
             for (CBWebSessionEventHandler sessionEventHandler : sessionEventHandlers) {
+                try {
+                    sessionEventHandler.handeWebSessionEvent(sessionExpiredEvent);
```
alexander-skoblikov, line 161: Why we cant use `addSessionEvent` here?

EmYar: Because it is a system event.
It shouldn't be filtered. We want to send it right before the session is closed, within one lock. It is also not required to write an error to the session log, because it is planned to be closed - if there is a problem, it will be written to the general application log

PR #1698, opened 2023-05-25
CB-3434 include letter spacing computation

devnaumov:


webapp/packages/core-utils/src/TextTools.ts
```diff
@@ -12,6 +12,8 @@ interface IGetWidthOptions {
   text: string[];
 }
 
+const LETTER_SPACING = 0.4;
+
```
Wroud, line 16: you can set it as style?

webapp/packages/plugin-data-spreadsheet-new/src/DataGrid/useTableData.ts
```diff
@@ -66,16 +70,14 @@ export function useTableData(
...
-            return cell;
-          }
-          return rowStrings[i];
+        font: FONT,
+        text: columnNames.map((columnName, i) => {
+          const cellString = rowStrings[i] || '';
+          return columnName.length >= cellString.length ? columnName : cellString;
         }),
-      }).map(v => v + 16 + 32 + 20);
+      }).map(v => v + COLUMN_HEADER_CONTAINER_WIDTH + COLUMN_NAME_LEFT_MARGIN);
```
Wroud: offsets of column name and value are different, and offsets of column name are bigger.
we need to measure the size of the column's name and cell value separately and choose a bigger

Wroud (changes requested):

Wroud (approved):

PR #1755, opened 2023-06-12
CB-3513 remove leading slash from rm path

devnaumov:


webapp/packages/core-resource-manager/src/ResourceManagerResource.ts
```diff
@@ -261,6 +261,10 @@ export class ResourceManagerResource extends CachedTreeResource<RmResourceInfo,
 }
 
 export function getRmResourcePath(projectId: string, path?: string): string {
+  if (path) {
+    path = path.replace(/^\//, '');
+  }
+
   return createPath(projectId, path);
```
Wroud: probably want to fix `createPath` instead, try this case:
`createPath('a', '/b')`
the result must be `a/b` (we need to remove trailing `/` from the start and end of path parts after the first part:
`createPath('/a/', 'b')` -> `/a/b`

`createPath('/a/'. '/b/)` -> `/a/b`

`createPath('/a', '/b)` -> `/a/b`

Wroud (changes requested):

Wroud (approved):

PR #1803, opened 2023-07-05
CB-1652 add cross button to search fields

tfokina:


webapp/packages/core-blocks/src/ItemList/ItemListSearch.tsx
```diff
@@ -39,6 +39,15 @@ export const ItemListSearch: React.FC<IProps> = function ItemListSearch({ value,
...
 
+  const resetSearch = useCallback(() => {
+    if (value === undefined) {
+      setSearch('');
+    }
+    if (onChange) {
+      onChange('');
+    }
+  }, [value, onChange]);
+
```
Wroud: code duplication, you can change `changeHandler` so it will accept value instead of event, then you can call it in `onClick={() => changeHandler('')}`

webapp/packages/core-blocks/src/ItemList/ItemListSearch.tsx
```diff
@@ -67,9 +74,9 @@ export const ItemListSearch: React.FC<IProps> = function ItemListSearch({ value,
           onChange={changeHandler}
           {...use({ mod: 'surface' })}
         />
-        <search-button as="div" onClick={searchHandler}>
-          <ListSearchButton name="search" />
-        </search-button>
+        <action-button as="div">
+          {!onSearch && (value ?? search) ? <IconButton name="cross" onClick={resetSearch} /> : <IconButton name="search" onClick={searchHandler} />}
+        </action-button>
```
Wroud: why you use `onSearch` in condition? You need to:
```js
const inputValue = value ?? search;
...
{inputValue ? ... : ... }
```

webapp/packages/core-blocks/src/FormControls/Filter.tsx
```diff
@@ -106,10 +112,12 @@ export const Filter = observer<ControlledProps | ObjectsProps<any, any>>(functio
   const styles = useStyles(filterStyles, style, toggleMode && toggleModeButtonStyle);
   const [inputRef, ref] = useFocus<HTMLInputElement>({});
   const [toggled, setToggled] = useState(!toggleMode);
+  const [isFilled, setIsFilled] = useState(false);
 
   const filter = useCallback(
     (value: string | number, name?: string) => {
       value = String(value);
+      setIsFilled(!!value);
```
Wroud: You probably need to use `value` and if it's not empty or undefined then we can reset the value

Wroud (changes requested):

webapp/packages/core-blocks/src/ItemList/ItemListSearch.tsx
tfokina: Cause if we have onSearch handler, that means we can enter a string f.e. 'localhost' and click the search icon. We don't display cross button in this case.

Wroud (approved):

webapp/packages/core-blocks/src/PropertiesTable/styles.ts
```diff
@@ -57,14 +57,35 @@ export const PROPERTIES_TABLE_STYLES = css`
     height: 24px;
     pointer-events: none;
   }
+`;
 
-  ShadowInput {
+export const PROPERTIES_FIlTER_STYLES = css`
```
Wroud: `PROPERTIES_FIlTER_STYLES` -> `PROPERTIES_FILTER_STYLES `
l -> L

Wroud (approved):

PR #1806, opened 2023-07-06
CB-3549 fill connection config with defaults

devnaumov:


webapp/packages/plugin-connections/src/ConnectionForm/Options/useOptions.ts
```diff
@@ -32,13 +31,10 @@ export function useOptions(state: IConnectionFormState) {
...
       runInAction(() => {
         const {
-          prevName,
           state: { config, info, mode },
         } = refObject;
 
-        const isAutoFill = config.name === prevName || prevName === null;
-
-        if (mode === 'edit' || !isAutoFill) {
+        if (mode === 'edit') {
```
Wroud: it seems like it will reset the name to default every time driver is changed even custom name is set

webapp/packages/plugin-connections/src/ConnectionForm/Options/ConnectionOptionsTabService.ts
```diff
@@ -189,6 +216,7 @@ export class ConnectionOptionsTabService extends Bootstrap {
     }
 
     if (!state.info) {
+      await this.setDefaults(state);
       return;
     }
```
Wroud, line 221: it seems like it can reset the form on 'test connection' (when we creating new connection)

Wroud (changes requested):

Wroud (approved):

PR #1813, opened 2023-07-10
CB-3600 support files aggregation

devnaumov:


webapp/packages/core-localization/src/locales/en.ts
```diff
@@ -97,6 +97,7 @@ export default [
   ['ui_file_size_exceeds', 'Max file size exceeds. File size: {arg:size}. Max file size: {arg:maxSize}'],
   ['ui_upload_file', 'Upload file'],
   ['ui_upload_files', 'Upload files'],
+  ['ui_upload_files_duplicate_error', 'Files with the same name already exist'],
```
Wroud, line 100: not in use?

Wroud (changes requested):

Wroud (approved):

PR #1817, opened 2023-07-11
CB-3675 add project notifications

devnaumov:


Wroud:
we don't want to add this functionality to `StatusMessage`

PR #1825, opened 2023-07-18
CB-3720 fix: changed extensions order for editor

teunlao:


webapp/packages/plugin-codemirror6/src/Editor.tsx
```diff
@@ -22,7 +22,7 @@ export const Editor = observer<IEditorProps & IDefaultExtensions, IEditorRef>(
   forwardRef(function Editor({ lineNumbers, extensions, ...rest }, ref) {
     extensions = useCodemirrorExtensions(extensions);
     const defaultExtensions = useEditorDefaultExtensions({ lineNumbers });
-    extensions.set(...defaultExtensions);
+    extensions = new Map([defaultExtensions, ...extensions]);
```
Wroud: temporary fix, this behavior can be broken at any time
https://codemirror.net/examples/config/ #Precedence
https://codemirror.net/docs/ref/#view.gutter 

please try to use `Prec` to determine the gutters order
It will probably be enough to wrap `gutter({...})` with `Prec.lowest`

Wroud (changes requested):

Wroud (approved):

PR #1832, opened 2023-07-20
CB-3585 migrate itemList to css-modules

teunlao:


webapp/packages/core-blocks/src/ItemList/ItemList.m.css
```diff
@@ -1,106 +1,93 @@
...
-  }
-  list-item-icon {
+}
+.listItemIcon {
     display: flex;
     align-items: center;
     box-sizing: border-box;
     padding: 8px;
 
     & StaticImage {
```
Wroud: `.staticImage` ?

webapp/packages/core-blocks/src/ItemList/ItemList.m.css
```diff
@@ -1,106 +1,93 @@
...
+            box-sizing: border-box;
+            overflow: hidden;
+            border-radius: 2px;
+            height: 24px;
+            width: 24px;
+            margin: 0;
 
-        &[name='cross'] {
-          padding: 4px;
+            &[name='cross'] {
```
Wroud: not sure that it works, does `IconButton` passes `name` property to HTML element?

Wroud (changes requested):

webapp/packages/core-blocks/src/ItemList/ItemList.m.css
teunlao: you're right, fixed it.

webapp/packages/core-blocks/src/ItemList/ItemList.m.css
teunlao: yes, it really doesn't work, changed it to an additional class.
`name` prop is not passed to the HTML

Wroud (approved):

PR #1836, opened 2023-07-21
CB-3684 rename options in connection menu

teunlao:


Wroud:
please copy english localization to it and zh (replace)
and add russian translation

Wroud (approved):

PR #1846, opened 2023-07-25
CB-3476 feat: save text selection and cursor position for each tab

teunlao:


webapp/packages/plugin-sql-editor-new/src/SQLEditor/SQLCodeEditorPanel/SQLCodeEditorPanel.tsx
```diff
@@ -24,16 +24,38 @@ import { useSqlDialectAutocompletion } from '../useSqlDialectAutocompletion';
 import { useSqlDialectExtension } from '../useSqlDialectExtension';
 import style from './SQLCodeEditorPanel.m.css';
 import { useSQLCodeEditorPanel } from './useSQLCodeEditorPanel';
+import { EditorView } from '@codemirror/view';
 
+type ILocalSQLCodeEditorPanelState = {
+  selection: { from: number; to: number };
+};
```
Wroud: it's better to use `interface` instead or remove `I` prefix from the name

Wroud (approved):

webapp/packages/plugin-sql-editor-new/src/SQLEditor/SQLCodeEditorPanel/SQLCodeEditorPanel.tsx
teunlao: Replaced it with an `interface`.

PR #1854, opened 2023-07-26
CB-3727: load common fields of network handler by default

teunlao:


webapp/.gitignore
```diff
@@ -8,4 +8,6 @@ lib
 coverage
 ts-out
 npm-debug.log
-debug.log
\ No newline at end of file
+debug.log
+schema.graphql
+graphql.config.yml
```
teunlao, line 13: In `gitignore` I added files needed for graphql plugin to work in webstorm.

Wroud (approved):

PR #1868, opened 2023-08-02
CB-3628 clean key errors in ResourceError on refresh

devnaumov:


webapp/packages/core-sdk/src/Resource/ResourceError.ts
```diff
@@ -19,6 +19,7 @@ export class ResourceError extends LoadingError {
   ) {
     super(
       () => {
+        this.resource.cleanError(this.key);
```
Wroud, line 24: please add comment that this actually a bug inside CachedResource and this is work around

Wroud (approved):

PR #1869, opened 2023-08-02
CB-3632 do not preload connections if session update is in progress

devnaumov:


webapp/packages/plugin-object-viewer/src/ObjectViewerPanel.tsx
```diff
@@ -62,7 +62,7 @@ export const ObjectViewerPanel: TabHandlerPanelComponent<IObjectViewerTabState>
   const connected = getComputed(() => connection.data?.connected || false);
 
   const children = useResource(ObjectViewerPanel, NavTreeResource, objectId, {
-    active: connected,
+    active: !connection.resource.isSessionUpdate() && connected,
```
Wroud, line 68: is this also works?
`!connection.isOutdated() && connected`

PR #1901, opened 2023-08-11
CB-3753-add-field-hints-in-the-filter-field

tfokina:
Added a hook for hints that returns hints state.
Added separate component Hints.
The Menu component doesn't know where to bind without MenuButton component. So I added MenuButton component and made its width 0 (see Hints.tsx)


webapp/packages/core-ui/src/Hints/useHints.ts
```diff
@@ -0,0 +1,150 @@
...
+  readonly filteredHints: Hint[];
+  separatorLastIndex: number;
+  toggleHints: () => void;
+  getChangedValue: (hint: string) => string;
+}
+
+export function useHints(
+  menu: MenuStateReturn,
+  hints?: Hint[],
+  model?: IDatabaseDataModel<any, IDatabaseResultSet>,
```
devnaumov: Hook accepts model, it would better to move it closer to the model ( for example plugin-data-spreadsheet) or change the hook so it would be more universal

webapp/packages/core-ui/src/Hints/useHints.ts
```diff
@@ -0,0 +1,150 @@
...
+import { type IDatabaseDataModel, type IDatabaseResultSet, ResultSetViewAction } from '@cloudbeaver/plugin-data-viewer';
+
+export interface Hint {
+  key: string;
+  value: string;
+  icon?: string | React.ReactElement;
+  title?: string;
+}
+
+export interface HintsState {
```
devnaumov: please use this order of fields
1. readonly field
2. field
3. (): void

devnaumov (changes requested):

webapp/packages/core-ui/src/Hints/useHints.ts
```diff
@@ -0,0 +1,150 @@
...
+          hint =>
+            !filterValue.trim() ||
+            hint.value.toUpperCase().includes(filterValue.trim().toUpperCase()) ||
+            filterValue.toUpperCase().includes(hint.value.toUpperCase()),
+        );
+
+        return result;
+      },
+      get separatorLastIndex() {
+        let lastIndex = this.inputValue.trim().lastIndexOf(' ');
```
devnaumov: const lastIndex = this.inputValue.trim().lastIndexOf(' ');
return lastIndex === -1 ? 0 : lastIndex

webapp/packages/core-ui/src/Hints/useHints.ts
Wroud: better is:
1. fields
2. readonly fields
3. private fields (only for classes)
4. constructor (only for classes)
5. `has` or `is` or any other "flag" methods (like `hasKey`, `isReadonly`)
6. `get` methods
7. `set` methods
8. other methods
9. protected methods (only for classes)
10. private methods (only for classes)

webapp/packages/core-ui/src/Hints/useHints.ts
```diff
@@ -0,0 +1,150 @@
...
+ * Licensed under the Apache License, Version 2.0.
+ * you may not use this file except in compliance with the License.
+ */
+import { action, computed, observable } from 'mobx';
+import { useEffect, useRef } from 'react';
+import type { MenuStateReturn } from 'reakit';
+
+import { useObservableRef } from '@cloudbeaver/core-blocks';
+import type { SqlResultColumn } from '@cloudbeaver/core-sdk';
+import { type IDatabaseDataModel, type IDatabaseResultSet, ResultSetViewAction } from '@cloudbeaver/plugin-data-viewer';
```
Wroud: plugins can't be used in `core-*` packages

webapp/packages/core-ui/src/InlineEditor/InlineEditor.tsx
```diff
@@ -31,6 +33,7 @@ export interface InlineEditorProps extends Omit<React.InputHTMLAttributes<HTMLIn
   active?: boolean;
   loading?: boolean;
   style?: ComponentStyle;
+  hintsState?: HintsState;
```
Wroud: can't be part of this component, you need to track keys and value, you can track it outside of this component

webapp/packages/core-ui/src/Hints/useHints.ts
```diff
@@ -0,0 +1,150 @@
...
+
+export interface Hint {
```
Wroud: please use autocompletion instead of hint and make global exported names more specific like `AutocompletionComponentCompletion`

and use `I` prefix for global exported interfaces

webapp/packages/plugin-data-viewer/src/TableViewer/TableHeader/TableWhereFilter.tsx
```diff
@@ -26,6 +28,16 @@ export const TableWhereFilter: PlaceholderComponent<ITableHeaderPlaceholderProps
...
 
+  const menu = useMenuState({
+    placement: 'bottom-end',
+    gutter: 1,
+  });
+
+  const mapColumnsToHints = (columns: SqlResultColumn[]) =>
+    columns.map(column => ({ key: column.label || '', value: column.label || '', icon: column.icon || '' }));
+
+  const hintsState = useHints(menu, undefined, model, resultIndex, mapColumnsToHints);
```
Wroud: we can combine menu & hints hooks into one by now

Wroud (changes requested):

LonwoLonwo:
Not actual.

PR #1909, opened 2023-08-14
CB-3503 generate different token for the safari

devnaumov:


webapp/packages/plugin-connections/src/ConnectionForm/Options/Options.tsx
```diff
@@ -310,7 +311,7 @@ export const Options: TabContainerPanelComponent<IConnectionFormProps> = observe
                 <>
                   <Container wrap gap hideEmpty>
                     <ObjectPropertyInfoForm
-                      autofillToken="new-password"
+                      autofillToken={isSafari ? 'section-connection-authentication section-options' : 'new-password'}
```
Wroud, line 314: this will lead to password suggestion for every connection in safari

Wroud (approved):

PR #1920, opened 2023-08-16
CB-3778-bug-use-oracle-as-internal-db

yagudin10:


server/bundles/io.cloudbeaver.service.security/src/io/cloudbeaver/service/security/db/CBDatabase.java
```diff
@@ -534,12 +538,23 @@ private String getCurrentInstanceId() throws IOException {
...
+
     /**
      * Replaces all predefined prefixes in sql query.
      */
     @NotNull
-    public String normalizeTableNames(@NotNull String sql) {
-        return CommonUtils.normalizeTableNames(sql, databaseConfiguration.getSchema());
+    public String translateQuery(@NotNull String sql) {
+        var normalizedScript = CommonUtils.normalizeTableNames(sql, databaseConfiguration.getSchema());
+        DBPPreferenceStore prefStore = SQLQueryTranslator.getDefaultPreferenceStore();
```
serge-rider: Why we translate non-DDL queries?
It is really slow and also we shouldn't use non-standard SQL in select/update/delete.

serge-rider (changes requested):

server/bundles/io.cloudbeaver.service.security/src/io/cloudbeaver/service/security/db/CBDatabase.java
yagudin10: Because for some cases we can't use our scripts from SM controller (f.e. Oracle and SQL Server don't support offset and limit)

serge-rider (approved):

PR #1922, opened 2023-08-16
CB-3810

DenisSinelnikov:
Fixed display after add/update datasource for two or more (session/account) with same access datasources at the list of templates


alexander-skoblikov (approved):

server/bundles/io.cloudbeaver.model/src/io/cloudbeaver/model/rm/local/LocalResourceController.java
```diff
@@ -312,6 +312,9 @@ public void deleteProject(@NotNull String projectId) throws DBException {
                 log.debug("Deleting project '" + projectId + "'");
                 IOUtils.deleteDirectory(targetPath);
                 getSecurityController().deleteAllObjectPermissions(projectId, SMObjects.PROJECT);
+                synchronized (projectRegistries) {
+                    projectRegistries.remove(projectId);
```
fnasibov, line 316: I think it's better to use ConcurrentHashMap

PR #1942, opened 2023-08-23
CB-3602 fix: sql context management

Wroud:


devnaumov (approved):

webapp/packages/plugin-sql-editor/src/SqlEditor/useSqlEditor.ts
```diff
@@ -163,7 +163,9 @@ export function useSqlEditor(state: ISqlEditorTabState): ISQLEditorData {
 
               untracked(() => {
                 this.sqlDialectInfoService.loadSqlDialectInfo(key).then(async dialect => {
-                  await this.updateParserScriptsThrottle();
+                  try {
+                    await this.updateParserScriptsThrottle();
+                  } catch {}
```
devnaumov, line 170: Do we handle errors inside the parsing function somehow?

Wroud: fixed, now we will log parsing errors from backend within notifications

PR #1946, opened 2023-08-24
CB-3726: fix duplicated notifications

teunlao:


webapp/packages/plugin-connections/src/ConnectionAuthService.ts
```diff
@@ -44,19 +46,23 @@ export class ConnectionAuthService extends Dependency {
...
     } catch (exception: any) {
       this.notificationService.logException(exception);
       throw exception;
     }
   }
 
   async auth(key: IConnectionInfoParams, resetCredentials?: boolean): Promise<Connection | null> {
+    if (!this.networkStateService.state) {
+      throw new Error("Can't establish new connection in offline mode.");
+    }
```
Wroud: It's not true in some cases.
For example, if you run database and cloudbeaver on the same machine and this machine doesn't have internet access, then this condition will pass.

Wroud (changes requested):

webapp/packages/plugin-connections/src/ConnectionAuthService.ts
```diff
@@ -43,17 +43,12 @@ export class ConnectionAuthService extends Dependency {
...
-        return;
-      }
-      connection.connection = tempConnection;
-    } catch (exception: any) {
-      this.notificationService.logException(exception);
-      throw exception;
+    if (!tempConnection?.connected) {
+      return;
     }
+    connection.connection = tempConnection;
```
teunlao: I removed logException here, because it is first logged here and then goes higher and is logged one more time (loadNestedNodes).

Wroud: nice

Wroud (approved):

Wroud (approved):

Wroud (approved):

PR #1989, opened 2023-09-07
CB-3859 migrate table from reshadow to css modules

devnaumov:


Wroud (commented):
Why you haven't removed BASE_TABLE_STYLES?

Wroud (approved):

PR #1994, opened 2023-09-11
CB-3860 migrate dialog to css modules

devnaumov:


Wroud:
Please also update every component that uses `CommonDialogWrapper` in ce and ee repositories (we also have style duplication; try to reduce it)

Wroud (approved):

PR #2006, opened 2023-09-18
CB-3839. Added async error events messages for te

DenisSinelnikov:


server/bundles/io.cloudbeaver.model/src/io/cloudbeaver/model/rm/local/LocalResourceController.java
```diff
@@ -962,6 +964,19 @@ protected <T> T doFileReadOperation(String projectId, Path file, RMFileOperation
...
             } catch (Exception e) {
+                if (credentialsProvider.getActiveUserCredentials() != null) {
+                    WebAppUtils.getWebApplication().getEventController().addEvent(
+                        new WSErrorEvent(
+                            WSEventType.SESSION_LOG_MESSAGE_ADDED,
+                            credentialsProvider.getActiveUserCredentials().getSmSessionId(),
+                            credentialsProvider.getActiveUserCredentials().getUserId(),
+                            e.getMessage()));
+                } else {
+                    WebAppUtils.getWebApplication().getEventController().addEvent(
```
alexander-skoblikov: we don't need event in this case

PR #2010, opened 2023-09-21
CB-3456 refactor: user form

Wroud:


webapp/packages/core-blocks/src/FormControls/Form.tsx
```diff
@@ -0,0 +1,68 @@
...
+};
+
+export const Form = forwardRef<HTMLFormElement, FormDetailedProps>(function Form(
+  { context, disabled: disabledProp, disableEnterSubmit, focusFirstChild, children, onSubmit, onChange = () => {}, ...rest },
+  ref,
+) {
+  const [focusedRef] = useFocus<HTMLFormElement>({ focusFirstChild });
+  let [disabled, setDisabled] = useState(false);
+
+  disabled = disabled || disabledProp || false;
```
devnaumov: The disabled state is already initialized with false. If disabledProp is meant to be a prop that can override the initial state, it would be better to initialize disabled with disabledProp instead. Also better to change disabledProp name to isDisabled

webapp/packages/core-ui/src/Form/FormBaseService.ts
```diff
@@ -0,0 +1,89 @@
...
+    if (!validation.valid) {
+      ExecutorInterrupter.interrupt(contexts);
+    }
+
+    if (validation.messages.length > 0) {
+      const messages = validation.messages.map(message => this.localizationService.translate(message));
+      this.notificationService.notify(
+        {
+          title:
+            data.mode === FormMode.Edit ? 'connections_administration_connection_save_error' : 'connections_administration_connection_create_error',
```
devnaumov: Seems like we cant use connections locale token in the core-ui package

webapp/packages/core-view/src/locales/ru.ts
```diff
@@ -2,6 +2,7 @@ export default [
   ['core_view_layout', 'Упорядочить'],
   ['core_view_zoom_in', 'Приблизить'],
   ['core_view_zoom_out', 'Отдалить'],
+  ['core_view_action_create_label', 'Create'],
```
devnaumov: Создать?

devnaumov (changes requested):

webapp/packages/core-ui/src/Form/FormBaseService.ts
Wroud: why?

webapp/packages/core-blocks/src/FormControls/Form.tsx
Wroud: Having `disabled` and `isDisabled` properties in the same closure is confusing

devnaumov (approved):

PR #2022, opened 2023-09-25
CB-4052 await login properly

devnaumov:


webapp/packages/plugin-administration/src/locales/en.ts
```diff
@@ -42,7 +42,7 @@ export default [
   ['administration_configuration_wizard_configuration_custom_connections', 'Enable private connections'],
   [
     'administration_configuration_wizard_configuration_custom_connections_description',
-    'Allows users to create private connections. Otherwise, all new connections can be created from the administration part only',
+    'Allows users to create private connections. Otherwise, all new connections can be created by the administrator only',
```
Wroud: This is alright for EE / AWS / CE but can confuse TE because connection creation depends on users' project permissions.
It may be clear to keep only "Allows users to create private connections."

Wroud (approved):

PR #2033, opened 2023-10-02
CB-3454 ai

devnaumov:


webapp/packages/plugin-version-update-administration/src/PluginBootstrap.ts
```diff
@@ -23,7 +23,7 @@ export class PluginBootstrap extends Bootstrap {
     this.administrationItemService.create({
       name: 'version-update',
       type: AdministrationItemType.Administration,
-      order: 5,
+      order: 8,
```
Wroud: just remove `order`, so it will be set to `Number.MAX_SAFE_INTEGER` and always will be last element

Wroud (changes requested):

Wroud (approved):

Wroud (approved):

PR #2067, opened 2023-10-18
CB-4099. Create logic for save comfig and user for CB TE.

DenisSinelnikov:


server/bundles/io.cloudbeaver.model/src/io/cloudbeaver/model/session/WebAuthInfo.java
```diff
@@ -157,4 +157,8 @@ public String getUserDomain() {
     public String getUserName() {
         return user.getUserId();
     }
+
+    public String getAuthProviderConfigurationId() {
```
alexander-skoblikov: remove method

PR #2100, opened 2023-10-30
CB-3777 api for getting row count

yagudin10:


server/bundles/io.cloudbeaver.server/schema/service.sql.graphqls
```diff
@@ -356,4 +356,12 @@ extend type Mutation {
...
     asyncSqlExplainExecutionPlanResult(taskId: ID!): SQLExecutionPlan !
 
+    asyncSqlRowDataCount(
+        projectId: ID,
+        connectionId: ID!,
+        contextId: ID!,
+        resultsId: ID!
+    ): AsyncTaskInfo!
+
+    asyncSqlRowDataCountResult(taskId: ID!): Int
```
Wroud: please use string because js doesn't support long numbers

Wroud: also function must return not optional value or exception

Wroud: and maybe we need to rename it to `asyncResultSetRowDataCount` or something that lets us know that we can use it with any result and not only sql queries

webapp/packages/plugin-data-viewer/src/DatabaseDataModel/DatabaseDataSource.ts
```diff
@@ -26,6 +26,7 @@ export abstract class DatabaseDataSource<TOptions, TResult extends IDatabaseData
   results: TResult[];
   offset: number;
   count: number;
+  totalRowsCount: number | null;
```
Wroud: please rename to `totalCount` (because it can be documents and not rows)

webapp/packages/plugin-data-viewer/src/TableViewer/TableFooter/TableFooterRowCount.tsx
```diff
@@ -0,0 +1,96 @@
...
+
+      const { count } = await graphQLService.sdk.getSqlRowDataCountResult({ taskId: info.id });
+
+      if (count !== undefined) {
+        model.source.setTotalRowsCount(count);
+      }
+    } catch (exception: any) {
+      notificationService.logException(exception, 'data_viewer_total_rows_count_failed');
+    } finally {
+      setTask(null);
```
Wroud: this logic must be implemented in `IDatabaseDataSource` (probably we also need some kind of flag to check if it's implemented and supported)

also this action must be executed as blocking (DatabaseDataSource must be in loading state)

Wroud (changes requested):

server/bundles/io.cloudbeaver.server/schema/service.sql.graphqls
yagudin10: @Wroud  `sql` at the beginning is the name of module

yagudin10: @Wroud we return `Int` in every GQL functions where long is used in BE (f.e. FSFile#length or ServerConfig#sessionExpireTime)

Wroud: ok about `sql`
about `Int` than we need to change it in FSFile, but sessionExpireTime maybe can be kept as it is (if we sure that it's lower than 2^53 – 1)

Wroud: we can add new scalar like `Long` for that type of numbers

yagudin10: added

Wroud (approved):

PR #2137, opened 2023-11-16
CB-4220. Show real S3 message

DenisSinelnikov:


server/bundles/io.cloudbeaver.service.fs/src/io/cloudbeaver/service/fs/model/WebFSServlet.java
```diff
@@ -98,7 +98,7 @@ private void doPost(WebSession session, HttpServletRequest request, HttpServletR
                 }
             }
         } catch (Exception e) {
```
serge-rider, line 101: Why use e.getCause()? It may produce NPE.

Use both messages: "File upload filed: " + e.getMessage()

If you need original error use GeneralUtils.getRootCause

serge-rider (changes requested):

serge-rider (approved):

PR #2148, opened 2023-11-20
CB-4259 update vulnerable packages

devnaumov:
We also should remove reshadow dep to fix some vulnerabilities, but first we need to migrate all packages to the css modules


Wroud:
please update dependencies for EE repo also (including yarn.lock)

Wroud (approved):

PR #2154, opened 2023-11-21
CB-4281 send saveCredentials flag only if credentials exist

devnaumov:


Wroud:
probably we want to migrate DBAuthDialogController to hooks and use:
```
  let authModelId: string | null = null;

  if (connection.connectionInfo?.authNeeded || payload.resetCredentials) {
    authModelId = connection.connectionInfo?.authModel || driver?.defaultAuthModel || null;
  }
  ```
  
  `authModelId` to determine when we requested user to enter database credentials and send `config.credentials` and `config.saveCredentials` only in this case

Wroud (approved):

PR #2184, opened 2023-11-29
CB-4266 change notification display

sergeyteleshev:


webapp/packages/core-blocks/src/Snackbars/ActionSnackbar.m.css
```diff
@@ -0,0 +1,3 @@
+.wrapper {
+  height: 176px;
```
sergeyteleshev: i want to implement here a global variable into _vairables.scss in order to use it in all notifications, spreaded all over the packages. how do I import it in the packages in a true way?

Wroud (approved):

PR #2189, opened 2023-12-04
CB-4317 fix: require authentication to load connection origin details

Wroud:


webapp/packages/plugin-connections/src/ConnectionForm/OriginInfo/OriginInfo.tsx
```diff
@@ -38,12 +40,25 @@ const style = css`
...
 
-export const OriginInfo: TabContainerPanelComponent<IConnectionFormProps> = observer(function OriginInfo({ tabId, state: { info, resource } }) {
+export const OriginInfo: TabContainerPanelComponent<IConnectionFormProps> = observer(function OriginInfo({
+  tabId,
+  state: { info, resource, config },
+}) {
   const tab = useTab(tabId);
   const translate = useTranslate();
-  // const userInfoService = useService(UserInfoResource);
+  const userInfoService = useService(UserInfoResource);
```
sergeyteleshev: why is it useService instead of useResource?

sergeyteleshev (approved):

webapp/packages/plugin-connections/src/ConnectionForm/OriginInfo/OriginInfo.tsx
Wroud: fixed (actually here is both acceptable, but we can modify it to useResource to ensure that user is loaded and up to date in the component scope)

devnaumov (approved):

PR #2211, opened 2023-12-07
CB-4356 useForm does not switch tabs like insane after press enter

sergeyteleshev:


webapp/packages/core-blocks/src/useExecutor.ts
```diff
@@ -74,5 +74,9 @@ export function useExecutor<T>(options: IUseExecutorOptions<T>): void {
         executor.removeBefore(before);
       }
     };
-  }, [executor, props.handlers?.length, props.postHandlers?.length]);
+  }, [
+    executor, props.handlers?.length, props.postHandlers?.length, 
+    props.next, props.before, props.handlers, 
+    props.postHandlers,
```
Wroud: props.postHandlers need to be removed from dependencies:
we depend only on handlers count in array (handlers wrapped with function that use index from current array)

you can move out of useEffect:
```
    const next: IExecutorHandlersCollection<T> | undefined = props.next;
    const before: IExecutorHandlersCollection<T> | undefined = props.before;
```
and use this constants in the dependencies

webapp/packages/core-ui/src/Tabs/TabPanelList.tsx
```diff
@@ -42,7 +43,7 @@ export const TabPanelList = observer<Props>(function TabPanelList({ style, child
   const displayed = state.container.getDisplayed(state.props);
 
   return styled(styles)(
-    <>
+    <TabPanelValidationHandler>
```
Wroud: probably will be better to move it to TabsState (so we will be sure that all uses of TabPanel in TabsState have this context)

webapp/packages/core-ui/src/Tabs/TabPanelValidationHandler.tsx
```diff
@@ -0,0 +1,76 @@
...
+    };
+  }, []);
+
+  const value = useMemo(() => ({ addInvalidTab, invalidTabs }), []);
+
+  return (
+    <TabPanelValidationHandlerContext.Provider value={value}>
+      <div ref={containerRef}>
+        {children}
+      </div>
```
Wroud: seems like this div is extra (ref not in use)

webapp/packages/core-ui/src/Tabs/TabPanelValidationHandler.tsx
```diff
@@ -0,0 +1,76 @@
...
+
+export interface ITabPanelHandlerProps {
+  children: React.ReactNode;
+}
+
+export const TabPanelValidationHandler = observer(function TabPanelValidationHandler({ children }: ITabPanelHandlerProps) {
+  const state = useContext(TabsContext);
+  const containerRef = useRef<HTMLDivElement>(null);
+  const formContext = useContext(FormContext);
+  const selectedTab = state?.state.selectedId;
```
Wroud: this `selectedTab` will be captured on first render and then in `useEffect` we will get "old" state (if tabs will be changed)

Wroud: you can read this value in useEffect or add it to effect dependencies

webapp/packages/core-ui/src/Tabs/TabPanelValidationHandler.tsx
```diff
@@ -0,0 +1,76 @@
...
+      resetInvalidTabs();
+    }
+    
+    formContext?.parent?.ref?.removeEventListener('invalid', goNextTab, true);
+    formContext?.parent?.ref?.addEventListener('invalid', goNextTab, true);
+
+    return () => {
+      formContext?.parent?.ref?.removeEventListener('invalid', goNextTab, true);
+    };
+  }, []);
```
Wroud: We can simplify logic:
in each TabPanel notify context when panel is invalid

in this context schedule update when tab notify about validation
now check if selected tab invalid (if so skip tab switching)
else select first invalid tab

after tab switching execute validation again to display browser validation message (optionally we can track exact field to prevent full form validation)

webapp/packages/core-ui/src/Tabs/TabPanelValidationHandler.tsx
```diff
@@ -0,0 +1,76 @@
...
+
+  const value = useMemo(() => ({ addInvalidTab, invalidTabs }), []);
```
Wroud: you can use `useObjectRef` for such cases (it returns persist object that can be passed to contexts to prevent re-renders while fields of object mutated on every render)

Wroud (changes requested):

webapp/packages/core-ui/src/Tabs/TabPanel.tsx
```diff
@@ -9,52 +9,45 @@ import { observer } from 'mobx-react-lite';
...
     async function trackValidity(event: Event) {
-      if (selected || state?.context.get(DATA_CONTEXT_TABS_FORM_VALIDATION) !== null) {
-        return;
-      }
-      state?.context.set(DATA_CONTEXT_TABS_FORM_VALIDATION, tabId);
-      await state?.open(tabId);
+      tabPanelValidationHandlerContext?.addInvalidTab(tabId);
     }
 
+    panelRef.current?.removeEventListener('invalid', trackValidity, true);
```
devnaumov: Extra one? Seems like we remove listener in the useEffect callback function

webapp/packages/core-ui/src/Tabs/TabPanelValidationHandler.tsx
```diff
@@ -0,0 +1,76 @@
+import { observer } from 'mobx-react-lite';
```
devnaumov: please add license text

webapp/packages/core-ui/src/Tabs/TabPanelValidationHandlerContext.ts
```diff
@@ -0,0 +1,8 @@
+import { MutableRefObject, createContext } from 'react';
```
devnaumov: please add license text

devnaumov (dismissed):

Wroud (approved):

PR #2212, opened 2023-12-08
CB-4348 fix: parse sql script

Wroud:


webapp/packages/core-blocks/src/IconButton.tsx
```diff
@@ -20,6 +20,7 @@ import { useS } from './useS';
 import { useStyles } from './useStyles';
 
 interface Props {
+  tag?: 'button' | 'a' | 'div';
```
sergeyteleshev, line 23: Why do we need this? I believe it is better to divide functionality. For example if we want a button to behave like a link it should be something like this:

```
<Link>
    <Button>Button Text></Button>
</Link>
```

webapp/packages/plugin-sql-editor/src/SqlEditor/useSqlEditor.ts
```diff
@@ -228,7 +226,7 @@ export function useSqlEditor(state: ISqlEditorTabState): ISQLEditorData {
           const formatted = await this.sqlDialectInfoService.formatScript(this.dataSource.executionContext, script.query);
 
           this.onFormat.execute([script, formatted]);
-          this.setQuery(query.substring(0, script.begin) + formatted + query.substring(script.end));
+          this.setScript(query.substring(0, script.begin) + formatted + query.substring(script.end));
```
sergeyteleshev, line 229: Sometimes JS behaves crazy in this cases of concatenation if data is not as it expected to be. I would recommend to use `.concat` method to avoid bugs in future

sergeyteleshev (approved):

webapp/packages/core-blocks/src/IconButton.tsx
Wroud: according to HTML, it's a really bad idea to place tags like this: `<a><button></button></a>`
the main idea here is that sometimes we want to display button looking component as a link or div so we can place it in another component like a button or input label

(it's really rare cases usually)

PR #2225, opened 2023-12-13
CB-4386 unify feature naming

sergeyteleshev:


webapp/packages/core-localization/src/locales/ru.ts
```diff
@@ -12,7 +12,7 @@ export default [
   ['ui_processing_reload', 'Перезагрузить'],
   ['ui_processing_retry', 'Повторить'],
   ['ui_processing_ok', 'Принять'],
-  ['ui_processing_create', 'Создать'],
+  ['ui_processing_create', 'Готово'],
```
Wroud: we need to verify it in every place (not sure about it)

webapp/packages/plugin-resource-manager-scripts/src/locales/zh.ts
```diff
@@ -1,5 +1,5 @@
 export default [
-  ['plugin_resource_manager_scripts_action_enable_label', 'Show Scripts'],
+  ['plugin_resource_manager_scripts_action_enable_label', 'show Scripts'],
```
Wroud: seems strange

Wroud (changes requested):

webapp/packages/core-localization/src/locales/en.ts
```diff
@@ -14,7 +14,7 @@ export default [
   ['ui_processing_reload', 'Reload'],
   ['ui_processing_retry', 'Retry'],
   ['ui_processing_ok', 'Ok'],
-  ['ui_processing_create', 'Create'],
+  ['ui_processing_create', 'Finish'],
```
Wroud: in that case, we need to add a new key, `ui_processing_finish` to avoid confusing

Wroud (changes requested):

Wroud (approved):

PR #2235, opened 2023-12-15
CB-4392 support string type in useErrorDetails

devnaumov:


webapp/packages/core-notifications/src/NotificationsItem/NotificationsItem.tsx
```diff
@@ -7,24 +7,29 @@
...
     const Custom = notification.customComponent();
     return <Custom notification={notification} {...notification.extraProps} />;
   }
 
+  let closeDelay = 0;
+
+  if (notification.type !== ENotificationType.Error && !notification.persistent && notification.autoClose !== false) {
+    closeDelay = FADE_TIMEOUT;
+  }
+
```
sergeyteleshev, line 32: I suggest to add something like this to avoid component being overloaded with logic complicity

```
const calcDelay = (notification: Props['notification']) => {
  if (notification.type !== ENotificationType.Error && !notification.persistent && notification.autoClose !== false) {
    return FADE_TIMEOUT;
  }

  return 0;
};
```

sergeyteleshev (approved):

webapp/packages/core-blocks/src/useErrorDetails.ts
```diff
@@ -13,11 +13,14 @@ import { DetailsError } from '@cloudbeaver/core-sdk';
 import { errorOf, LoadingError } from '@cloudbeaver/core-utils';
 
 import { ErrorDetailsDialog } from './ErrorDetailsDialog/ErrorDetailsDialog';
+import { useTranslate } from './localization/useTranslate';
+
+type IError = Error | string | null;
```
Wroud: Please don't use I prefix for not interfaces

webapp/packages/core-blocks/src/useErrorDetails.ts
```diff
@@ -13,11 +13,14 @@ import { DetailsError } from '@cloudbeaver/core-sdk';
...
 
 interface IErrorDetailsHook {
   name?: string;
   message?: string;
-  error: Error | null;
+  error: IError;
```
Wroud: Not sure that we want this changes, this hook is for displaying details of exception, in that case this hook will do nothing, it's same as pass null as argument

devnaumov: We support it in ErrorDetailsDialog
`ErrorDetailsDialog: DialogComponent<Error | string, null>`
also we can pass `DetailsError` as extra prop

PR #2242, opened 2023-12-19
Cb 4280 refactor cb application configuration logic

yagudin10:


server/bundles/io.cloudbeaver.model/src/io/cloudbeaver/model/app/BaseServerConfigurationController.java
```diff
@@ -0,0 +1,31 @@
+/*
+ * DBeaver - Universal Database Manager
+ * Copyright (C) 2010-2023 DBeaver Corp
```
LonwoLonwo: 2024 is already here. Please update your files.

server/bundles/io.cloudbeaver.model/src/io/cloudbeaver/model/app/WebServerConfigurationController.java
```diff
@@ -0,0 +1,28 @@
+/*
+ * DBeaver - Universal Database Manager
+ * Copyright (C) 2010-2023 DBeaver Corp and others
```
LonwoLonwo: same here

server/bundles/io.cloudbeaver.model/src/io/cloudbeaver/model/app/BaseServerConfigurationController.java
alexander-skoblikov: use CE copyright

PR #2256, opened 2023-12-26
Cb 3555 view byte type data decoded

sergeyteleshev:
Снимок экрана 2023-12-27 в 14 10 17

I removed html/xml/json tabs for blobs. it seems redundant for me cause it basically shows the same thing as the "text" tab

I also decided not to go with BinaryValue entity and use TextValue entity due to the fact base64 and hex are a texts and there are less duplicated code


webapp/packages/plugin-data-viewer/src/locales/en.ts
```diff
@@ -28,6 +28,8 @@ export default [
   ['data_viewer_presentation_value_text_html_title', 'HTML'],
   ['data_viewer_presentation_value_text_xml_title', 'XML'],
   ['data_viewer_presentation_value_text_json_title', 'JSON'],
+  ['data_viewer_presentation_value_text_hex_title', 'HEX'],
```
devnaumov, line 31: No need to add these to locales

webapp/packages/plugin-data-viewer/src/ValuePanelPresentation/ImageValue/isImageValuePresentationAvailable.ts
```diff
@@ -0,0 +1,20 @@
+import { getMIME, isImageFormat, isValidUrl } from '@cloudbeaver/core-utils';
```
devnaumov, line 8: License text here and below

webapp/packages/plugin-data-viewer/src/ValuePanelPresentation/TextValue/TextValuePresentationBootstrap.ts
```diff
@@ -47,18 +48,36 @@ export class TextValuePresentationBootstrap extends Bootstrap {
...
       order: Number.MAX_SAFE_INTEGER,
       panel: () => React.Fragment,
+      isHidden: (_, context) => isBlobPresentationAvailable(context),
     });
     this.textValuePresentationService.add({
       key: 'application/json',
       name: 'data_viewer_presentation_value_text_json_title',
       order: Number.MAX_SAFE_INTEGER,
       panel: () => React.Fragment,
+      isHidden: (_, context) => isBlobPresentationAvailable(context),
```
devnaumov, line 65: image
image

webapp/packages/plugin-data-viewer/src/ValuePanelPresentation/TextValue/useAutoFormat.ts
```diff
@@ -25,6 +28,20 @@ export function useAutoFormat() {
           return value;
         }
       },
+      async formatBlob(type: string, value: IResultSetContentValue) {
```
devnaumov: seems like we dont "await" anything here

webapp/packages/plugin-data-viewer/src/ValuePanelPresentation/TextValue/useTextValue.ts
```diff
@@ -0,0 +1,45 @@
...
+export function useTextValue({ model, resultIndex, currentContentType }: Props) {
+  const format = model.source.getAction(resultIndex, ResultSetFormatAction);
+  const editor = model.source.getAction(resultIndex, ResultSetEditAction);
+  const selection = model.source.getAction(resultIndex, ResultSetSelectAction);
+  const focusCell = selection.getFocusedElement();
+  const formatter = useAutoFormat();
+  const [value, setValue] = useState<string>('');
+  const firstSelectedCell = selection.elements?.[0] ?? focusCell;
+  const autoFormat = !!firstSelectedCell && !editor.isElementEdited(firstSelectedCell);
+
```
devnaumov, line 30: Do wee need to use useEffect here?

```
 const candidate = format.get(firstSelectedCell);
 
  if (isResultSetContentValue(candidate)) {
      return formatter.formatBlob(currentContentType, candidate);
    }
    
 return formatter.format(currentContentType, format.getText(firstSelectedCell))
```

Also seems like a lot of initialization logic is here but its not in use until autoFormat is true

devnaumov (changes requested):

webapp/packages/plugin-data-viewer/src/ValuePanelPresentation/TextValue/useTextValue.ts
```diff
@@ -0,0 +1,45 @@
...
+import { isResultSetContentValue } from '../../DatabaseDataModel/Actions/ResultSet/isResultSetContentValue';
+import { ResultSetEditAction } from '../../DatabaseDataModel/Actions/ResultSet/ResultSetEditAction';
+import { ResultSetFormatAction } from '../../DatabaseDataModel/Actions/ResultSet/ResultSetFormatAction';
+import { ResultSetSelectAction } from '../../DatabaseDataModel/Actions/ResultSet/ResultSetSelectAction';
+import type { IDatabaseDataModel } from '../../DatabaseDataModel/IDatabaseDataModel';
+import type { IDatabaseResultSet } from '../../DatabaseDataModel/IDatabaseResultSet';
+import { useAutoFormat } from './useAutoFormat';
+
+type Props = {
```
Wroud: Please use interfaces

webapp/packages/core-utils/src/base64ToHex.ts
```diff
@@ -0,0 +1,17 @@
...
+
+export function base64ToHex(base64String: string): string {
+  const raw = atob(base64String);
+  let result = '';
+  for (let i = 0; i < raw.length; i++) {
+    const hex = raw.charCodeAt(i).toString(16);
+    result += hex.length === 2 ? hex : `0${hex}`;
+  }
+  return result.toUpperCase();
+}
```
Wroud, line 19: This function is synchronous and will lead to thread blocking and memory consumption. Please be careful with use cases for this function and verify that in the value panel it can handle max binary value without lags.

webapp/packages/plugin-data-viewer/src/ValuePanelPresentation/TextValue/useTextValue.ts
```diff
@@ -0,0 +1,45 @@
...
+    const candidate = format.get(firstSelectedCell);
+
+    if (isResultSetContentValue(candidate)) {
+      formatter.formatBlob(currentContentType, candidate).then(data => {
+        data && setValue(data);
+      });
+      return;
+    }
+
+    setValue(formatter.format(currentContentType, format.getText(firstSelectedCell)));
```
Wroud: This code is unstable, imagine that you selected big value and then before promise resolved you selected another one that is not content value. This will lead to set outdated data.

Also please don't use `a && b()` construction at all. Please don't use `a && b` constructions in not inline context.

webapp/packages/plugin-data-viewer/src/ValuePanelPresentation/TextValue/useTextValue.ts
```diff
@@ -0,0 +1,45 @@
...
+  const format = model.source.getAction(resultIndex, ResultSetFormatAction);
+  const editor = model.source.getAction(resultIndex, ResultSetEditAction);
+  const selection = model.source.getAction(resultIndex, ResultSetSelectAction);
+  const focusCell = selection.getFocusedElement();
+  const formatter = useAutoFormat();
+  const [value, setValue] = useState<string>('');
+  const firstSelectedCell = selection.elements?.[0] ?? focusCell;
+  const autoFormat = !!firstSelectedCell && !editor.isElementEdited(firstSelectedCell);
+
+  useEffect(() => {
```
Wroud: Here we probably want something like this:
```js
const cellValue = ...;
const formattedValue = useMemo(() => format(cellValue), [cellValue]);
```

It's main idea you can rewrite it. The main goal here is to sync cellValue and formattedValue and prevent cases when wrong formattedValue displayed. (For example we switched cells fast or cell value updated)

Wroud (changes requested):

webapp/packages/plugin-data-viewer/src/locales/en.ts
sergeyteleshev: why? all other tabs has locales, it should be consistent, I believe

webapp/packages/core-utils/src/base64ToHex.ts
sergeyteleshev: image

I tried do upload a video and the interface shows that only limited size is allowed. On this limited size feature works just fine. I added a comment to this helper so we use it carefully. Maybe when we need it we can make worker do the calculations in its own thread so the main thread do not freeze or do some other event loop optimizations

This is the great notice, by the way, thanks!

webapp/packages/plugin-data-viewer/src/ValuePanelPresentation/TextValue/TextValuePresentationBootstrap.ts
sergeyteleshev: fixed here:
https://github.com/dbeaver/cloudbeaver/commit/56333d524e9a280ba869ea4fd28984819dc81abf

webapp/packages/plugin-data-viewer/src/ValuePanelPresentation/TextValue/useTextValue.ts
sergeyteleshev: removed it, thanks!

webapp/packages/plugin-data-viewer/src/ValuePanelPresentation/TextValue/useTextValue.ts
sergeyteleshev: the hook is fully synchronous right now so it has more predictable behaviour excluding fast cell switching

webapp/packages/plugin-data-viewer/src/ValuePanelPresentation/TextValue/isTextValuePresentationAvailable.ts
```diff
@@ -0,0 +1,34 @@
...
+  const focusedElement = selection.getFocusedElement();
+
+  if (selection.elements.length > 0 || focusedElement) {
+    const view = context.model.source.getAction(context.resultIndex, ResultSetViewAction);
+
+    const firstSelectedCell = selection.elements[0] || focusedElement;
+
+    const cellValue = view.getCellValue(firstSelectedCell);
+
+    return isResultSetContentValue(cellValue) && cellValue.contentType === 'application/octet-stream' && Boolean(cellValue?.binary);
```
Wroud: Maybe we need to move contentType === 'application/octet-stream' check to separate function so we will know what we want to ensure in this case

Wroud (approved):

devnaumov (approved):

PR #2259, opened 2023-12-27
CB-4275 feat: app height set via js

sergeyteleshev:


Wroud (commented):
Please move this logic to separate hook and some tips:
1. Will be better to use ResizeObserver?
2. You can use useLayoutEffect without deps (not empty), root component won't be rerendered a lot and we don't want side effects related to react or browser behavior.
3. Set default value for css variable to 100vh

sergeyteleshev:
> 1. Will be better to use ResizeObserver?

During the test it appeared that resize observer only tracks element width, which is really weird. so I decided to go with `window.addEventListener()` due to the fact it works just fine

devnaumov (approved):

PR #2260, opened 2023-12-28
CB-4388 add credentials menu

devnaumov:


webapp/packages/plugin-authentication-administration/src/Administration/Users/UsersTable/UserCredentialsList.tsx
```diff
@@ -32,11 +36,43 @@ export const UserCredentials = observer<IUserCredentialsProps>(function UserCred
...
+      {visibleCredentials.map(origin => (
         <UserCredentials key={origin.type + origin.subType} origin={origin} />
       ))}
-    </Fragment>
+
+      {user.origins.length > MAX_VISIBLE_CREDENTIALS && (
+        <>
+          <MenuButton {...menu} className={s(styles, { menuButton: true })}>
+            <div className={s(styles, { hasMoreIndicator: true })}>
+              <span>+{user.origins.length - MAX_VISIBLE_CREDENTIALS}</span>
```
sergeyteleshev, line 58: Not a big fan of "+" for type casting. it is easy to forget or accidentally remove. More safe approach:

`<span>{Number(user.origins.length - MAX_VISIBLE_CREDENTIALS)}</span>`

webapp/packages/plugin-authentication-administration/src/Administration/Users/UsersTable/UserCredentialsList.tsx
```diff
@@ -32,11 +36,43 @@ export const UserCredentials = observer<IUserCredentialsProps>(function UserCred
...
+    gutter: 8,
+  });
+
+  const visibleCredentials = user.origins.slice(0, MAX_VISIBLE_CREDENTIALS);
+
+  return styled(BASE_DROPDOWN_STYLES)(
     <Fragment key="user-credentials-list">
-      {user.origins.map(origin => (
+      {visibleCredentials.map(origin => (
         <UserCredentials key={origin.type + origin.subType} origin={origin} />
```
sergeyteleshev: string concatenation can sometimes lead to unexpected results. it is better to use `concat` method. In this case subtype can be undefined so it will lead to something like: **"originundefined"**. And if those keys are duplicated react can rerender incorrect

I believe it is better to give a unique id here with some `generateId` helper

webapp/packages/plugin-authentication-administration/src/Administration/Users/UsersTable/UserCredentialsList.tsx
```diff
@@ -32,11 +36,43 @@ export const UserCredentials = observer<IUserCredentialsProps>(function UserCred
...
+            </div>
+          </MenuButton>
+
+          <Menu className={s(styles, { menu: true })} {...menu} modal>
+            {user.origins.slice(MAX_VISIBLE_CREDENTIALS).map(origin => {
+              const isLocal = origin.type === AUTH_PROVIDER_LOCAL_ID;
+              const title = isLocal ? translate('authentication_administration_user_local') : origin.displayName;
+
+              return (
+                <MenuItem key={origin.type + origin.subType} className={s(styles, { menuItem: true })} title={title}>
```
sergeyteleshev: same here with unique id

sergeyteleshev (approved):

PR #2261, opened 2023-12-28
CB-3972 safe way to get action check for the document

devnaumov:


sergeyteleshev (approved):

webapp/packages/plugin-data-viewer-result-set-grouping/src/DVResultSetGroupingPluginBootstrap.ts
```diff
@@ -89,8 +89,12 @@ export class DVResultSetGroupingPluginBootstrap extends Bootstrap {
...
               return true;
             }
-            const selectionAction = model.source.getAction(resultIndex, ResultSetSelectAction);
-            const dataAction = model.source.getAction(resultIndex, ResultSetDataAction);
+            const selectionAction = model.source.tryGetAction(resultIndex, ResultSetSelectAction);
+            const dataAction = model.source.tryGetAction(resultIndex, ResultSetDataAction);
+
+            if (!selectionAction || !dataAction) {
+              return true;
+            }
```
Wroud: it's will be better to:

1. in DATA_CONTEXT_DV_DDM remove second generic from `IDatabaseDataModel` so it will be set to `IDatabaseDataResult`
```
export const DATA_CONTEXT_DV_DDM = createDataContext<IDatabaseDataModel<IDatabaseDataOptions>>('data-viewer-database-data-model');
```

2. in `DVResultSetGroupingPluginBootstrap.ts`
don't register data presentation and menu creator if `model.source.getResult(resultIndex)?.dataFormat !== ResultDataFormat.Resultset`

Wroud (changes requested):

webapp/packages/plugin-data-viewer-result-set-grouping/src/DVResultSetGroupingPluginBootstrap.ts
Wroud: or in case if we want to enable this presentation for other data formats it's better to use `getActionImplementation(resultIndex, DatabaseSelectAction)` and `getActionImplementation(resultIndex, DatabaseDataAction)` or check data format and get proper action for this format (or just cast type because getActionImplementation will return proper action but type will be common)

Wroud (approved):

PR #2263, opened 2023-12-28
CB-3443 add format number util function

devnaumov:


webapp/packages/core-utils/src/formatNumber.ts
```diff
@@ -0,0 +1,24 @@
+/*
+ * CloudBeaver - Cloud Database Manager
+ * Copyright (C) 2020-2023 DBeaver Corp and others
+ *
+ * Licensed under the Apache License, Version 2.0.
+ * you may not use this file except in compliance with the License.
+ */
+
+export function formatNumber(n: number, d: number) {
```
sergeyteleshev, line 9: I suggest to add unit tests for this one cause it involves unobvious math calculations and we need to be sure it gives a correct output for happy pass and border cases

sergeyteleshev (approved):

PR #2268, opened 2024-01-03
Cb 4461 json and xml formatting doesnt work in value panel 🔥

sergeyteleshev:
I suggest to mark hotfix PRs with fire emoji so it can indicate it is hotfix and it should be reviewed with high priority in order to deliver fixes faster


webapp/packages/plugin-data-viewer/src/DatabaseDataModel/Actions/ResultSet/ResultSetViewAction.ts
```diff
@@ -127,6 +127,10 @@ export class ResultSetViewAction extends DatabaseDataAction<any, IDatabaseResult
   }
 
   getCellValue(cell: IResultSetElementKey): IResultSetValue {
+    if (!cell?.row) {
+      return null;
+    }
```
Wroud: but according to types this can't be true

Wroud (changes requested):

Wroud (approved):

PR #2269, opened 2024-01-03
CB-4309 add keep alive interval for data source

yagudin10:


webapp/packages/plugin-connections/src/ConnectionForm/Options/ProviderPropertiesForm.tsx
```diff
@@ -28,9 +30,17 @@ interface Props {
...
+
 export const ProviderPropertiesForm = observer<Props>(function ProviderPropertiesForm({ config, properties, disabled, readonly }) {
   const translate = useTranslate();
 
+  const keepAliveRef = useCustomInputValidation<string>(value => {
+    if (Number(value) > MAX_KEEP_ALIVE_INTERVAL_IN_SECONDS) {
+      return translate('connections_connection_keep_alive_max_value_error', undefined, { keepAliveInterval: MAX_KEEP_ALIVE_INTERVAL_IN_SECONDS });
+    }
+    return null;
+  });
```
Wroud: do we really need custom validation here? we set `max` value for input it should add browser default validation for this field

Wroud (changes requested):

Wroud (approved):

PR #2275, opened 2024-01-04
Cb 4467 utils unit tests

sergeyteleshev:


webapp/packages/core-utils/src/LoadingError.test.ts
```diff
@@ -0,0 +1,66 @@
...
+    const error = new LoadingError(() => {}, 'test');
+
+    expect(error instanceof Error).toBeTruthy();
+  });
+
+  it('should have name', () => {
+    const error = new LoadingError(() => {}, 'test');
+
+    expect(error.name).toBeTruthy();
+  });
```
Wroud: Error have a name by default, do we need to test it?

webapp/packages/core-utils/src/LoadingError.test.ts
```diff
@@ -0,0 +1,66 @@
...
+    const error = new LoadingError(onRefresh, 'test');
+
+    error.refresh();
+
+    expect(onRefresh).toHaveBeenCalledTimes(1);
+  });
+
+  it('should refresh cause', () => {
+    const onRefresh = jest.fn();
+    const cause = 'unit test';
```
Wroud: seems like cause must be `LoadingError` or name of test is not clear

webapp/packages/core-utils/src/copyToClipboard.test.ts
```diff
@@ -0,0 +1,26 @@
+import { copyToClipboard } from './copyToClipboard';
+
+document.execCommand = jest.fn();
```
Wroud: probably you need to reset this after test finished

webapp/packages/core-utils/src/copyToClipboard.test.ts
```diff
@@ -0,0 +1,26 @@
...
+
+describe('copyToClipboard', () => {
+  it('should copy data to clipboard', () => {
+    copyToClipboard('test');
+
+    expect(document.execCommand).toHaveBeenCalledWith('copy');
+  });
+
+  it('should focus on active element after copy', () => {
+    document.body.focus = jest.fn();
```
Wroud, line 15: and this

webapp/packages/core-utils/src/getMIME.test.ts
```diff
@@ -0,0 +1,27 @@
...
+  });
+
+  it('should return image/webp if binary starts with U', () => {
+    expect(getMIME('U')).toBe('image/webp');
+  });
+
+  it('should return null if binary starts with anything else', () => {
+    expect(getMIME('a')).toBe(null);
+  });
+});
```
Wroud: we need to generate small images in binary strings and test on real decoded images
(now test is reflects function implementation but not the actual expected behaviour)

webapp/packages/core-utils/src/getPathName.test.ts
```diff
@@ -0,0 +1,23 @@
+import { getPathName } from './getPathName';
+
+jest.mock('./getPathParts', () => ({
+  getPathParts: (path: string) => path.split('/'),
+}));
```
Wroud, line 5: not sure that we need to mock functions to isolate them, because in real world this is possibility that change in one function will break others but with mocks all other functions will pass tests

webapp/packages/core-utils/src/isSameDay.test.ts
```diff
@@ -0,0 +1,15 @@
+import { isSameDay } from './isSameDay';
+
+describe('isSameDay', () => {
+  it('should be same day', () => {
+    isSameDay(new Date(), new Date());
+    isSameDay(new Date(2020, 1, 1), new Date(2020, 1, 1));
```
Wroud: please add dates with different time (because now we can change isSameDay implementation so it will pass tests)

webapp/packages/core-utils/src/throttle.ts
```diff
@@ -41,6 +42,8 @@ export function throttle<T extends (...args: any[]) => void | Promise<void>>(f:
         }
       }, delay);
     }
+
+    return result;
```
Wroud: currently is better to not return any result here because we can't guarantee that we will return it (when throttled we just returning undefined)

so actually type of throttle is wrong it's return synchronous function that will return undefined all time

so we need to change function type OR return proper Promise that will be resolved or rejected

Wroud (changes requested):

webapp/packages/core-utils/src/copyToClipboard.test.ts
sergeyteleshev: After the test is executed, Jest automatically resets all the mock functions to their original implementations. I moved this part of the code to `BeforeAll()` just in case

webapp/packages/core-utils/src/copyToClipboard.test.ts
sergeyteleshev: this gonna be fine. jest does that automatically

webapp/packages/core-utils/src/getPathName.test.ts
sergeyteleshev: unit test supposed to test only logic of current unit. if the test has some dependencies it is have to be mocked. this is the main point of unit tests

for example we have 2 functions:
`getPathParts`
`getPathName` which has dep of `getPathParts`

1st case when we covered everything with unit tests and mocked `getPathParts` in `getPathName` tests:
- if some of the test is fallen, we can actually say which one has a bug
- we test only the logic of current unit not thinking about what happens in dependency function

2nd case where we don't use mocks:
- if `getPathParts` has a bug. it is gonna affect other tests. instead of 1 fallen you can have **N** fallen tests. Imagine this function was used in **50 tests**. You need to spend some time to find the reason why the test is fell.
- it is not unit tests cause we also test interaction between modules, so it is more like some kind of integration testing

so what we have on a plate here:
- 1st approach is gonna show us where you for sure have no bug and where you have it (once it covered with tests). its will show us the bonuses once we have our code base mostly covered with unit tests
- 2nd approach is cool when we don't have a lot of coverage and we need to indicate that something somewhere in the fog went wrong. through the time it is gonna be a huge pain and tests will be useless, which will lead us to rewrite it to the 1st case way

so my call is to use mocks due to the fact it can save us time to find a place where something went wrong, and through the time will be more helpful with each test being written

I also suggest to have a discussion with the full frontend team what is our main approach for unit testing

webapp/packages/core-utils/src/isSameDay.test.ts
sergeyteleshev: I added some extra checks for time: hour, minutes, seconds and miliseconds

webapp/packages/core-utils/src/getPathName.test.ts
Wroud: i see the point, let's leave it as is by now

Wroud (approved):

PR #2279, opened 2024-01-05
CB-4479 always satisfy hash function argument type

devnaumov:


sergeyteleshev:
I suggest to add a form validation on frontend side of the app. Seems like we need backend dev to send us required fields for login, password and all of the providers here

sergeyteleshev (approved):

Wroud (approved):

PR #2282, opened 2024-01-09
CB-4485 fix: bundle splitting

Wroud:


webapp/packages/plugin-data-spreadsheet-new/src/DataGrid/CellEditor.tsx
```diff
@@ -108,7 +110,7 @@ export const CellEditor = observer<Pick<RenderEditCellProps<IResultSetRowKey>, '
         onMouseDown={preventClick}
         onMouseUp={preventClick}
       >
-        {
+        {editorPortal &&
           createPortal(
```
sergeyteleshev, line 114: is there a chance that portal creates on each render of this component?

sergeyteleshev (approved):

webapp/packages/plugin-data-spreadsheet-new/src/DataGrid/CellEditor.tsx
sergeyteleshev: I suggest to move portal to `ref` and once `editorPortal` is not `null`, reassign this ref to `createPortal` value

devnaumov (approved):

PR #2286, opened 2024-01-09
Cb 4010 auth methods dont update when switching drivers

sergeyteleshev:


webapp/packages/core-blocks/src/ResourcesHooks/useResource.ts
```diff
@@ -161,7 +161,12 @@ export function useResource<
       propertiesRef.includes = includes;
     }
 
-    if (key === null || propertiesRef.key === null || !propertiesRef.resource.isIntersect(key, propertiesRef.key)) {
+    if (
+      key === null ||
+      propertiesRef.key === null ||
+      !propertiesRef.resource.isIntersect(key, propertiesRef.key) ||
+      !propertiesRef.resource.isEqual(key, propertiesRef.key)
```
sergeyteleshev: May be we should put a `shouldCompareEqual` flag in useResource arguments
and use it here like `shouldCompareEqual && !propertiesRef.resource.isEqual(key, propertiesRef.key)` if we dont want extra calculations for equality

webapp/packages/core-utils/src/isArraysEqual.ts
sergeyteleshev, line 1: I believe optimization here is has to be, because it is basic utility, which is gonna be used for big calculations for sure. It is already used in resources and I'm sure somewhere there are a lot of fields comes from backend side

Your call guys 👾

webapp/packages/core-blocks/src/ResourcesHooks/useResource.ts
Wroud: Seems like isIntersect is not suitable for useResource anymore because our data depends on the key, and in some cases, we want this data not just to include data that we need but also we don't want data to contain extra data

webapp/packages/core-resource/src/Resource/Resource.ts
```diff
@@ -107,6 +107,35 @@ export abstract class Resource<
...
+      return param.isEqual(second);
+    }
+
+    if (isResourceAlias(param) || isResourceAlias(second)) {
+      return false;
+    }
+
+    if (isResourceKeyList(param) && isResourceKeyList(second)) {
+      return isArraysEqual(param, second);
+    }
```
Wroud: this will work only for key list of primitives and not objects, we need to add `isEqual` to `ResourceKeyList`

webapp/packages/core-utils/src/isArraysEqual.test.ts
```diff
@@ -55,4 +55,12 @@ describe('Is array equals', () => {
...
     expect(isArraysEqual([{ a: 1 }], [{ a: 1 }], (a, b) => a.a === b.a, true)).toBe(true);
   });
+
+  test('should pass with array of objects (length > 1)', () => {
+    expect(isArraysEqual([{ b: 3 }, { a: 1 }], [{ a: 1 }, { b: 3 }])).toBe(true);
+  });
+
+  test('should pass with primitive and non primitive in array', () => {
+    expect(isArraysEqual([1, 1, { a: 1 }, 2], [2, { a: 1 }, 1, 1])).toBe(true);
+  });
```
Wroud: this function should not deeply compare objects only if third argument is passed

Wroud (changes requested):

Wroud (approved):

PR #2296, opened 2024-01-12
CB-4039 logout with context

alexander-skoblikov:


server/bundles/io.cloudbeaver.service.auth/schema/service.auth.graphqls
```diff
@@ -139,8 +144,13 @@ extend type Query {
     authUpdateStatus(authId: ID!, linkUser: Boolean): AuthInfo!
 
     # Logouts user. If provider not specified then all authorizations are revoked from session.
+    @deprecated
     authLogout(provider: ID, configuration: ID): Boolean
 
+    # Same as #authLogout, but returns additional information
+    @since(version: "23.3.3")
+    authLogoutExtended(provider: ID, configuration: ID): LogoutInfo!
```
Wroud, line 152: it's not clear what `Extended` suffix means especially if we will remove `authLogout`

webapp/packages/core-authentication/src/UserInfoResource.ts
```diff
@@ -152,11 +155,25 @@ export class UserInfoResource extends CachedDataResource<UserInfo | null, void,
...
+    if (redirectLinks.length) {
+      const oktaLink = authLogoutExtended?.redirectLinks[0];
+      const id = `okta-logout-id-${getUniqueId()}`;
+
+      this.windowsService.open(id, {
+        url: oktaLink,
+        width: 400,
+        height: 400,
+      });
+    }
```
Wroud: this logic is out of resource responsibilities

webapp/packages/core-utils/src/getUniqueId.ts
```diff
@@ -0,0 +1,10 @@
+function getNextIdGenerator() {
+  let id = 0;
+  return () => id++;
+}
+
+const getNextId = getNextIdGenerator();
+
+export function getUniqueId() {
+  return getNextId();
+}
```
Wroud: why not to use `uuid`?

webapp/packages/core-authentication/src/UserInfoResource.ts
Wroud: please change `logout` function in `AuthenticationService`

Wroud (changes requested):

server/bundles/io.cloudbeaver.service.auth/schema/service.auth.graphqls
sergeyteleshev: Alexander decided `authLogout` function to be for a while for back backward compatibility. so this is why we have new function with `Extended` suffix

Wroud: problem is that `Extended` means nothing this is not good suffix for api
`authLogoutInfo` sound better

webapp/packages/plugin-authentication/src/AuthenticationService.ts
```diff
@@ -106,7 +107,21 @@ export class AuthenticationService extends Bootstrap {
...
+      if (redirectLinks.length) {
+        const oktaLink = redirectLinks[0];
+        const id = `okta-logout-id-${uuid()}`;
+
+        this.windowsService.open(id, {
+          url: oktaLink,
+          width: 400,
+          height: 400,
+        });
+      }
```
Wroud: this logic must replace 
```ts
if (userAuthConfiguration?.configuration.signOutLink) {
      this.logoutConfiguration(userAuthConfiguration);
    }
```

we need to rewrite or remove it or change and maybe to change `logoutConfiguration`

Wroud (changes requested):

PR #2299, opened 2024-01-15
CB-4134 adjust cursor position

devnaumov:


webapp/packages/plugin-sql-editor-new/src/SQLEditor/SQLCodeEditorPanel/SQLCodeEditorPanel.tsx
```diff
@@ -112,6 +112,64 @@ export const SQLCodeEditorPanel: TabContainerPanelComponent<ISqlEditorModeProps>
...
+      },
+    ],
+  });
+
+  useExecutor({
+    executor: data.dataSource?.onSetScript,
+    postHandlers: [
+      function setScript({ script, source }) {
+        if (source === 'insert') {
+          // Note: Use setTimeout to ensure this runs after the current transaction is fully processed
```
sergeyteleshev: I don't really understand why `setTimeout` is here. Is this cause you want to onSetScript being executed first, and then executes what is inside `setTimeout`, right?

If so, I believe it is better to go with `next` props of `useExecutor`. It should be something like this in order to have more expected behavior from EventLoop execution perspective:

```
const setScriptExecutor = new SyncExecutor<ISetScriptData>();

  useExecutor({
    executor: data.dataSource?.onSetScript,
    next: setScriptExecutor,
  });

  useExecutor({
    executor: setScriptExecutor,
    handlers: [
      function setScript({ script, source }) {
        if (source === 'insert') {
          // Note: Use setTimeout to ensure this runs after the current transaction is fully processed
          const view = editorRef?.view;

          if (!view) {
            return;
          }

          const currentContent = view.state.doc.toString();
          const startPos = currentContent.lastIndexOf(script);

          if (startPos === -1) {
            return;
          }

          const endPos = startPos + script.length;

          view.dispatch({
            selection: {
              anchor: endPos,
              head: endPos,
            },
          });
        }
      },
    ],
  });
```

webapp/packages/plugin-sql-editor-new/src/SQLEditor/SQLCodeEditorPanel/SQLCodeEditorPanel.tsx
```diff
@@ -112,6 +112,64 @@ export const SQLCodeEditorPanel: TabContainerPanelComponent<ISqlEditorModeProps>
     ],
   });
 
+  useExecutor({
+    executor: data.onFormat,
```
sergeyteleshev: Same for this one as the example below

webapp/packages/core-blocks/src/FormControls/Textarea.tsx
```diff
@@ -79,13 +82,21 @@ export const Textarea: TextareaType = observer(function Textarea({
 
   const value = state ? state[name] : controlledValue;
 
+  useEffect(() => {
+    if (cursorInitiallyAtEnd && typeof value === 'string') {
+      const position = value.trim().length;
+      textareaRef.current?.setSelectionRange(position, position);
+    }
+  }, []);
```
Wroud: in case when we want to set position at the end we don't need to trim value because it will lead to unexpected behaviour like: `" |  f     ")` (cursor at position 1 after first space)

Wroud: Also, the current effect will not cover the following:
1. switch from `cursorInitiallyAtEnd=false` to `true`
2. value change from an empty string to not empty
3. possible cursor blink from the beginning to the end

webapp/packages/plugin-sql-editor/src/SqlDataSource/ISqlDataSource.ts
```diff
@@ -18,9 +18,11 @@ export interface ISqlDataSourceKey {
   readonly key: string;
 }
 
+export type ScriptDataSourceType = 'history' | 'insert';
```
Wroud: we don't want type here for extension purposes please revert related changes

webapp/packages/plugin-sql-editor/src/SqlEditor/useSqlEditor.ts
```diff
@@ -225,8 +225,8 @@ export function useSqlEditor(state: ISqlEditorTabState): ISQLEditorData {
           this.readonlyState = true;
           const formatted = await this.sqlDialectInfoService.formatScript(this.dataSource.executionContext, script.query);
 
-          this.onFormat.execute([script, formatted]);
           this.setScript(query.substring(0, script.begin) + formatted + query.substring(script.end));
+          this.onFormat.execute([script, formatted]);
```
Wroud: `onFormat` is not used. It will be better to remove it completely (and don't use it in `webapp/packages/plugin-sql-editor-new/src/SQLEditor/SQLCodeEditorPanel/SQLCodeEditorPanel.tsx`)

webapp/packages/plugin-sql-editor-new/src/SQLEditor/SQLCodeEditorPanel/SQLCodeEditorPanel.tsx
```diff
@@ -112,6 +112,64 @@ export const SQLCodeEditorPanel: TabContainerPanelComponent<ISqlEditorModeProps>
...
+          view.dispatch({
+            selection: {
+              anchor: selection.from,
+              head: selection.from,
+            },
+          });
+        });
+      },
+    ],
+  });
```
Wroud: seems like we want to achieve something like in `webapp/packages/plugin-sql-editor-new/src/SQLEditor/SQLCodeEditorPanel/SQLCodeEditorPanel.tsx` (line 88)

Wroud (changes requested):

webapp/packages/plugin-codemirror6/src/ReactCodemirror.tsx
```diff
@@ -136,14 +138,27 @@ export const ReactCodemirror = observer<IReactCodeMirrorProps, IEditorRef>(
...
+        view &&
+        memoizedCursor &&
+        (view.state.selection.main.anchor !== memoizedCursor.anchor || view.state.selection.main.head !== memoizedCursor.head)
+      ) {
+        view.dispatch({
+          selection: memoizedCursor,
+          annotations: [External.of(true)],
+        });
+      }
+    }, [view, memoizedCursor]);
```
Wroud: it's better to join this two because we want to be sure that they will be executed together you can conditionally set selection and/or changes together

webapp/packages/plugin-codemirror6/src/ReactCodemirror.tsx
```diff
@@ -136,14 +138,27 @@ export const ReactCodemirror = observer<IReactCodeMirrorProps, IEditorRef>(
...
+
     useLayoutEffect(() => {
```
Wroud, line 173: we need to add back sync event (onCursorChange) to update cursor value (like we have for value)

Wroud (changes requested):

webapp/packages/plugin-sql-editor-new/src/SQLEditor/SQLCodeEditorPanel/useSQLCodeEditorPanel.ts
```diff
@@ -35,6 +36,9 @@ export function useSQLCodeEditorPanel(data: ISQLEditorData, editor: IEditor) {
       onQueryChange(query: string) {
         this.data.setScript(query);
       },
+      onCursorChange(begin: number, end?: number) {
+        this.data.setCursor(begin, end);
+      },
       onUpdate(update: ViewUpdate) {
         const transactions = update.transactions.filter(t => t.selection !== undefined);
         const lastTransaction = transactions[transactions.length - 1] as Transaction | undefined;
```
Wroud: seems like we can remove onUpdate completely (it's now replaced with onQUeryChange and onCursorChange

Wroud (approved):

PR #2306, opened 2024-01-16
CB-3015: adds an fadeout animation & min close delay time to all noti…

sergeyteleshev:


Wroud:
i also can suggest to not to show notification right when some process started
it's better to not show it at all if process is fast
so lets assert that if some process takes longer then 300ms then we need to show some indication (notification or loader)

Wroud (approved):

webapp/packages/core-notifications/src/NotificationsItem/NotificationsItem.tsx
```diff
@@ -24,7 +26,7 @@ export const NotificationsItem = observer<Props>(function Notification({ notific
     return <Custom notification={notification} {...notification.extraProps} />;
   }
 
-  let closeDelay = 0;
+  let closeDelay = MIN_CLOSE_DELAY;
 
   if (notification.type !== ENotificationType.Error && !notification.persistent && notification.autoClose !== false) {
     closeDelay = FADE_TIMEOUT;
```
Wroud: it's possible that we now broke this behavior (that some notifications must be closed instantly and some not)
and we have two different delays 

now, all notifications will be closed after `MIN_CLOSE_DELAY` even when we want them to be open

check `useActivationDelay` in `Snackbar`

Wroud (changes requested):

webapp/packages/core-blocks/src/Snackbars/ProcessSnackbar.tsx
```diff
@@ -35,7 +36,14 @@ export const ProcessSnackbar: NotificationComponent<ProcessSnackbarProps> = obse
...
   const details = useErrorDetails(error);
-  const displayed = useStateDelay(notification.state.deleteDelay === 0, displayDelay);
+  const [delayState, setDelayState] = useState(false);
+  const displayed = useStateDelay(delayState, displayDelay);
+
+  useEffect(() => {
+    if (notification.state.deleteDelay === 0) {
+      setDelayState(true);
+    }
+  }, [notification.state.deleteDelay]);
```
Wroud:
```suggestion
  const [delayState, setDelayState] = useState(false);
  const displayedReal = notification.state.deleteDelay === 0;
  const displayed = useStateDelay(delayState, displayDelay);

  useEffect(() => {
    if (displayedReal) {
      setDelayState(true);
    }
  }, [displayedReal]);
```

Wroud (approved):

PR #2320, opened 2024-01-22
CB-4539 fix: value panel auto type detection

Wroud:


webapp/packages/plugin-data-viewer/src/ValuePanelPresentation/TextValue/TextValuePresentation.tsx
```diff
@@ -146,41 +149,71 @@ export const TextValuePresentation: TabContainerPanelComponent<IDataValuePanelPr
...
+          />
+          <Fill />
           {shouldShowPasteButton && (
+            <Container keepSize>
+              <Button disabled={model.isLoading()} onClick={pasteFullText}>
+                {translate('data_viewer_presentation_value_content_full_text_button')}
+              </Button>
+            </Container>
+          )}
+          <Container keepSize>
```
sergeyteleshev: this button now shows all the time. it should only appear for truncated texts only

sergeyteleshev: image

webapp/packages/plugin-data-viewer/src/ValuePanelPresentation/TextValue/TextValuePresentation.tsx
```diff
@@ -125,16 +128,16 @@ export const TextValuePresentation: TabContainerPanelComponent<IDataValuePanelPr
     const limit = bytesToSize(quotasService.getQuota('sqlBinaryPreviewMaxLength'));
     const canSave = firstSelectedCell && contentAction.isDownloadable(firstSelectedCell);
     const shouldShowPasteButton = isTextColumn && isTruncated;
-    const typeExtension = useMemo(() => getTypeExtension(state.currentContentType) ?? [], [state.currentContentType]);
+    const typeExtension = useMemo(() => getTypeExtension(contentType!) ?? [], [contentType]);
```
sergeyteleshev, line 131: should we go with `contentType ?? DEFAULT_CONTENT_TYPE`? if something went wrong and `contentType` is `null`, then interface probably will be broken at all

webapp/packages/plugin-data-viewer/src/ValuePanelPresentation/TextValue/TextValuePresentation.tsx
```diff
@@ -146,41 +149,71 @@ export const TextValuePresentation: TabContainerPanelComponent<IDataValuePanelPr
...
-          >
-            <TabList style={[BASE_TAB_STYLES, styles, UNDERLINE_TAB_STYLES]} />
-          </TabsState>
-        </actions>
-        <EditorLoader key={isReadonly ? '1' : '0'} value={textValue} readonly={isReadonly} extensions={extensions} onChange={handleChange} />
-        {isTruncated && <QuotaPlaceholder limit={limit} size={valueSize} />}
-        <div className={s(moduleStyle, { toolsContainer: true })}>
+      <Container vertical gap dense overflow>
+        <Container keepSize center overflow>
+          <Container keepSize>
```
sergeyteleshev, line 166: image

sergeyteleshev: I suppose something went wrong inside container component if we have to do this twice Container thing in order to get a correct layout

webapp/packages/plugin-data-viewer/src/ValuePanelPresentation/TextValue/TextValuePresentation.tsx
```diff
@@ -95,27 +72,53 @@ export const TextValuePresentation: TabContainerPanelComponent<IDataValuePanelPr
...
+      if (contentValue.contentType) {
+        switch (contentValue.contentType) {
+          case 'text/json':
+            autoContentType = 'application/json';
+            break;
+          case 'application/octet-stream':
+            autoContentType = 'application/octet-stream;type=base64';
+            autoLineWrapping = true;
+            break;
+          default:
```
sergeyteleshev, line 100: `'application/octet-stream;type=hex':` - should we consider this case?

sergeyteleshev (changes requested):

webapp/packages/plugin-data-viewer/src/ValuePanelPresentation/TextValue/TextValuePresentation.tsx
Wroud: no we shouldn't because this is typescript check and we will never have contentType = null at this point (only if we will broke something)

webapp/packages/plugin-data-viewer/src/ValuePanelPresentation/TextValue/TextValuePresentation.tsx
Wroud: we will detect tab automatically later, by now we just set base64 as default for all binary data

webapp/packages/plugin-data-viewer/src/ValuePanelPresentation/TextValue/TextValuePresentation.tsx
Wroud: fixed

webapp/packages/plugin-data-viewer/src/ValuePanelPresentation/TextValue/TextValuePresentation.tsx
Wroud: in this case it's because i don't want to add extra styles to `TabsList`
I suppose it's not always about problem in the `Container` but in matching other components together

sergeyteleshev (approved):

devnaumov (approved):

PR #2323, opened 2024-01-22
CB-4531 adds extra check for wether connection is connected or not in Table viewer

sergeyteleshev:


webapp/packages/plugin-object-viewer/src/ObjectViewerPanel.tsx
```diff
@@ -90,7 +94,7 @@ export const ObjectViewerPanel: TabHandlerPanelComponent<IObjectViewerTabState>
     }
   }, [connectionKey]);
 
-  if (connection.data && !connection.data.connected) {
```
sergeyteleshev, line 93: `connection.`data return `false` so backend fix is also required

PR #2325, opened 2024-01-23
CB-4023 sets equal window sizes on edit dataset

sergeyteleshev:


webapp/packages/plugin-sql-editor/src/SqlEditor/SqlEditor.tsx
```diff
@@ -81,7 +81,8 @@ export const SqlEditor = observer<ISqlEditorProps>(function SqlEditor({ state, c
       split.fixate('maximize', true);
     } else if (split.state.disable) {
       split.fixate('resize', false);
-      split.state.setSize(-1);
+      const size = Math.round(split.state.getContainerSize() / displayedEditors);
+      split.state.setSize(size);
```
Wroud: actually setting state to -1 must trigger auto size detecting.
displayedEditors - they will be displayed in the same box (not stack)

so we need to check why -1 not triggering auto size detection
or we need to change size calculation

Wroud (changes requested):

Wroud (approved):

PR #2329, opened 2024-01-24
#984 Changed to 23.10 ubuntu

ggxed:


pan3793:
May I know the background of this decision instead of using the Ubuntu LTS version?

PR #2334, opened 2024-01-25
CB-3703 save product configuration from ui

yagudin10:


webapp/packages/core-administration/src/AdministrationSettingsService.ts
```diff
@@ -7,19 +7,20 @@
  */
 import { injectable } from '@cloudbeaver/core-di';
 import { PluginManagerService, PluginSettings } from '@cloudbeaver/core-plugin';
+import { schema } from '@cloudbeaver/core-utils';
 
-const defaultSettings = {
-  baseFeatures: [] as string[],
-};
+const settingsSchema = schema.object({
+  baseFeatures: schema.array(schema.string()).default([]),
```
sergeyteleshev, line 13: should we use array of enums? so it is more typescript friendly

webapp/packages/core-blocks/src/FormControls/Checkboxes/FieldCheckbox.tsx
```diff
@@ -31,7 +32,7 @@ export const FieldCheckbox: CheckboxType = observer(function FieldCheckbox({
 
   return (
     <Field {...layoutProps} className={s(styles, { field: true }, className)}>
-      <Checkbox {...(checkboxProps as CheckboxBaseProps & ICheckboxControlledProps)} className={s(styles, { checkbox: true })} />
+      <Checkbox {...(checkboxProps as any)} className={s(styles, { checkbox: true })} />
```
sergeyteleshev, line 35: It would be nice to get rid of this any

webapp/packages/core-blocks/src/FormControls/useCustomInputValidation.ts
```diff
@@ -16,6 +16,31 @@ export function useCustomInputValidation<T = void>(validation: (value: T) => str
...
+    }
+
+    const valid = element.validity.valid;
+    const result = validation(value);
+
+    try {
+      if (typeof result === 'string') {
+        element.setCustomValidity(result);
+        return false;
+      } else {
```
sergeyteleshev: no need for this else you can just go with 

```
element.setCustomValidity('');
return true;
```

webapp/packages/core-blocks/src/FormControls/useCustomInputValidation.ts
```diff
@@ -16,6 +16,31 @@ export function useCustomInputValidation<T = void>(validation: (value: T) => str
...
+
+    if (element instanceof HTMLInputElement) {
+      value = element.value as unknown as T;
+    }
+
+    const valid = element.validity.valid;
+    const result = validation(value);
+
+    try {
+      if (typeof result === 'string') {
```
sergeyteleshev, line 32: what is expected behavior in case where result is an empty string? feels like in this case we should go in else statement

so I suggest to add extra check

`if (typeof result === 'string' && result)`

webapp/packages/core-blocks/src/FormControls/useCustomInputValidation.ts
```diff
@@ -48,14 +62,29 @@ export function useCustomInputValidation<T = void>(validation: (value: T) => str
...
+        validate(target);
+      }
+    }
+
+    function handleBlur(event: Event) {
+      const target = event.target as HTMLInputElement;
+      if (target.value === '') {
+        return;
+      }
+      if (target.validity.valid === true) {
```
sergeyteleshev, line 75: why do we validate it if it is already valid?
probably, this should be
`target.validity.valid === true`

webapp/packages/core-blocks/src/ToolsPanel/ToolsAction.tsx
```diff
@@ -20,11 +21,24 @@ interface Props extends ButtonHTMLAttributes<any> {
...
 
-export const ToolsAction: React.FC<Props> = function ToolsAction({ icon, viewBox, loading, children, className, ...rest }) {
+export const ToolsAction: React.FC<Props> = function ToolsAction({ icon, viewBox, disabled, loading, children, className, onClick, ...rest }) {
   const styles = useS(style);
+  const [loadingState, setLoadingState] = useState(false);
+
+  async function handleClick(event: React.MouseEvent<HTMLButtonElement>) {
+    try {
+      setLoadingState(true);
+      await onClick?.(event);
```
sergeyteleshev, line 32: `onClick` is not async. we can delete `await` and `async` a little bit above

webapp/packages/core-localization/src/LocalizationService.ts
```diff
@@ -187,23 +167,44 @@ export class LocalizationService extends Bootstrap {
...
-      for (const provider of this.localeProviders) {
-        for (const [key, value] of await provider(localeKey)) {
-          locale.set(key, value);
-        }
+    for (const provider of this.localeProviders) {
+      for (const [key, value] of await provider(localeKey)) {
+        locale.set(key, value);
       }
-      this.localeMap.set(localeKey, locale);
-    } catch (error: any) {
```
sergeyteleshev, line 199: I think this was not a bad idea. and we should left this catch just in case to know wether something went wrong

webapp/packages/core-plugin/src/SettingsManager/SettingsGroup.ts
```diff
@@ -0,0 +1,35 @@
...
+
+export class SettingsGroup {
+  readonly id: string;
+  readonly subGroups: SettingsGroup[];
+
+  protected groups: Map<string, SettingsGroup>;
+
+  constructor(readonly name: string, readonly parent?: SettingsGroup) {
+    this.id = uuid();
+    this.subGroups = [];
```
sergeyteleshev: is strict order here required? may be we can go with Set (if not) or even create an OrderedSet like you did with OrderedMap (if yes). I just want this staff to be immutable so we don't mess up the settings and it has unique tabs and stuff

webapp/packages/core-plugin/src/SettingsManager/SettingsManagerService.ts
```diff
@@ -0,0 +1,50 @@
...
+  schema: schema.AnyZodObject;
+  settingsGetter: SettingsDescriptionGetter<T>;
+}
+
+@injectable()
+export class SettingsManagerService {
+  settings: ScopeSettingsItem[];
+
+  constructor() {
+    this.settings = [];
```
sergeyteleshev: same here for `Set` if you think it is a valuable idea

webapp/packages/core-server-localization/src/ServerLocalizationService.ts
```diff
@@ -0,0 +1,32 @@
...
+
+    if (supportedLanguages) {
+      this.localizationService.setSupportedLanguages(
+        supportedLanguages.map(lang => ({
+          isoCode: lang.isoCode,
+          displayName: lang.displayName ?? lang.nativeName ?? lang.isoCode,
+        })),
+      );
+    }
+  }
```
sergeyteleshev, line 31: Should supported languages always be appeared? If so, I suggest to add notification here. so if there are no supported languages, we can find a bug once it appeared

webapp/packages/plugin-settings-panel/src/SettingsPanel/SettingsList.tsx
```diff
@@ -0,0 +1,88 @@
...
+
+    const group = ROOT_SETTINGS_GROUP.get(groupId)!;
+    const groupSettings = settings.get(group);
+
+    list.push({ group, settings: groupSettings || [] });
+  }
+
+  useEffect(() => {
+    if (ref.current) {
+      ref.current.addEventListener('scroll', e => {
```
sergeyteleshev: please, add `removeEventListener`

webapp/packages/plugin-settings-panel/src/SettingsPanel/SettingsList.tsx
```diff
@@ -0,0 +1,88 @@
...
+    const groupId = groups[0];
+    groups.splice(0, 1, ...treeData.getChildren(groupId));
+
+    const group = ROOT_SETTINGS_GROUP.get(groupId)!;
+    const groupSettings = settings.get(group);
+
+    list.push({ group, settings: groupSettings || [] });
+  }
+
+  useEffect(() => {
```
sergeyteleshev: I would like to see this logic moved to the hook so it is easier to read this component

sergeyteleshev:
also do `yarn install` for EE to update lock files with new lib

webapp/packages/core-plugin/src/PluginSettings.ts
```diff
@@ -5,44 +5,62 @@
...
+  getDefaultValue<TKey extends keyof schema.infer<TSchema>>(key: TKey): schema.infer<TSchema>[TKey] {
+    const value = this.source.getDefaultValue(this.scopedKey(key));
+    const schema = this.schema.shape[key as any];
+    const result = schema.safeParse(value);
+
+    if (result.success) {
+      return result.data;
     }
 
+    return schema.parse(undefined);
```
devnaumov, line 35: Why do we want to parse undefined here?

webapp/packages/core-plugin/src/SettingsManager/ISettingDescription.ts
```diff
@@ -0,0 +1,34 @@
...
+ * Licensed under the Apache License, Version 2.0.
+ * you may not use this file except in compliance with the License.
+ */
+import type { schema } from '@cloudbeaver/core-utils';
+
+import type { ESettingsValueType } from './ESettingsValueType';
+import type { SettingsGroup } from './SettingsGroup';
+
+export interface ISettingOptions {
+  id: string | number;
```
devnaumov: maybe we want to use one type for "id"?

webapp/packages/core-root/src/ServerSettingsService.ts
```diff
@@ -5,16 +5,112 @@
...
+  clear(): void {
+    this.settings.clear();
+    this.changes.clear();
+  }
+
+  async save() {
+    const changes = {};
+
+    for (const [key, value] of this.changes) {
+      setByPath(changes, key, value);
```
devnaumov, line 91: Probably we do not want to mutate object from util function

webapp/packages/core-settings-localization/src/LocalizationSettingsManagerService.ts
```diff
@@ -0,0 +1,42 @@
...
+    this.settingsManagerService.registerSettings(this.settingsLocalizationService.pluginSettings, () => [
+      {
+        group: INTERFACE_SETTINGS_GROUP,
+        key: 'defaultLanguage',
+        name: 'Default language',
+        type: ESettingsValueType.Select,
+        options: this.localizationService.supportedLanguages.map(language => ({
+          id: language.isoCode,
+          name: language.displayName,
+        })) || [{ id: 'en', name: 'English' }],
```
devnaumov: Seems supportedLanguages are [] by default so the code will never fallback to the default object

```
const defaultOptions = [{ id: 'en', name: 'English' }];
const options = this.localizationService.supportedLanguages.length > 0 ? this.localizationService.supportedLanguages.map(language => ({
          id: language.isoCode,
          name: language.displayName,
        })) : defaultOptions
```

webapp/packages/plugin-settings-administration/src/SettingsAdministration.tsx
```diff
@@ -0,0 +1,63 @@
...
+  });
+
+  function handleReset() {
+    serverSettingsService.resetChanges();
+  }
+
+  const changed = serverSettingsService.isChanged;
+
+  return (
+    <Form context={form} style={{ height: '100%', display: 'flex', flexDirection: 'column' }}>
```
devnaumov: Could you move inline styles to the css module?

webapp/packages/core-administration/src/AdministrationSettingsService.ts
Wroud: features is extendable entity, so we can't declare all possible values

webapp/packages/core-blocks/src/FormControls/Checkboxes/FieldCheckbox.tsx
Wroud: we need to refactor Checkbox and FieldCheckbox for that a bit :C

webapp/packages/core-plugin/src/SettingsManager/ISettingDescription.ts
```diff
@@ -0,0 +1,34 @@
...
+  name: string;
+}
+
+export interface ISettingDescription<T = object> {
+  key: keyof T;
+
+  type: ESettingsValueType;
+
+  group: SettingsGroup;
+  name: string;
```
devnaumov: TLocalizationToken instead of string for "name" and "description"

webapp/packages/core-blocks/src/FormControls/useCustomInputValidation.ts
Wroud: 1. we want to validate on blur
2. browser validation focus first invalid field (so if it's invalid it will be impossible to focus other element)

webapp/packages/plugin-settings-panel/src/SettingsPanel/SettingsGroups/SettingsGroups.tsx
```diff
@@ -0,0 +1,29 @@
...
+  treeData: ITreeData;
+}
+
+export const SettingsGroups = observer<Props>(function SettingsGroups({ treeData }) {
+  function getNodeHeight(id: string) {
+    return 24;
+  }
+
+  function handleClick(id: string) {
+    document.querySelector('#' + getSettingGroupId(id))?.scrollIntoView();
```
devnaumov, line 25: Can we do it in react way using ref?

webapp/packages/plugin-settings-panel/src/SettingsPanel/SettingsGroups/SettingsGroups.tsx
```diff
@@ -0,0 +1,29 @@
...
+import { getSettingGroupId } from '../getSettingGroupId';
+import { groupNodeRenderer } from './groupNodeRenderer';
+
+interface Props {
+  treeData: ITreeData;
+}
+
+export const SettingsGroups = observer<Props>(function SettingsGroups({ treeData }) {
+  function getNodeHeight(id: string) {
+    return 24;
```
devnaumov, line 21: const NODE_HEIGHT= 24;

devnaumov (changes requested):

webapp/packages/core-blocks/src/FormControls/useCustomInputValidation.ts
Wroud: empty string will be treated as invalid field but with no message = no message must be treated as a bug (validation function must not return empty strings)

webapp/packages/core-blocks/src/FormControls/useCustomInputValidation.ts
Wroud: fixed

webapp/packages/core-blocks/src/ToolsPanel/ToolsAction.tsx
Wroud: onClick now can be asynchronous function

webapp/packages/core-plugin/src/PluginSettings.ts
Wroud: to get default value from schema (there no better way to do it)

webapp/packages/core-plugin/src/SettingsManager/ISettingDescription.ts
Wroud: agree

webapp/packages/core-plugin/src/PluginSettings.ts
devnaumov: what about schema.shape[name].default?

webapp/packages/core-plugin/src/SettingsManager/SettingsGroup.ts
Wroud: added better types

webapp/packages/core-plugin/src/SettingsManager/SettingsManagerService.ts
Wroud: added better typing

webapp/packages/core-root/src/ServerSettingsService.ts
Wroud: in this case it's expected, function accepts `object`, `key` and `value` and it will mutate object
we don't need immutable object here

webapp/packages/core-server-localization/src/ServerLocalizationService.ts
Wroud: we can find it by trying to switch to another language (we will see only english option)
actually we don't want to display any notifications to users because we will try to fall back to english and give users access to application
they will notice that language is not as should be in case they use other langoages than english

webapp/packages/core-settings-localization/src/LocalizationSettingsManagerService.ts
Wroud: fixed in other way, now `this.localizationService.supportedLanguages` must never be empty

webapp/packages/plugin-settings-panel/src/SettingsPanel/SettingsGroups/SettingsGroups.tsx
Wroud: there is no need (used once, keep simple)

webapp/packages/plugin-settings-panel/src/SettingsPanel/SettingsGroups/SettingsGroups.tsx
Wroud: we can but it will be rocketsciense

webapp/packages/core-plugin/src/PluginSettings.ts
Wroud: it's same as schema.string()`.default('ass')` (shape returns schema object) we can't get default value this way

webapp/packages/plugin-settings-panel/src/SettingsPanel/SettingsList.tsx
Wroud: fixed

webapp/packages/plugin-settings-panel/src/SettingsPanel/SettingsList.tsx
Wroud: fixed

webapp/packages/plugin-settings-administration/src/SettingsAdministration.tsx
Wroud: fixed

sergeyteleshev (approved):

devnaumov (approved):

PR #2335, opened 2024-01-25
Cb 4429 change divider behavior

sergeyteleshev:


webapp/packages/core-blocks/src/Split/SplitControls.tsx
```diff
@@ -11,15 +11,39 @@ import { useS } from '../useS';
...
+// и учитывать эти значения при определении состояние mode
+// 2. сделать геттер на реф дива который рендерит Split (его контейнер) - splitRef (protected)
+
+// setSize, getMainStyleSize
+// либа должна коллапсить не до 0 а до state.minSize
+// либа должна коллапсить не до 100% если maxSize > 0, то до state.maxSize
+// либа должна коллапсить не до 100% если maxSize < 0, то до calc(100% - maxSize in px)
+
+// финальная рыба
+// 1. дофиксить сплит контролс так как он не учитывает minSize/maxSize
```
sergeyteleshev: Left it so you have notes to check our plan. Will delete it before the merge

webapp/packages/core-blocks/src/Split/Split.tsx
```diff
@@ -12,13 +12,25 @@ import { s } from '../s';
...
   const horizontal = split === 'horizontal';
 
-  return <BaseSplit className={s(styles, { split: true, vertical, horizontal }, className)} split={split} {...rest} />;
+  if (!disableAutoMargin) {
+    minSize = AUTO_MARGIN;
+    maxSize = -AUTO_MARGIN;
+  }
+
+  return (
+    // @ts-ignore
```
sergeyteleshev: TODO remove it

webapp/packages/core-blocks/src/Split/SplitControls.tsx
```diff
@@ -11,15 +11,39 @@ import { useS } from '../useS';
...
+  const isResizeMode = split.state.mode === 'resize';
 
   const inverse = split.state.isMainSecond();
 
   let inverseMode = split.state.mode;
 
-  if (split.state.size > split.state.getContainerSize()) {
+  if (split.state.size > split.state.getContainerSize() - split.state.maxSize) {
     inverseMode = 'maximize';
   }
```
Wroud: it's may be a chance that after changes in go-split we will no need in this logic, because `split.state.mode` must return correct mode

webapp/packages/core-blocks/src/Split/Pane.m.css
```diff
@@ -3,3 +3,8 @@
   overflow: auto;
   z-index: 0;
 }
+
+.hiddenContent {
```
devnaumov: This content will be accessible to screen readers

webapp/packages/core-blocks/src/Split/Pane.tsx
```diff
@@ -11,9 +11,12 @@ import { observer } from 'mobx-react-lite';
 import { s } from '../s';
 import { useS } from '../useS';
 import style from './Pane.m.css';
+import { useSplit } from './useSplit';
 
 export const Pane = observer<PaneProps>(function Pane({ className, ...rest }) {
   const styles = useS(style);
+  const split = useSplit();
+  const hiddenContent = rest.main && split.state.mode === 'minimize';
```
Wroud: do we want to also hide content of second pane when mode is 'maximize'

webapp/packages/core-blocks/src/Split/Pane.m.css
Wroud: moreover elements will be accessible with TAB

webapp/packages/core-blocks/src/Split/Pane.tsx
sergeyteleshev: I dont think so. We should see that something left here we can work with. Otherwise it is easy to get lost in the app

webapp/packages/core-blocks/src/Split/Pane.tsx
```diff
@@ -7,13 +7,22 @@
...
 import { useS } from '../useS';
 import style from './Pane.m.css';
+import { useSplit } from './useSplit';
 
-export const Pane = observer<PaneProps>(function Pane({ className, ...rest }) {
+export const Pane = observer<PaneProps>(function Pane({ className, children, ...rest }) {
   const styles = useS(style);
+  const split = useSplit();
+  const shouldHideContent = rest.main && split.state.mode === 'minimize';
+  const childrenMemo = useMemo(() => (shouldHideContent ? null : children), [shouldHideContent, children]);
```
Wroud: but children will be always new, isn't it?
what purpose of using useMemo?

Wroud (approved):

PR #2344, opened 2024-01-31
CB-4400 removed project margin styles to prevent tree jump on DnD

sergeyteleshev:


webapp/packages/plugin-navigation-tree-rm/src/Tree/ProjectsRenderer/navigationTreeProjectsRendererRenderer.tsx
```diff
@@ -113,16 +112,14 @@ const ProjectRenderer: NavigationNodeRendererComponent = observer(function Manag
     );
   }
 
-  const project = node.nodeType === NAV_NODE_TYPE_RM_PROJECT && singleProject && !isDragging;
```
Wroud, line 116: This may break scripts/datasets trees (in these trees, we display fake projects instead of Scripts or Datasets folder)

Wroud (approved):

PR #2347, opened 2024-01-31
CB-4605 fix: sql editor shows datasets panels correctly

sergeyteleshev:


webapp/packages/core-app/src/AppScreen/RightArea.tsx
```diff
@@ -55,7 +55,7 @@ export const RightArea = observer<Props>(function RightArea({ className }) {
             </Loader>
           </Pane>
           <ResizerControls />
-          <Pane className={s(styles, { pane: true })} basis="30%" main>
+          <Pane className={s(styles, { pane: true })} main>
```
Wroud: please revert

webapp/packages/plugin-sql-editor/src/SqlEditor.tsx
```diff
@@ -37,15 +40,30 @@ export const SqlEditor = observer<Props>(function SqlEditor({ state }) {
...
+  const isEditorEmpty = displayedEditors === 0;
+
+  useEffect(() => {
+    if (isEditorEmpty) {
+      split.fixate('maximize', true);
+    } else if (split.state.disable) {
+      split.fixate('resize', false);
+      split.state.setSize(-1);
+    }
+  }, [isEditorEmpty]);
```
sergeyteleshev: I moved this logic in one place cause it looks weird when it is separate and harder to nail its behavior

webapp/packages/plugin-sql-editor/src/SqlEditor/SqlEditor.tsx
```diff
@@ -101,15 +90,15 @@ export const SqlEditor = observer<ISqlEditorProps>(function SqlEditor({ state, c
       lazy
       onChange={handleModeSelect}
     >
-      <sql-editor className={className}>
-        <SQLEditorActions data={data} state={state} />
-        <TabPanelList />
-        {displayedEditors > 1 ? (
+      {displayedEditors > 1 ? (
```
sergeyteleshev: Also I did that to prevent render this AI stuff bug and other similar issues

Wroud: this now totally wrong because we will not render SQL Editor at all when we have only one editor (for example in CE)

this logic was only for hiding tabs when we can't switch to different editor (for example SQL / VQB)

Wroud (changes requested):

Wroud (approved):

PR #2357, opened 2024-02-05
Cb 4633 unit tests for utils

sergeyteleshev:
Снимок экрана 2024-02-05 в 15 10 55

locally tests works just fine. I believe it is because job launches with cache and some dist causes a problem


webapp/packages/core-utils/src/OrderedMap.test.ts
```diff
@@ -0,0 +1,143 @@
...
+  it('should add and get items', () => {
+    const map = new OrderedMap<number, string>();
+    map.add(1, 'one');
+    map.add(2, 'two');
+    map.add(3, 'three');
+
+    expect(map.get(1)).toBe('one');
+    expect(map.get(2)).toBe('two');
+    expect(map.get(3)).toBe('three');
+  });
```
Wroud: please add negative checks like `get not exist item returns undefined`, `has returns false for not exist element`

webapp/packages/core-utils/src/__custom__mocks__/ErrorBoundaryMock.tsx
```diff
@@ -0,0 +1,28 @@
...
+  }
+
+  render() {
+    if (this.state.hasError) {
+      return <h1>{this.state.error.message}</h1>;
+    }
+
+    return this.props.children;
+  }
+}
```
Wroud: will be better to move it to `core-blocks` cuz we have `ErrorBoundary` component here

Wroud (changes requested):

Wroud (approved):
please add license message and update TS project references

Wroud (approved):

PR #2363, opened 2024-02-06
CB-4596 fix http cookies

alexander-skoblikov:


YuraKril (commented):
A cookie has been set without the HttpOnly flag, which means that the cookie can be accessed by JavaScript. If a malicious script can be run on this page then the cookie will be accessible and can be transmitted to another site. If this is a session cookie then session hijacking may be possible.

PR #2367, opened 2024-02-07
CB-4628 remove unused logic

devnaumov:


webapp/packages/core-blocks/src/FormControls/Filter.tsx
```diff
@@ -16,12 +16,12 @@ import filterStyle from './Filter.m.css';
...
 
 interface BaseProps {
-  toggleMode?: boolean;
   placeholder?: string;
   disabled?: boolean;
   max?: boolean;
   className?: string;
-  onToggle?: (status: boolean) => void;
+  onFilter?: (value: string) => void;
+  onClean?: () => void;
```
Wroud: what difference between `onFilter` and `onChange`?
is `onClean` is the same as `onChange` with `value === ''`

devnaumov: onFilter triggers only on filter button click in manual mode. onChange fires every time
What about onClean, we want to know when the user clean the input by pressing clean button,  not when he deletes everything from input by pressing delete for example

webapp/packages/core-blocks/src/FormControls/Filter.tsx
```diff
@@ -73,55 +70,53 @@ export const Filter = observer<ControlledProps | ObjectsProps<any, any>>(functio
...
 
-  const toggle = useCallback(() => {
-    if (!toggleMode) {
-      return;
-    }
-
-    if (toggled) {
-      filter('');
-    }
+  let value: any = valueControlled;
```
sergeyteleshev, line 77: add type please

sergeyteleshev (approved):

PR #2370, opened 2024-02-08
CB-3563 cloudbeaver preference store

alexander-skoblikov:


webapp/packages/core-settings/src/SettingsScopeType.ts
```diff
@@ -6,4 +6,4 @@
  * you may not use this file except in compliance with the License.
  */
 
-export type SettingsScopeType = 'plugin' | 'core';
+export type SettingsScopeType = 'plugin' | 'core' | 'sql';
```
Wroud: maybe `'plugin' | 'core' | string` will be better

webapp/packages/plugin-codemirror6/src/ReactCodemirror.tsx
```diff
@@ -155,12 +155,13 @@ export const ReactCodemirror = observer<IReactCodeMirrorProps, IEditorRef>(
     useLayoutEffect(() => {
       if (view) {
         const transaction: TransactionSpec = { annotations: [External.of(true)] };
+        const isCursorInDoc = cursor && cursor.anchor > 0 && cursor.anchor < view.state.doc.length;
 
         if (value !== undefined && value !== view.state.doc.toString()) {
           transaction.changes = { from: 0, to: view.state.doc.length, insert: value };
         }
```
Wroud:
```suggestion
        let isCursorInDoc = cursor && cursor.anchor > 0 && cursor.anchor < view.state.doc.length;

        if (value !== undefined && value !== view.state.doc.toString()) {
          transaction.changes = { from: 0, to: view.state.doc.length, insert: value };
          isCursorInDoc = cursor && cursor.anchor > 0 && cursor.anchor < value.length
        }
```

webapp/packages/plugin-sql-editor/src/SqlEditor/useSqlEditor.ts
```diff
@@ -177,6 +177,16 @@ export function useSqlEditor(state: ISqlEditorTabState): ISQLEditorData {
       },
 
       setCursor(begin: number, end = begin): void {
+        if (begin > end) {
+          throw new Error('Cursor begin can not be greater than the end of it');
+        }
+
+        const script = this.parser.actualScript;
```
Wroud:
```suggestion
        const scriptLength = this.value.length;
```

webapp/packages/plugin-sql-editor/src/SqlEditor/useSqlEditor.ts
```diff
@@ -177,6 +177,16 @@ export function useSqlEditor(state: ISqlEditorTabState): ISQLEditorData {
...
+
+        if (end > script.length) {
+          end = script.length;
+        }
+
         this.cursor = {
           begin,
           end,
```
Wroud:
```suggestion
        this.cursor = {
          begin: Math.min(begin, scriptLength),
          end: Math.min(end, scriptLength),
```

webapp/packages/plugin-sql-editor/src/SqlEditor/useSqlEditor.ts
```diff
@@ -502,6 +512,10 @@ export function useSqlEditor(state: ISqlEditorTabState): ISQLEditorData {
     executor: data.dataSource?.onSetScript,
     handlers: [
       function setScript({ script }) {
+        if (data.cursor.end > script.length) {
+          data.setCursor(script.length);
+        }
```
Wroud:
```suggestion
        // ensure that cursor is in script boundaries
        data.setCursor(data.cursor.begin, data.cursor.end);
```

webapp/packages/plugin-sql-editor/src/locales/ru.ts
```diff
@@ -1,5 +1,7 @@
 export default [
   ['plugin_sql_editor_sql_editor_settings_group', 'SQL Редактор'],
+  ['sql_editor_settings_insert_table_aliases_name', 'Вставлять алиасы таблиц (в FROM условии)'],
+  ['sql_editor_settings_insert_table_aliases_desc', 'Automatically inserts table aliases'],
```
Wroud: please translate

Wroud (changes requested):

webapp/packages/plugin-sql-editor/src/SqlEditorSettingsService.ts
```diff
@@ -6,21 +6,34 @@
...
+import { SQL_EDITOR_SETTINGS_GROUP } from './SQL_EDITOR_SETTINGS_GROUP';
+
 const defaultSettings = schema.object({
   maxFileSize: schema.coerce.number().default(10 * 1024), // kilobyte
   disabled: schema.coerce.boolean().default(false),
   autoSave: schema.coerce.boolean().default(true),
 });
 
+const defaultProposalInsertTableSettings = schema.object({
+  alias: schema.coerce.boolean().default(true),
```
devnaumov, line 28: please use boolean name, `aliasesEnabled` or something like this

webapp/packages/plugin-sql-editor/src/SqlEditorSettingsService.ts
```diff
@@ -6,21 +6,34 @@
...
+});
+
 export type SqlEditorSettings = schema.infer<typeof defaultSettings>;
 
 @injectable()
 export class SqlEditorSettingsService extends Dependency {
   readonly settings: PluginSettings<typeof defaultSettings>;
+  readonly proposalInsertTableSettings: PluginSettings<typeof defaultProposalInsertTableSettings>;
```
devnaumov, line 36: why do we need another settings object?
```
const defaultSettings = schema.object({
  maxFileSize: schema.coerce.number().default(10 * 1024), // kilobyte
  disabled: schema.coerce.boolean().default(false),
  autoSave: schema.coerce.boolean().default(true),
  aliasesEnabled: schema.coerce.boolean().default(true),
});
```

devnaumov (changes requested):

webapp/packages/plugin-sql-editor/src/SqlEditorSettingsService.ts
sergeyteleshev: `sql.proposals.insert.table.alias` - this is a pretty strict setting name I cannot rename cause it set on the backend side of the app. and it should be consistent to dbeaver desktop analogue setting

in the code of course it is used with the boolean prefix:
https://github.com/dbeaver/cloudbeaver/pull/2370/files#diff-603b6bb92ed26d86a6494be8e6ba52611dd550d3754f8bb9ed8b22ce7fa9ec81R42

webapp/packages/plugin-sql-editor/src/SqlEditorSettingsService.ts
sergeyteleshev: cause `defaultSettings` has its own paths in product config
and new setting has another its own string path path which is not similar to `this.settings` at all

I tried to push the backend side to give us setting with the path like you want (so we have consistensy here) , but they said it is important to do it this way. And this is why we have this setting like this

Wroud (approved):

devnaumov (approved):

PR #2377, opened 2024-02-12
CB-4436 update session only in touchSession

alexander-skoblikov:


webapp/packages/core-app/src/useClientActivity.ts
```diff
@@ -0,0 +1,41 @@
...
+ */
+import { useEffect } from 'react';
+
+import { ClientActivityService } from '@cloudbeaver/core-activity';
+import { useService } from '@cloudbeaver/core-di';
+
+export function useClientActivity() {
+  const clientActivityService = useService(ClientActivityService);
+
+  function updateActivity() {
```
sergeyteleshev: this one executes a lot. I suppose we can add here a throttle with 100ms. what do you think, guys?

webapp/packages/plugin-session-expiration/src/SessionExpireWarningDialog/SessionExpireWarningDialogBootstrap.ts
```diff
@@ -66,15 +73,17 @@ export class SessionExpireWarningDialogBootstrap extends Bootstrap {
 
   private async open(): Promise<void> {
     if (!this.dialogInternalPromise) {
-      this.dialogInternalPromise = this.commonDialogService.open(SessionExpireWarningDialog, null);
+      this.dialogInternalPromise = this.commonDialogService.open(SessionExpireWarningDialog, {
+        onButtonClick: () => this.clientActivityService.updateActivity(),
```
sergeyteleshev: also since we control activity each user action, there is no need for this due to the fact `touchSession` will be executed because of mouse being moved or clicked

but we can leave it like this just in case if something went wrong in `useClientActivity` hook

webapp/packages/core-activity/src/ClientActivityService.ts
```diff
@@ -0,0 +1,60 @@
...
+ * Copyright (C) 2020-2024 DBeaver Corp and others
+ *
+ * Licensed under the Apache License, Version 2.0.
+ * you may not use this file except in compliance with the License.
+ */
+import { makeObservable, observable } from 'mobx';
+
+import { injectable } from '@cloudbeaver/core-di';
+import { Executor, IExecutor } from '@cloudbeaver/core-executor';
+import { SessionResource } from '@cloudbeaver/core-root';
```
Wroud: please remove core-root dependency

webapp/packages/core-activity/src/ClientActivityService.ts
```diff
@@ -0,0 +1,60 @@
...
+
+const INACTIVE_PERIOD_TIME = 1000 * 60;
+
+@injectable()
+export class ClientActivityService {
+  private timer: NodeJS.Timeout | null;
```
Wroud: please use `ReturnType<typeof setTimeout>`

webapp/packages/core-app/src/useClientActivity.ts
Wroud: yep it's a good idea you can use `throttle` from `core-utils`

webapp/packages/core-root/src/SessionResource.ts
```diff
@@ -95,6 +92,23 @@ export class SessionResource extends CachedDataResource<SessionState | null> {
     return session;
   }
 
+  touchSession = () => {
```
Wroud: please avoid the use of arrow functions. It's hard to debug

webapp/packages/core-root/src/SessionResource.ts
```diff
@@ -95,6 +92,23 @@ export class SessionResource extends CachedDataResource<SessionState | null> {
...
+    if (this.data?.valid) {
+      this.graphQLService.sdk.touchSession();
+    }
+
+    this.touchSessionTimer = setTimeout(() => {
+      if (this.touchSessionTimer) {
+        clearTimeout(this.touchSessionTimer);
+        this.touchSessionTimer = null;
+      }
+    }, SESSION_TOUCH_TIME_PERIOD);
```
Wroud: please move it to `SessionExpireService`
so:
in `SessionExpireService` you will run interval and check if user active or not and if it is active you will fire `touchSession`

webapp/packages/plugin-session-expiration/src/SessionExpireWarningDialog/SessionExpireWarningDialogBootstrap.ts
```diff
@@ -5,13 +5,19 @@
...
 import { Bootstrap, injectable } from '@cloudbeaver/core-di';
 import { CommonDialogService, DialogueStateResult } from '@cloudbeaver/core-dialogs';
 import { ServerConfigResource, SessionExpireService, SessionResource } from '@cloudbeaver/core-root';
 import { GraphQLService } from '@cloudbeaver/core-sdk';
 
-import { SessionExpireWarningDialog } from './SessionExpireWarningDialog';
+const SessionExpireWarningDialog = React.lazy(async () => {
+  const { SessionExpireWarningDialog } = await import('./SessionExpireWarningDialog');
+  return { default: SessionExpireWarningDialog };
+});
```
Wroud: please use `importLazyComponent` instead

webapp/packages/plugin-session-expiration/src/SessionExpireWarningDialog/SessionExpireWarningDialog.tsx
```diff
@@ -20,18 +20,27 @@ import {
...
+
   return (
     <CommonDialogWrapper size="small" fixedSize>
       <CommonDialogHeader title="app_root_session_expire_warning_title" onReject={rejectDialog} />
       <CommonDialogBody noOverflow>
         <p className={s(styles, { text: true })}>{translate('app_root_session_expire_warning_message')}</p>
       </CommonDialogBody>
       <CommonDialogFooter className={s(styles, { footer: true })}>
-        <Button type="button" mod={['unelevated']} onClick={rejectDialog}>
+        <Button type="button" mod={['unelevated']} onClick={onContinueClick}>
```
Wroud: please revert all related changes
actually when user resolves or closes dialog we will execute `touchSession`
when session expired dialog will be closed automatically and we will check this state

Wroud (changes requested):

webapp/packages/plugin-session-expiration/src/SessionExpireWarningDialog/SessionExpireWarningDialogBootstrap.ts
Wroud: actually it's must work like this:
each 1m execute `touchSession` if user active
execute `touchSession` if user closed or resolved `SessionExpireWarningDialog` dialog before session expiration

webapp/packages/core-root/src/SessionResource.ts
```diff
@@ -95,6 +92,23 @@ export class SessionResource extends CachedDataResource<SessionState | null> {
     return session;
   }
 
+  touchSession = () => {
+    if (this.touchSessionTimer) {
+      return;
+    }
+
+    if (this.data?.valid) {
+      this.graphQLService.sdk.touchSession();
```
Wroud: This call will update the session lifetime. We need to read the session from this method and update `setData` of the resource like it was in `refreshSilent`

Wroud (changes requested):

webapp/packages/core-root/src/SessionExpireSource.ts
sergeyteleshev, line 1: I added this abstraction to avoid cyclic dependency

webapp/packages/core-activity/src/ClientActivityService.ts
```diff
@@ -0,0 +1,58 @@
...
+
+    makeObservable(this, {
+      isActive: observable,
+    });
+  }
+
+  private setActivity = (value: boolean) => {
+    this.isActive = value;
+    this.onActiveStateChange.execute(value);
+  };
```
Wroud: please don't use arrow functions

webapp/packages/core-root/src/SessionExpireSource.ts
```diff
@@ -0,0 +1,22 @@
+import { makeObservable, observable } from 'mobx';
+
+import { injectable } from '@cloudbeaver/core-di';
+
+@injectable()
+export class SessionExpireSource {
```
Wroud: please rename `SessionExpireSourceService`

Wroud: and add license info

Wroud: this service now works as state and i can see one big problem:
now if you call `setExpired` manually when session not expired what will happen?
seems like we can't create service like this

Wroud (changes requested):

webapp/packages/core-root/src/SessionResource.ts
sergeyteleshev: This leads to cyclic dependency. My solution of this did not bring us any fine result. I also had several tries to reduce it manually and find the place where something went wrong, but it feels like there is no easy peasy way to reduce this problem without causing a possible new ones 

I decided to move logic back to the resource so we have no cyclic dependency. I believe this problem deserves a tech ticket

webapp/packages/plugin-administration/src/ConfigurationWizard/ServerConfiguration/Form/ServerConfigurationInfoForm.tsx
```diff
@@ -8,14 +8,16 @@
...
-import { ServerConfigResource } from '@cloudbeaver/core-root';
+import { ServerConfigResource, SESSION_EXPIRE_WARN_IN_TIME, SESSION_TOUCH_TIME_PERIOD } from '@cloudbeaver/core-root';
 
 import type { IServerConfigurationPageState } from '../IServerConfigurationPageState';
 
 interface Props {
   state: IServerConfigurationPageState;
 }
 
+const minSessionExpireTime = Math.ceil((SESSION_EXPIRE_WARN_IN_TIME + SESSION_TOUCH_TIME_PERIOD) / 60000) + 1;
```
devnaumov: If the const is outside of the component use upper case MIN_SESSION_EXPIRE_TIME

webapp/packages/plugin-session-expiration/src/SessionExpireWarningDialog/SessionExpireWarningDialogBootstrap.ts
```diff
@@ -6,15 +6,13 @@
...
 import { CommonDialogService, DialogueStateResult } from '@cloudbeaver/core-dialogs';
-import { ServerConfigResource, SessionExpireService, SessionResource } from '@cloudbeaver/core-root';
+import { ServerConfigResource, SESSION_EXPIRE_WARN_IN_TIME, SessionExpireService, SessionResource } from '@cloudbeaver/core-root';
 import { GraphQLService } from '@cloudbeaver/core-sdk';
 
-import { SessionExpireWarningDialog } from './SessionExpireWarningDialog';
-
-const WARN_IN = 5 * 1000 * 60;
-
+const SessionExpireWarningDialogLoader = importLazyComponent(() => import('./SessionExpireWarningDialog').then(m => m.SessionExpireWarningDialog));
```
devnaumov: Its not a loader, so just SessionExpireWarningDialog

webapp/packages/core-root/src/SessionExpireService.ts
```diff
@@ -9,16 +9,26 @@ import { Bootstrap, injectable } from '@cloudbeaver/core-di';
 import { Executor, IExecutor } from '@cloudbeaver/core-executor';
 import { EServerErrorCode, GQLError, GraphQLService, SessionError } from '@cloudbeaver/core-sdk';
 import { errorOf } from '@cloudbeaver/core-utils';
+
+export const SESSION_EXPIRE_WARN_IN_TIME = 5 * 1000 * 60;
+
```
Wroud: seems as not the best place for this constant (not in use)

webapp/packages/core-root/src/SessionResource.ts
```diff
@@ -95,6 +99,30 @@ export class SessionResource extends CachedDataResource<SessionState | null> {
     return session;
   }
 
+  async touchSession(force?: boolean) {
+    if (!this.data?.valid || this.touchSessionTimer || !this.clientActivityService.isActive || !force) {
+      return;
+    }
+
+    const valid = await this.graphQLService.sdk.touchSession();
+
```
Wroud: this function will never be executed without `force=true`, `force` is `false` by default so any call `touchSession()` will do nothing

Wroud (changes requested):

webapp/packages/core-root/src/SessionResource.ts
```diff
@@ -48,6 +48,11 @@ export class SessionResource extends CachedDataResource<SessionState | null> {
       this,
     );
 
+    this.touchSession = this.touchSession.bind(this);
+    this.processAction = this.processAction.bind(this);
+    this.setDefaultLocale = this.setDefaultLocale.bind(this);
+    this.changeLanguage = this.changeLanguage.bind(this);
+
```
Wroud: do we really need this?

webapp/packages/core-root/src/SessionResource.ts
```diff
@@ -95,6 +93,23 @@ export class SessionResource extends CachedDataResource<SessionState | null> {
...
+    }
+
+    const valid = await this.graphQLService.sdk.touchSession();
+
+    if (!valid) {
+      this.setData({
+        ...this.data,
+        valid,
+      });
+    }
```
Wroud: how we will now how many time left to session expire? (we need to get new `remainingTime` value because `touchSession` will adjust it)

webapp/packages/core-root/src/SessionTouchService.ts
```diff
@@ -0,0 +1,39 @@
...
+    this.touchSession = this.touchSession.bind(this);
+    this.clientActivityService.onActiveStateChange.addHandler(this.touchSession);
+  }
+
+  private touchSession = () => {
+    if (this.touchSessionTimer || !this.clientActivityService.isActive) {
+      return;
+    }
+
+    this.sessionResource.touchSession();
```
Wroud: in case this async function will be finished with exception we will get 'uncaught exception error in console log'
please add try/catch or .then.catch

webapp/packages/core-root/src/SessionTouchService.ts
```diff
@@ -0,0 +1,39 @@
...
+
+  constructor(private readonly clientActivityService: ClientActivityService, private readonly sessionResource: SessionResource) {
+    super();
+    this.touchSession = this.touchSession.bind(this);
+    this.clientActivityService.onActiveStateChange.addHandler(this.touchSession);
+  }
+
+  private touchSession = () => {
+    if (this.touchSessionTimer || !this.clientActivityService.isActive) {
+      return;
```
Wroud: PLEASE DONT USE ARROW FUNCTIONS (!!!!!!!!!) (AND `this.touchSession = this.touchSession.bind(this);` make no sense for arrow functions)

it will be better to rename this service not by API method but for logic, something like `SessionActivityService` and method something like `touchSession ` -> `notifyClientActivity`

webapp/packages/plugin-session-expiration/src/SessionExpireWarningDialog/SessionExpireWarningDialogBootstrap.ts
```diff
@@ -74,7 +72,7 @@ export class SessionExpireWarningDialogBootstrap extends Bootstrap {
         const { sessionState } = await this.graphQLService.sdk.sessionState();
 
         if (sessionState.valid) {
-          await this.sessionResource.refreshSilent();
+          this.sessionResource.touchSession();
```
Wroud: in case this async function will be finished with exception we will get 'uncaught exception error in console log'
please add try/catch or .then.catch

Wroud (changes requested):

webapp/packages/core-root/src/SessionActivityService.ts
```diff
@@ -0,0 +1,43 @@
...
+
+  private async notifyClientActivity() {
+    if (this.touchSessionTimer || !this.clientActivityService.isActive) {
+      return;
+    }
+
+    try {
+      await this.sessionResource.touchSession();
+    } catch (e) {
+      console.error('Session touch error', e);
```
sergeyteleshev: It is forbidden to import `notificationService` into `core-root`, so I logged it with this

webapp/packages/core-root/src/SessionActivityService.ts
```diff
@@ -0,0 +1,43 @@
...
+
+  private async notifyClientActivity() {
+    if (this.touchSessionTimer || !this.clientActivityService.isActive) {
+      return;
+    }
+
+    try {
+      await this.sessionResource.updateSession();
+    } catch (e) {
+      console.error('Session update error', e);
```
sergeyteleshev, line 33: cannot import NotificationService here so just logged it

webapp/packages/core-root/src/SessionActivityService.ts
```diff
@@ -0,0 +1,43 @@
...
+    }
+
+    this.touchSessionTimer = setTimeout(() => {
+      if (this.touchSessionTimer) {
+        clearTimeout(this.touchSessionTimer);
+        this.touchSessionTimer = null;
+      }
+    }, SESSION_TOUCH_TIME_PERIOD);
```
Wroud, line 41: please keep naming for constants also (notifyClientActivity)

webapp/packages/core-root/src/SessionResource.ts
```diff
@@ -95,6 +88,18 @@ export class SessionResource extends CachedDataResource<SessionState | null> {
     return session;
   }
 
+  async updateSession() {
+    if (!this.data?.valid) {
+      return;
+    }
+
+    const session = (await this.graphQLService.sdk.updateSession()).updateSession;
```
Wroud: better to use object destructuring here

webapp/packages/core-root/src/SessionActivityService.ts
Wroud: you can move it to `plugin-session-expiration` (not the best place but it's ok)

Wroud (approved):

PR #2399, opened 2024-02-20
CB-4685 disable default theme setting (temporary)

sergeyteleshev:


webapp/packages/core-theming/src/ThemeSettingsManagementService.ts
```diff
@@ -16,14 +16,14 @@ export class ThemeSettingsManagementService extends Dependency {
...
     settingsManagerService.registerSettings(themeSettingsService.settings, () => [
-      {
-        key: 'defaultTheme',
-        type: ESettingsValueType.Select,
-        name: 'core_theming_settings_default_theme_name',
-        description: 'core_theming_settings_default_theme_description',
-        options: themeService.themes.map(theme => ({ id: theme.id, name: theme.name })),
-        group: INTERFACE_SETTINGS_GROUP,
-      },
+      // {
```
sergeyteleshev, line 19: I see no reason to comment other parts here
`webapp/packages/core-theming/src/ThemeSettingsService.ts`
or here
`webapp/packages/core-theming/src/ThemeService.ts`

Cause it just sets the theme to the light as default and doesn't (won't) affect anything

Wroud: exactly, we want to hide this setting from preferences only

Wroud (approved):

PR #2409, opened 2024-02-22
CB-4732 fix: fields auto hide behaviour

Wroud:


webapp/packages/core-blocks/src/FormControls/useFormStateControl.ts
```diff
@@ -34,12 +34,15 @@ export function useFormStateControl<TState extends Record<string, any>, TKey ext
...
     }
   });
-  if (isNotNullDefined(defaultValue) && isObject(defaultValue)) {
-    defaultValue = defaultValue[name];
-  }
+
+  let presented = true;
 
   if (isNotNullDefined(value) && isObject(value)) {
+    presented = isControlPresented(String(name), value, defaultValue);
```
sergeyteleshev: You pass also a symbol type here converted to string. so it transforms to something like on the screen. Which is probably won't be a part of an object at all

Should we add one more type Symbol to isControlPresented first argument?
And add here one more case where you can pass symbol time without converting it to the string





Снимок экрана 2024-02-22 в 12 57 16

webapp/packages/core-blocks/src/FormControls/useFormStateControl.ts
```diff
@@ -34,12 +34,15 @@ export function useFormStateControl<TState extends Record<string, any>, TKey ext
...
     value = value[name];
+  } else if (isNotNullDefined(defaultValue) && isObject(defaultValue)) {
+    presented = isControlPresented(String(name), defaultValue);
```
sergeyteleshev: same here

sergeyteleshev (approved):

webapp/packages/core-blocks/src/FormControls/useFormStateControl.ts
Wroud: fixed

devnaumov (approved):

sergeyteleshev (approved):

PR #2420, opened 2024-02-27
CB-4743 adds html sanitizer

sergeyteleshev:


webapp/packages/core-utils/src/sanitizeHtml.test.tsx
```diff
@@ -0,0 +1,42 @@
+/*
+ * CloudBeaver - Cloud Database Manager
+ * Copyright (C) 2020-2024 DBeaver Corp and others
+ *
+ * Licensed under the Apache License, Version 2.0.
+ * you may not use this file except in compliance with the License.
+ */
+import { sanitizeHtml } from './sanitizeHtml';
+
+describe('sanitize', () => {
```
sergeyteleshev: i just wanted to test how library works and decided to left this in case we want to replace the lib with our own or other solution

sergeyteleshev:
tests are okay. I just changed the file name and job here went crazy cause it's left in `dist` library

webapp/packages/core-utils/src/toSafeHtmlString.ts
```diff
@@ -0,0 +1,14 @@
+/*
+ * CloudBeaver - Cloud Database Manager
+ * Copyright (C) 2020-2024 DBeaver Corp and others
+ *
+ * Licensed under the Apache License, Version 2.0.
+ * you may not use this file except in compliance with the License.
+ */
+export function toSafeHtmlString(dirty: string): string {
```
devnaumov, line 8: Probably better to use lib like dompurify to sanitize html

sergeyteleshev: this was my first solution. @Wroud suggested to generate a html safe string
in this case we kill 2 rabbits:
- the code is not executable by browser
- the user will see that something illegal happened in his JSON tab editor and start investigation what happened
 
image
image

Wroud (approved):

devnaumov (approved):

PR #2422, opened 2024-02-28
CB-4222 feat: read blob data for value panel

Wroud:


webapp/packages/plugin-data-viewer/src/ValuePanelPresentation/TextValue/useTextValue.ts
```diff
@@ -5,82 +5,150 @@
...
+
+    if (!isNotNullDefined(elementKey)) {
+      return value;
+    }
+
+    const contentValue = formatAction.get(elementKey);
+    const isBinary = formatAction.isBinary(elementKey);
+    const cachedFullText = contentAction.retrieveFullTextFromCache(elementKey);
+
+    if (isBinary && isResultSetContentValue(contentValue)) {
```
devnaumov, line 125: Probably could be simplified

```
if (isBinary && isResultSetContentValue(contentValue)) {
      if (!contentValue.binary) {
        if (contentValue.text) {
          value = contentValue.text;
        }
      } else {
        value = contentValue.binary;
      }
    } else if (isResultSetBlobValue(contentValue)) {
      value = parsedBlobValueGetter() ?? '';
    } else {
      value = cachedFullText || formatAction.getText(elementKey);
    }
```

webapp/packages/plugin-data-viewer/src/ValuePanelPresentation/ImageValue/ImageValuePresentation.tsx
```diff
@@ -82,21 +83,19 @@ export const ImageValuePresentation: TabContainerPanelComponent<IDataValuePanelP
...
         },
-        get src(): string | null {
+        get src(): string | Blob | null {
           if (isResultSetBlobValue(this.cellValue)) {
             // uploaded file preview
-            return URL.createObjectURL(this.cellValue.blob);
+            return this.cellValue.blob;
           }
 
           if (this.staticSrc) {
```
sergeyteleshev, line 92: probably we want this condition to be at the end of the function
so if we have some blobs cached we return it first
and only then if there are nothing to give from cache we can give a link to download

webapp/packages/core-blocks/src/useSuspense.ts
```diff
@@ -0,0 +1,120 @@
...
+ *
+ * (!!!) Don't access the suspended value in the same component where useSuspense is declared
+ * @returns
+ */
+export function useSuspense(): ISuspense {
+  const state = useObservableRef<ISuspenseState>(
+    () => ({
+      observedValueMetadata: new Map(),
+
+      observedValue<TArgs, TValue>(key: string, args: () => TArgs, loader: (args: TArgs) => Promise<TValue>): () => TValue {
```
sergeyteleshev, line 42: we can consider the `key` argument to be a symbol, not a string. to avoid collisions when the same key can has 2 different values depending on context where it was called

Wroud: there is no difference because the main problem is that the same component can be mounted more than 1 time and will use the same key (no matter the type)

webapp/packages/plugin-data-viewer/src/ValuePanelPresentation/ImageValue/ImageValuePresentation.tsx
Wroud: priority is:
1. changed value (preview)
2. available src data (we don't need to have cache in this case)
3. available cache if src is not available

webapp/packages/plugin-data-viewer/src/ValuePanelPresentation/TextValue/useTextValue.ts
Wroud: fixed

sergeyteleshev (approved):

devnaumov (approved):

PR #2433, opened 2024-02-29
CB-4018 fix: hides chart settings if charts tab is not selected

sergeyteleshev:


webapp/packages/core-data-context/src/DataContext/TempDataContext.ts
```diff
@@ -28,6 +28,7 @@ export class TempDataContext implements IDataContext {
     this.fallback = fallback;
 
     makeObservable<this>(this, {
+      map: observable.shallow,
```
sergeyteleshev, line 31: should we do same thing for the? to avoid similar bug cases
`webapp/packages/core-data-context/src/DataContext/DynamicDataContext.ts`

Wroud: this map is special map that should not trigger mobx reactions, seems like problem here that we can skip subscription to the origin context (target)
example:

```ts
  hasOwn(context: DataContextGetter<any>): boolean {
    return this.map.has(context) || this.target.hasOwn(context);
  }

  has(context: DataContextGetter<any>, nested = true): boolean {
    if (this.hasOwn(context)) {
      return true;
    }

    if (nested && this.fallback?.has(context)) {
      return true;
    }

    return false;
  }
```

imagine this code in the render function:
```js
...
context.set(someContext, value);
context.has(someContext);
```

what will happen:
1. context added to unobservable `map`
2. `has` call will be equal to `this.map.has(context)`
3. component rendered
4. data flushed to original context
5. now `map` is empty and value is in original context
6. component hasn't subscribed to the original context and will skip all followed changes in the `context`

How to fix:
lets look at `hasOwn` method:
```ts
  hasOwn(context: DataContextGetter<any>): boolean {
    return this.map.has(context) || this.target.hasOwn(context);
  }
```
we need to rewrite it:
```ts
  hasOwn(context: DataContextGetter<any>): boolean {
    const isTargetHasOwn = this.target.hasOwn(context);
    return this.map.has(context) || isTargetHasOwn;
  }
```

now we will always subscribe to the target context

Wroud (changes requested):

webapp/packages/core-data-context/src/DataContext/TempDataContext.ts
Wroud: maybe we also need a `signal` when we flush data
https://mobx.js.org/custom-observables.html

so we will call `this.atom.reportObserved()` each time we access `TempDataContext` and `this.atom.reportChanged()` each time we flush

like this:
```ts
  hasOwn(context: DataContextGetter<any>): boolean {
    this.atom.reportObserved();
    const isTargetHasOwn = this.target.hasOwn(context);
    return this.map.has(context) || isTargetHasOwn;
  }
```

webapp/packages/core-data-context/src/DataContext/TempDataContext.ts
```diff
@@ -20,12 +20,14 @@ export class TempDataContext implements IDataContext {
   target: IDataContext;
   fallback?: IDataContextProvider;
   private flushTimeout: any;
+  atom: IAtom;
```
Wroud: private

Wroud (approved):

Wroud (approved):

PR #2462, opened 2024-03-08
CB-4686 add new client event for updating session info

yagudin10:


webapp/packages/core-root/src/SessionEventSource.ts
```diff
@@ -135,6 +137,26 @@ export class SessionEventSource implements IServerEventEmitter<ISessionEvent, IS
 
   disconnect() {
```
sergeyteleshev: https://rxjs.dev/api/webSocket/webSocket - the rxjs websocket is tried to reconnect cause there was some subscriptions

webapp/packages/core-root/src/SessionEventSource.ts
```diff
@@ -135,6 +137,26 @@ export class SessionEventSource implements IServerEventEmitter<ISessionEvent, IS
...
+      this.openSubject.unsubscribe();
+    }
+
+    if (!this.oldEventsSubject.closed) {
+      this.oldEventsSubject.unsubscribe();
+    }
+
+    if (!this.errorSubject.closed) {
+      this.errorSubject.unsubscribe();
+    }
```
Wroud: this will lead to dead `SessionEventSource` and all it's subscriptions.
What if we want to pause and resume websocket?

webapp/packages/core-root/src/SessionEventSource.ts
```diff
@@ -76,10 +76,12 @@ export class SessionEventSource implements IServerEventEmitter<ISessionEvent, IS
 
     this.openSubject.subscribe(() => {
       this.onInit.execute();
+      this.disconnected = false;
```
Wroud: `disconnected` is internal flag that we need to prevent any in and out activity but it's for tracking current websocket status

webapp/packages/core-root/src/SessionResource.ts
```diff
@@ -73,6 +58,25 @@ export class SessionResource extends CachedDataResource<SessionState | null> {
...
+      ...this.data,
+      valid: event?.isValid ?? this.data?.valid,
+      remainingTime: event.remainingTime,
+      actionParameters: event.actionParameters,
+      cacheExpired: event?.isCacheExpired ?? this.data.cacheExpired,
+      lastAccessTime: String(event.lastAccessTime),
+      locale: event.locale,
+    };
+
+    this.setData(sessionState);
```
Wroud: seems like we need to prohibit using setData outside of `loader` or `performUpdate` because it will change data but event's won't be fired

Wroud (changes requested):

webapp/packages/core-root/src/SessionResource.ts
sergeyteleshev: image

webapp/packages/core-root/src/SessionResource.ts
```diff
@@ -73,6 +58,27 @@ export class SessionResource extends CachedDataResource<SessionState | null> {
...
   }
 
+  private handleSessionStateEvent(event: ISessionStateEvent) {
+    if (!this.data) {
+      return;
+    }
+
+    const sessionState: SessionState = {
+      ...this.data,
+      valid: event?.isValid ?? this.data?.valid,
```
devnaumov: seems like this.data is defined here

webapp/packages/core-root/src/SessionResource.ts
```diff
@@ -73,6 +58,27 @@ export class SessionResource extends CachedDataResource<SessionState | null> {
...
+      remainingTime: event.remainingTime,
+      actionParameters: event.actionParameters,
+      cacheExpired: event?.isCacheExpired ?? this.data.cacheExpired,
+      lastAccessTime: String(event.lastAccessTime),
+      locale: event.locale,
+    };
+
+    this.performUpdate(undefined, [], async () => {
+      this.setData(sessionState);
+      this.onStatusUpdate.execute(sessionState);
```
Wroud: performUpdate is asynchronous function and `this.data` may change before update will be executed, please move const here.
Also you can remove `onStatusUpdate` and use `onDataUpdate` instead (it will be called by `performUpdate`)

there also one note, that `performUpdate` probably will trigger to almost all data reload

Wroud (changes requested):

server/bundles/io.cloudbeaver.server/schema/service.events.graphqls
```diff
@@ -114,8 +115,12 @@ type WSSocketConnectedEvent implements CBServerEvent {
 type WSSessionStateEvent implements CBServerEvent {
     id: CBServerEventId!
     topicId: CBEventTopic
+    lastAccessTime: Int!
     remainingTime: Int!
     isValid: Boolean
+    isCacheExpired: Boolean
+    locale: String!
+    actionParameters: Object
```
Wroud, line 123: can we rewrite it like this?:
```gql
type WSSessionStateEvent implements CBServerEvent {
    id: CBServerEventId!
    topicId: CBEventTopic
    session: SessionInfo!
}
```

Wroud (approved):

PR #2478, opened 2024-03-19
CB-4707 move connection shield logic to separate component

devnaumov:


webapp/packages/plugin-connections/src/ConnectionShield.tsx
```diff
@@ -0,0 +1,60 @@
...
+}
+
+export const ConnectionShield = observer<PropsWithChildren<Props>>(function ConnectionShield({ connectionKey, children }) {
+  const translate = useTranslate();
+  const connectionsManagerService = useService(ConnectionsManagerService);
+  const notificationService = useService(NotificationService);
+
+  const connection = useResource(ConnectionShield, ConnectionInfoResource, connectionKey);
+
+  const [connecting, setConnecting] = useState(false);
```
sergeyteleshev, line 27: I suggest to move this flag to `ConnectionsManagerService` and make it observable. from this point connecting state will be available not only here but wherever it is needed in future

webapp/packages/plugin-connections/src/ConnectionShield.tsx
```diff
@@ -0,0 +1,60 @@
...
+    try {
+      await connectionsManagerService.requireConnection(connectionKey);
+    } catch (exception: any) {
+      notificationService.logException(exception);
+    } finally {
+      setConnecting(false);
+    }
+  }
+
+  if (connection.data && !connection.data.connected) {
```
sergeyteleshev, line 45: I dont like we have drawing logic here and in the parent component. Lets make it in one place here

we can add extra prop something like `noDataPlaceholderText`

and if there is no `connectcion.data` we draw placeholder here.

if `noDataPlaceholderText` is not provided, draw a default defined placeholder

after that you can safely remove this code from the parent component:
`<TextPlaceholder>{translate('plugin_object_viewer_table_no_items')}</TextPlaceholder>`

sergeyteleshev (approved):

Wroud (approved):

PR #2495, opened 2024-03-25
Frontend/script for license

sergeyteleshev:


scripts/license.sh
```diff
@@ -0,0 +1,16 @@
+#!/bin/bash
+
+add_missed_license() {
+    local TEXT="$1"
+    local FILES_PATH="$2"
+    TEXT=$(echo "$TEXT" | tr '\n' '\0' | xargs -0 -n1)
+    FILES=$(git diff --cached --name-only --diff-filter=ACM | grep -E "$FILES_PATH" || true)
```
Wroud: ACMR

webapp/package.json
```diff
@@ -20,7 +20,8 @@
   },
   "scripts": {
     "prebuild": "lerna run gql:gen --stream --parallel",
-    "pretest": "lerna run gql:gen --stream --parallel && yarn lerna run test --scope @cloudbeaver/tests-runner"
+    "pretest": "lerna run gql:gen --stream --parallel && yarn lerna run test --scope @cloudbeaver/tests-runner",
+    "prepare": "cd .. && husky install webapp/.husky"
```
Wroud: it seems that we need to move this scripts to webapp

PR #2510, opened 2024-03-27
CB-4841 fix: doesn't show database auth login popup when filters enabled

sergeyteleshev:


webapp/packages/plugin-navigation-tree/src/NavigationTree/ElementsTree/useElementsTree.ts
```diff
@@ -502,9 +502,9 @@ export function useElementsTree(options: IOptions): IElementsTree {
...
       },
       async setFilter(value: string) {
-        runInAction(() => {
-          this.userData.filter = value;
+        this.userData.filter = value;
 
+        runInAction(() => {
           for (const nodeState of this.state.values()) {
             nodeState.showInFilter = false;
           }
```
Wroud: seems like case fix but not fix of the problem
actual problem is here:
https://github.com/dbeaver/cloudbeaver/blob/1105f1300a7fc6f8b0e4403c597cd74391cf929c/webapp/packages/plugin-navigation-tree/src/NavigationTree/ElementsTree/useElementsTree.ts#L203

explanation:
`loadTree` triggered concurrently with user change
`userData` updated but it state isn't applied yet

Wroud (changes requested):

Wroud (approved):

webapp/packages/plugin-connections/src/ConnectionAuthService.ts
```diff
@@ -56,12 +56,10 @@ export class ConnectionAuthService extends Dependency {
       return null;
     }
 
-    let connection = this.connectionInfoResource.get(key);
+    const connection = await this.connectionInfoResource.load(key, ['includeAuthNeeded', 'includeNetworkHandlersConfig', 'includeCredentialsSaved']);
```
devnaumov: We probably want to load the ['includeAuthNeeded', 'includeNetworkHandlersConfig', 'includeCredentialsSaved'] includes only if the function is near the network handlers initialization. This would avoid loading useless data, for example, if we are already connected and the function merely returns the connection.

Wroud (changes requested):
current fix seems too complicated

PR #2519, opened 2024-03-28
Cb 4495 enable minification for function and class names for ee repository

sergeyteleshev:


Wroud:
Unfortunately, we can't implement it this way. We can return to this issue in the future, but right now, it seems like we don't have the right environment for it

PR #2532, opened 2024-04-04
Cb 4178 te cloud storage improvements

sergeyteleshev:


webapp/packages/plugin-navigation-tree/src/NavigationTree/NavigationTreeService.ts
```diff
@@ -66,6 +73,14 @@ export class NavigationTreeService extends View<string> {
     return this.navNodeManagerService.getTree(id);
   }
 
+  getChildrenNodes(id: string): NavNode[] {
+    return (this.getChildren(id) ?? []).map(id => this.getNode(id)).filter(isNotNullDefined);
+  }
+
+  getNode(id: string) {
+    return this.navNodeManagerService.getNode(id);
+  }
```
Wroud: seems extra

webapp/packages/plugin-navigation-tree/src/NavigationTree/NavigationTreeService.ts
```diff
@@ -66,6 +73,14 @@ export class NavigationTreeService extends View<string> {
     return this.navNodeManagerService.getTree(id);
   }
 
+  getChildrenNodes(id: string): NavNode[] {
+    return (this.getChildren(id) ?? []).map(id => this.getNode(id)).filter(isNotNullDefined);
+  }
```
Wroud: empty children is not the same as undefined

Wroud (changes requested):

webapp/packages/core-browser/src/selectFiles.ts
```diff
@@ -6,7 +6,7 @@
  * you may not use this file except in compliance with the License.
  */
 
-export function selectFiles(callback: (files: FileList | null) => any, multiple?: boolean): void {
+export function selectFiles(callback: (files: File[] | null) => any, multiple?: boolean): void {
```
Wroud: we can change type from `Files[] | null` to `Files[]` and use empty array in all cases instead of null
this will simplify future code

Wroud (changes requested):

Wroud (approved):

PR #2539, opened 2024-04-09
CB-4841 fix: concurrent tree update

Wroud:


webapp/packages/plugin-navigation-tree/src/NavigationTree/ElementsTree/useElementsTree.ts
```diff
@@ -170,98 +169,96 @@ export function useElementsTree(options: IOptions): IElementsTree {
       return false;
     }
   }
+  const [loadingNodes] = useState(() => observable.map(new Map<string, Promise<void>>(), { deep: false }));
```
sergeyteleshev, line 172: I think we should move it to `elementsTree` state as `observable.shallow` to keep all trees tools in one place

webapp/packages/plugin-navigation-tree/src/NavigationTree/ElementsTree/useElementsTree.ts
```diff
@@ -170,98 +169,96 @@ export function useElementsTree(options: IOptions): IElementsTree {
...
-
-        if (preloadedRoot !== options.folderExplorer.state.folder) {
-          if (preloadedRoot === undefined) {
-            options.folderExplorer.open([], options.baseRoot);
-          } else {
-            this.exitNodeFolder(preloadedRoot);
-          }
-          return;
+    async loadTree(...nodes: string[]) {
+      await Promise.all(loadingNodes.values());
```
sergeyteleshev, line 176: I suppose we can move it a little bit down in order to preload all resources first and when all data  is being fresh we can start working with upcoming tree nodes

sergeyteleshev (approved):

devnaumov (approved):

PR #2543, opened 2024-04-10
CB-4933 focus snackbar on open

devnaumov:


webapp/packages/core-blocks/src/Snackbars/SnackbarMarkups/SnackbarWrapper.tsx
```diff
@@ -34,7 +36,7 @@ export const SnackbarWrapper: React.FC<React.PropsWithChildren<Props>> = functio
   }, []);
 
   return (
-    <div className={s(styles, { notification: true, mounted, closing }, className)}>
+    <div ref={focusedRef} className={s(styles, { notification: true, mounted, closing }, className)}>
```
Wroud: I'm not sure that we want to steal focus right when a notification appears. The original issue is that's impossible to navigate to snackbar with TAB, maybe we just need to add a proper tabIndex?

Wroud (approved):

sergeyteleshev (approved):

PR #2546, opened 2024-04-11
Cb 4629 end user preferences

alexander-skoblikov:


devnaumov (approved):

server/bundles/io.cloudbeaver.model/src/io/cloudbeaver/WebProjectImpl.java
```diff
@@ -23,22 +23,25 @@
...
-
+    private final DBPPreferenceStore preferenceStore;
     public WebProjectImpl(
         @NotNull DBPWorkspace workspace,
         @NotNull RMController resourceController,
         @NotNull SMSessionContext sessionContext,
         @NotNull RMProject project,
-        @NotNull DataSourceFilter dataSourceFilter
+        @NotNull DataSourceFilter dataSourceFilter,
+        DBPPreferenceStore preferenceStore
```
yagudin10: add annotation

yagudin10 (approved):

webapp/packages/plugin-user-profile-settings/src/UserProfileSettings.tsx
```diff
@@ -58,6 +62,9 @@ export const UserProfileSettings = observer(function UserProfileSettings() {
             <ToolsAction icon="admin-cancel" viewBox="0 0 24 24" disabled={!changed} onClick={handleReset}>
               {translate('ui_processing_cancel')}
             </ToolsAction>
+            <ToolsAction icon="admin-cancel" viewBox="0 0 24 24" onClick={handleRestoreDefaults}>
```
sergeyteleshev, line 62: I dont like we have 2 similar icons. This is misleading the user

webapp/packages/plugin-user-profile-settings/src/UserProfileSettings.tsx
```diff
@@ -47,6 +47,10 @@ export const UserProfileSettings = observer(function UserProfileSettings() {
     userSettingsService.resetChanges();
   }
 
+  function handleRestoreDefaults() {
+    userSettingsService.restoreDefaults();
```
sergeyteleshev, line 48: Usually restore settings goes this scenario:
1. press restore settings
2. popup "are you sure you want to restore settings"
3. click yes/no
4. settings applies

Since we have another pipline here, which is fine, by the way. I believe we should add here some kind of notification here that we pasted here default settings but did not save it yet.

Cuase for example: user had first 5 setttings as default. All settings below visible window area were different from default. You clicked this button and user did not see that something were changed. We need to cover this case so user would not have been missleaded. Generally we need to notify somehow that we pasted something successfully

Or we can go always go first scenario

webapp/packages/core-view/src/Menu/MenuService.ts
```diff
@@ -42,20 +41,42 @@ export class MenuService {
...
+    });
   }
 
-  getHandler(context: IDataContextProvider): IMenuHandler | null {
-    for (const handler of this.handlers.values()) {
-      if (handler.isApplicable(context)) {
+  getHandler(contexts: IDataContextProvider): IMenuHandler | null {
+    const menu = contexts.getOwn(DATA_CONTEXT_MENU);
+
+    handlers: for (const handler of this.handlers.values()) {
```
sergeyteleshev, line 66: seems like this logic duplicates in handlig actions and bindings. I beleive we can unify it in function with this cycle. but may be it is not a bad idea to leave it as it is for flexibility reasons

sergeyteleshev (changes requested):

sergeyteleshev (approved):

devnaumov (approved):

webapp/packages/plugin-user-profile-settings/src/UserProfileSettings.tsx
Wroud: fixed

PR #2573, opened 2024-04-23
CB-4980 cancel file upload

devnaumov:


sergeyteleshev (approved):

webapp/packages/core-sdk/src/CustomGraphQLClient.ts
```diff
@@ -156,6 +159,10 @@ export class CustomGraphQLClient extends GraphQLClient {
 
       return response.data;
     } catch (error: any) {
+      if (isCancel(error)) {
+        error.message = 'ui_processing_canceled';
+      }
```
Wroud: seems like it's better to throw new exception

webapp/packages/plugin-data-import/src/DataImportService.ts
```diff
@@ -39,28 +39,51 @@ export class DataImportService {
...
+    let taskId: string | undefined;
+    const abortController = new AbortController();
+
+    const cancel = async () => {
+      if (taskId) {
+        await this.asyncTaskInfoService.cancel(taskId);
+      } else {
+        abortController.abort();
+      }
+    };
```
Wroud: better to have `let cancelImplementation: (() => void) | null = null`
and then set this on every step to proper cancellation function
```
let cancelImplementation: (() => void) | null = null;
function cancel() {
  cancelImplementation?.();
}
```

webapp/packages/plugin-data-import/src/DataImportService.ts
```diff
@@ -39,28 +39,51 @@ export class DataImportService {
...
+
     const { controller, notification } = this.notificationService.processNotification(
       () => ProcessSnackbar,
-      {},
+      {
+        onCancel: cancel,
+      },
       { title: 'plugin_data_import_process_title', message: file.name },
```
Wroud: please add notification's `onClose` pointing to`cancel` fn

Wroud (changes requested):

webapp/packages/core-blocks/src/Snackbars/ProcessSnackbar.tsx
```diff
@@ -54,12 +54,13 @@ export const ProcessSnackbar: NotificationComponent<ProcessSnackbarProps> = obse
     return null;
   }
 
+  function close() {
+    onCancel?.();
+    notification.close(false);
+  }
```
Wroud: notification can be closed programmatically, because of that, you need to listen for notification close event and invoke the close handler

Wroud (approved):

PR #2574, opened 2024-04-24
CB-4873 allow set blob data to null

devnaumov:


sergeyteleshev (approved):

webapp/packages/plugin-data-spreadsheet-new/src/DataGrid/Formatters/CellFormatters/BlobFormatter.tsx
```diff
@@ -35,7 +35,7 @@ export const BlobFormatter = observer<RenderCellProps<IResultSetRowKey>>(functio
   const rawValue = getComputed(() => formatter.get(cell));
   const displayString = getComputed(() => formatter.getDisplayString(cell));
 
-  const nullValue = rawValue === null;
+  const nullValue = isResultSetContentValue(rawValue) ? rawValue.text === 'null' : rawValue === null;
```
Wroud, line 38: seems like backend must return null in that case instead of text (like a bug)

Wroud (approved):

PR #2584, opened 2024-04-30
CB-4931 make export and import dialog focusable

devnaumov:


sergeyteleshev (approved):

webapp/packages/plugin-data-export/src/Dialog/ProcessorSelectDialog.tsx
```diff
@@ -31,7 +32,7 @@ export const ProcessorSelectDialog = observer<Props>(function ProcessorSelectDia
   const processors = dataTransferProcessorsResource.resource.values.slice().sort(sortProcessors);
 
   return (
-    <CommonDialogWrapper size="large" fixedSize>
+    <CommonDialogWrapper ref={focusedRef} size="large" fixedSize>
```
Wroud: seems like it's better to move useFocus to CommonDialogWrapper and add a prop to disable this behaviour if needed (enabled by default) so we can remove this duplication in every dialog

Wroud (changes requested):

webapp/packages/plugin-data-export/src/Dialog/ProcessorSelectDialog.tsx
devnaumov: Yes, and its probably better to focus dialog itself by default and not the first item

Wroud: not always, for example in auth dialog we want first field to be focused

Wroud (approved):

PR #2586, opened 2024-04-30
CB-4698 open value panel for truncated and big texts only

sergeyteleshev:


Wroud:
not covering some cases

PR #2609, opened 2024-05-13
CB-931 ldap identity provider

alexander-skoblikov:


server/bundles/io.cloudbeaver.model/src/io/cloudbeaver/auth/provider/fa/AbstractSessionExternal.java
```diff
@@ -0,0 +1,75 @@
+/*
+ * DBeaver - Universal Database Manager
+ * Copyright (C) 2010-2024 DBeaver Corp
+ *
+ * All Rights Reserved.
+ *
```
LonwoLonwo, line 8: Incorrect license?

alexander-skoblikov: fixed

PR #2645, opened 2024-05-21
CB-5139 domain manager improvements

Wroud:


devnaumov (approved):

webapp/packages/core-ui/src/Form/FormState.ts
```diff
@@ -211,6 +214,7 @@ export class FormState<TState> implements IFormState<TState> {
 
   async save(): Promise<boolean> {
     try {
+      this.isSaving = true;
       this.isDisabled = true;
```
sergeyteleshev: I beleive, we can replace this prop with `isSaving` then all over where it used

webapp/packages/core-ui/src/Form/FormPart.ts
```diff
@@ -89,7 +89,10 @@ export abstract class FormPart<TPartState, TFormState = any> implements IFormPar
         return;
       }
 
-      await this.saveChanges();
+      await this.saveChanges(data, contexts);
+      if (ExecutorInterrupter.isInterrupted(contexts)) {
+        return;
```
sergeyteleshev, line 94: `this.loading` can stuck in `true` state then.

`this.loading = false;` - can be added here

sergeyteleshev (approved):

webapp/packages/core-ui/src/Form/FormPart.ts
Wroud: No, it can't it will be set to false in finally

webapp/packages/core-ui/src/Form/FormState.ts
Wroud: fixed

PR #2646, opened 2024-05-21
CB-4853 updates docker containerId view in update instruction for CE

sergeyteleshev:


webapp/packages/product-default/src/UpdateInstruction.tsx
```diff
@@ -9,7 +9,8 @@ import { Code } from '@cloudbeaver/core-blocks';
 import type { InstructionComponent } from '@cloudbeaver/core-version-update';
 
 export const UpdateInstruction: InstructionComponent = function UpdateInstruction({ version, containerId, className }) {
-  const id = containerId || 'cloudbeaver';
+  const id = `containerId("${containerId || 'cloudbeaver'}")`;
+
```
Wroud: seems weird in the case when containerId is presented and correct (because now the script itself is invalid)
I propose to replace `id` with constant `{your container id}`

Wroud (changes requested):

Wroud (approved):

PR #2647, opened 2024-05-21
CB-4868 extend editor highlight

devnaumov:


Wroud (changes requested):
please also add `tok-labelName`, `tok-className` and `tok-punctuation` seems to be too contrast, maybe we can use different color (not black)
and also `tok-atom` has wrong contrast in black theme

Wroud (approved):

sergeyteleshev (approved):

PR #2670, opened 2024-05-28
Added French translation files (fr.ts) to respective directories

matthieukhl:
This pull request introduces the French translation files (fr.ts) to the relevant directories within the project. Each fr.ts file contains translations for the strings used in the corresponding module or feature of CloudBeaver. By adding these files, I aim to provide French-speaking users with a localized experience, improving accessibility and usability. Please review and merge this pull request to integrate the French translations into the project. Thank you for your attention to this enhancement!


sergeyteleshev:
Hey! Many thanks for your contribution!
One more thing. You need to register each file in localization service for each package. Usually it calls `LocalizationService.ts` but the name can vary a bit.

For instance you have added this localization:
`webapp/packages/core-administration/src/locales/fr.ts`

This means you need to register new locale for `core-administration` package in `LocalizationService` here
`webapp/packages/core-administration/src/AdministrationLocaleService.ts`

You should do it like this:
image

So all in all you should do this procedure for each translation file you had added so it can be applied to the user interface

P.S. updated the guide for localization

sergeyteleshev (changes requested):
Requested changes

sergeyteleshev (approved):

PR #2674, opened 2024-05-30
Cb 5198 unit tests

sergeyteleshev:


webapp/packages/core-utils/src/ILoadableState.test.ts
```diff
@@ -0,0 +1,64 @@
...
+    expect(isContainsException(new Error())).toBe(true);
+  });
+
+  it('should return false if exception is not present', () => {
+    expect(isContainsException(null)).toBe(false);
+  });
+
+  it('should check array of exceptions', () => {
+    expect(isContainsException([null, new Error()])).toBe(true);
+    expect(isContainsException([null, null])).toBe(false);
```
Wroud, line 30: please add empty array check

webapp/packages/core-utils/src/ILoadableState.test.ts
```diff
@@ -0,0 +1,64 @@
...
+  });
+});
+
+describe('getFirstException', () => {
+  it('should return first exception', () => {
+    const error = new Error();
+
+    expect(getFirstException(error)).toBe(error);
+    expect(getFirstException([null, error])).toBe(error);
```
Wroud, line 41: please add at least two different exceptions to an array

webapp/packages/core-utils/src/ILoadableState.test.ts
```diff
@@ -0,0 +1,64 @@
...
+  });
+
+  it('should not get exception', () => {
+    expect(getFirstException([null, null])).toBe(null);
```
Wroud, line 47: please add empty array check

webapp/packages/core-utils/src/ILoadableState.test.ts
```diff
@@ -0,0 +1,64 @@
...
+ * you may not use this file except in compliance with the License.
+ */
+import { getFirstException, ILoadableState, isContainsException, isLoadableStateHasException } from './ILoadableState';
+
+const getMockedLoadableState = (state: Partial<ILoadableState>): ILoadableState => ({
+  exception: new Error(),
+  isLoading: () => false,
+  isLoaded: () => false,
+  isError: () => false,
+  load: () => undefined,
```
Wroud: this doesn't seems to be correct mock, because isError always returns false, but you specified default exception value

probably you want to implement load function that accepts callback and catches excetions and for isError you need to use isContainsException

webapp/packages/core-utils/src/ILoadableState.test.ts
```diff
@@ -0,0 +1,64 @@
...
+    const state: ILoadableState = getMockedLoadableState({ exception: new Error() });
+    const stateWithArray: ILoadableState = getMockedLoadableState({ exception: [null, new Error()] });
+
+    expect(isLoadableStateHasException(state)).toBe(true);
+    expect(isLoadableStateHasException(stateWithArray)).toBe(true);
+  });
+
+  it('should return false if exception is not present', () => {
+    const state: ILoadableState = getMockedLoadableState({ exception: null });
+    const stateWithArray: ILoadableState = getMockedLoadableState({ exception: [null, null] });
```
Wroud: please add empty array check

webapp/packages/core-utils/src/MetadataMap.test.ts
```diff
@@ -0,0 +1,223 @@
...
+      [3, 'three'],
+      [Infinity, 'infinity'],
+      [NaN, 'nan'],
+    ];
+    map.sync(data);
+
+    data.forEach(([key, value]) => {
+      expect(map.get(key)).toBe(value);
+    });
+  });
```
Wroud: seems like a test for sync function rather then for add and get items

Wroud (changes requested):

webapp/packages/core-utils/src/ILoadableState.test.ts
sergeyteleshev: I think you a bit overcomplicating here

We dont need to test integration between functions, it not quite a unit test apporach

Changed `getMockedLoadableState` with null exception. + in test cases added `isError` parameter just in case so the mock is more truthful

But since it is (`isError`) not tested yet it is not important. But we can add some logic here if it is important since cause you mentioned it.

The most important moment here we should implement new rule in our team:
If you have changed helper's/component's/etc logic. You have to also support this new logic with unit tests written for this particular file

Wroud: true

webapp/packages/core-utils/src/MetadataMap.test.ts
```diff
@@ -0,0 +1,237 @@
...
+      [2, 'two'],
+      [3, 'three'],
+    ]);
+
+    try {
+      map.get(1, undefined, z.object({}));
+      expect(true).toBeFalsy(); // should not be called
+    } catch (e) {
+      expect(e).toBeInstanceOf(Error);
+    }
```
Wroud: `expect(() =>  map.get(1, undefined, z.object({}))).toThrowError()`

webapp/packages/core-utils/src/MetadataMap.test.ts
```diff
@@ -0,0 +1,237 @@
...
+
+  it('should get function default value', () => {
+    const map = new MetadataMap<number, string>();
+    map.sync([
+      [1, 'one'],
+      [2, 'two'],
+      [3, 'three'],
+    ]);
+
+    expect(JSON.stringify(map.get(1, key => `default ${key}`, z.object({})))).toStrictEqual(JSON.stringify('default 1'));
```
Wroud: This should not work this way; also, why do you pass an empty zod object?
In this case, `map.get(1, key => `default ${key}`)` should return `one`
also, I can't understand the purpose of using JSON.stringify

Wroud: I see. Please rename the test case because it tests fallback to the default value in case when schema validation failed

Wroud: but still why you use JSON.stringify in such many places?

webapp/packages/core-utils/src/MetadataMap.test.ts
```diff
@@ -0,0 +1,237 @@
...
+  });
+
+  it('should return default value', () => {
```
Wroud: when? why?

webapp/packages/core-utils/src/MetadataMap.test.ts
```diff
@@ -0,0 +1,237 @@
...
+    map.sync([
+      [1, 'one'],
+      [2, 'two'],
+      [3, 'three'],
+    ]);
+
+    expect(map.size).toBe(3);
+  });
+
+  it('should iterate', () => {
```
Wroud: it's not :) But it should return an iterator

webapp/packages/core-utils/src/MetadataMap.test.ts
```diff
@@ -0,0 +1,237 @@
...
+    const map = new MetadataMap<number, string>();
+    const data: [number, string][] = [
+      [1, 'one'],
+      [2, 'two'],
+      [3, 'three'],
+    ];
+    map.sync(data);
+
+    const items = Array.from(map);
```
Wroud: Array.from is not a good method to test iterators

webapp/packages/core-utils/src/MetadataMap.test.ts
```diff
@@ -0,0 +1,237 @@
...
+    map.sync([
+      [1, 'one'],
+      [2, 'two'],
+      [3, 'three'],
+    ]);
+
+    map.delete(4);
+
+    expect(map.size).toBe(3);
+  });
```
Wroud: can be tested as a part of "should delete"

Wroud: You also better check the snapshot of values rather than size (size also is good to check, but it's a weak check if we check only size)

Wroud (changes requested):

webapp/packages/core-utils/src/MetadataMap.ts
```diff
@@ -97,7 +97,12 @@ export class MetadataMap<TKey, TValue> implements Map<TKey, TValue> {
       }
 
       const value = provider(key, this);
-      this.temp.set(key, observable(value as any));
+
+      if (typeof value === 'object' && value !== null) {
+        this.temp.set(key, observable(value));
+      }
+
+      this.temp.set(key, value);
```
Wroud: this will lead to double set and final value will be always observer, probably you forgot to `return`

Wroud (changes requested):

Wroud (approved):

PR #2690, opened 2024-06-05
CB-4705 make data viewer model loaders more flexible

devnaumov:


webapp/packages/plugin-data-viewer/src/useDataViewerModel.ts
```diff
@@ -0,0 +1,119 @@
...
+        return this._exception;
+      },
+      isLoading(): boolean {
+        return connection.isLoading() || this._loading;
+      },
+      isLoaded(): boolean {
+        return connection.isLoaded() && this.dataViewerTable.get(this.tab.handlerState.tableId || '') !== undefined;
+      },
+      isError(): boolean {
+        return this.exception !== null;
```
sergeyteleshev: exception is also can be undefined

I believe something like this is more suitable here:
`return isNotNullDefined(this.exception)`

webapp/packages/plugin-data-viewer/src/ContainerDataSource.ts
```diff
@@ -233,7 +250,7 @@ export class ContainerDataSource extends ResultSetDataSource<IDataContainerOptio
     }));
   }
 
-  private async ensureContextCreated(): Promise<IConnectionExecutionContext> {
+  private async getContext(): Promise<IConnectionExecutionContext> {
```
Wroud: please revert, this is not obvious that get* function can create context

webapp/packages/plugin-data-viewer/src/DataViewerTable.ts
```diff
@@ -0,0 +1,39 @@
...
+ * Licensed under the Apache License, Version 2.0.
+ * you may not use this file except in compliance with the License.
+ */
+import type { Connection } from '@cloudbeaver/core-connections';
+
+import type { IDatabaseDataModel } from './DatabaseDataModel/IDatabaseDataModel';
+import type { IDatabaseResultSet } from './DatabaseDataModel/IDatabaseResultSet';
+import type { IDataViewerTableStorage } from './IDataViewerTableStorage';
+
+export abstract class DataViewerTable {
```
Wroud: should not exists

webapp/packages/plugin-data-viewer/src/DataViewerTableService.ts
```diff
@@ -17,10 +17,11 @@ import { DatabaseDataAccessMode } from './DatabaseDataModel/IDatabaseDataSource'
 import type { IDatabaseResultSet } from './DatabaseDataModel/IDatabaseResultSet';
 import { DataViewerService } from './DataViewerService';
 import { DataViewerSettingsService } from './DataViewerSettingsService';
+import { DataViewerTable } from './DataViewerTable';
 import { TableViewerStorageService } from './TableViewer/TableViewerStorageService';
 
 @injectable()
-export class DataViewerTableService {
+export class DataViewerTableService extends DataViewerTable {
```
Wroud: please revert

sergeyteleshev (approved):

webapp/packages/plugin-data-viewer/src/useDataViewerModel.ts
```diff
@@ -0,0 +1,119 @@
...
+  dataViewerTable: DataViewerTable;
+  tab: Tab;
+  _exception?: Error[] | Error | null;
+  _loading: boolean;
+  _init(): Promise<void>;
+  init(tab: Tab, state: State): Promise<void>;
+  load(): Promise<void>;
+}
+
+export function useDataViewerModel<T extends Tab>(tab: T, dataViewerTable: DataViewerTable, init: (tab: T, state: State) => Promise<void>) {
```
Wroud: DatabaseDataModel should not be associated with Tabs, please pass needed information as arguments to this hook, you don't need DataViewerTable here, you can use `TableViewerStorageService` directly

to init function you can pass DatabaseDataModel and probably you don't need to pass state

webapp/packages/plugin-data-viewer/src/DataViewerTableService.ts
```diff
@@ -30,23 +31,8 @@ export class DataViewerTableService {
...
-  }
-
-  async removeTableModel(tableId: string): Promise<void> {
-    const model = this.tableViewerStorageService.get(tableId);
-
-    if (model) {
-      this.tableViewerStorageService.remove(tableId);
-      await model.dispose();
-    }
+  ) {
```
Wroud: seems like we must remove this methods from `DataViewerTableService` and `CollectionViewerTableService` (and maybe other places) and use `TableViewerStorageService` instead

webapp/packages/plugin-data-viewer/src/DataViewerPage/useDataViewerPanel.ts
```diff
@@ -0,0 +1,75 @@
...
+  const model = useDataViewerModel(objectViewerTab, dataViewerTableService, async (tab, state) => {
+    const node = navNodeManagerService.getNode({
+      nodeId: tab.handlerState.objectId,
+      parentId: tab.handlerState.parentId,
+    });
+
+    if (!navNodeManagerService.isNodeHasData(node)) {
+      state.resetException();
+      return;
+    }
```
Wroud: should be handled by useDataViewerModel. Don't call `state.resetException();` in such contexts (if init not throw an exception, then it should be cleared by `useDataViewerModel`)

webapp/packages/plugin-data-viewer/src/ContainerDataSource.ts
```diff
@@ -217,6 +182,58 @@ export class ContainerDataSource extends ResultSetDataSource<IDataContainerOptio
     return prevResults;
   }
 
+  getResultId(prevResults: IDatabaseResultSet[], context: IConnectionExecutionContextInfo) {
```
Wroud: They all public what is the reason for it?

Wroud (changes requested):

webapp/packages/plugin-data-viewer/src/useDataViewerModel.ts
Wroud: right, but as long as it implements ILoadableState it's better to use `isContainsException` cuz it accepts all variants of exception value

webapp/packages/plugin-data-viewer/src/DataViewerPage/DataViewerPanel.tsx
```diff
@@ -8,16 +8,17 @@
 import { observer } from 'mobx-react-lite';
 import { useCallback } from 'react';
 
-import { Loader, TextPlaceholder } from '@cloudbeaver/core-blocks';
+import { Loader, TextPlaceholder, useAutoLoad, useTranslate } from '@cloudbeaver/core-blocks';
```
sergeyteleshev: pls remove unused loader here

webapp/packages/plugin-data-viewer/src/DataViewerPage/useDataViewerPanel.ts
```diff
@@ -0,0 +1,80 @@
...
+        await connectionInfoResource.waitLoad();
+        const connectionInfo = connectionInfoResource.get(tab.handlerState.connectionKey!);
+
+        if (!connectionInfo) {
+          throw new Error("Connection doesn't exists");
+        }
+
+        model = dataViewerTableService.create(connectionInfo, node);
+        tab.handlerState.tableId = model.id;
+        model.source.setOutdated();
```
sergeyteleshev, line 58: why do we need it to set it to outdated here? may be we should incapsulate this logic inside 
`dataViewerTableService.create()` function?

sergeyteleshev (approved):

Wroud (approved):

webapp/packages/plugin-data-viewer/src/DataViewerPage/useDataViewerPanel.ts
devnaumov: I think its better when create fn just create the model without any side effects

PR #2696, opened 2024-06-10
dbeaver/pro#2830 Resource rename

E1izabeth:
When rename happens, two events are send - delete and create. We need additional information to know, that it's a rename operation.


LonwoLonwo (approved):

ShadelessFox (approved):

alexander-skoblikov (approved):

server/bundles/io.cloudbeaver.service.rm/src/io/cloudbeaver/service/rm/impl/WebServiceRM.java
```diff
@@ -205,13 +205,17 @@ public boolean moveResource(@NotNull WebSession webSession,
                 webSession,
                 oldResourcePath,
                 WSConstants.EventAction.DELETE,
-                WSResourceProperty.NAME);
+                WSResourceProperty.NAME,
+                newResourcePath
```
alexander-skoblikov, line 209: it's can be unclear that there's a modified path in the details.

E1izabeth:
We don't want to handle move operation, we want to show the notificaton, that the resource was deleted, if user is working on these resource right now.

PR #2708, opened 2024-06-13
Refactor cb-4018 data context api

Wroud:


webapp/packages/core-data-context/src/DataContext/DataContext.ts
```diff
@@ -37,119 +36,121 @@ export class DataContext implements IDataContext {
...
     if (this.hasOwn(context)) {
       return true;
     }
 
-    if (nested && this.fallback?.has(context)) {
+    if (this.fallback?.has(context)) {
       return true;
     }
 
     return false;
```
sergeyteleshev:
```suggestion
    return this.hasOwn(context) || this.fallback?.has(context);
```

webapp/packages/core-data-context/src/DataContext/DataContext.ts
```diff
@@ -37,119 +36,121 @@ export class DataContext implements IDataContext {
...
-      return this.delete.bind(this, context, version);
+    if (!data) {
+      data = observable(new Map(), { deep: false });
+      this.store.set(context, data);
     }
 
-    version++;
-    this.map.set(context, value);
-    this.versions.set(context, version);
+    data.set(id, value);
```
sergeyteleshev, line 76: in case where `data` was already defined it misses an if statement above where you set new context to the shallow observable `store`

this means that mobx wont be triggered if value changed in data, right? should we just move this line to the end of the function?

`this.store.set(context, data);`

webapp/packages/core-data-context/src/DataContext/DataContext.ts
```diff
@@ -37,119 +36,121 @@ export class DataContext implements IDataContext {
...
-  delete(context: DataContextGetter<any>, version?: number): this {
-    if (version !== this.versions.get(context)) {
-      return this;
+  deleteForId(id: string): this {
+    const ids = new Set<DataContextGetter<unknown>>();
+    for (const [context, data] of this.store) {
+      data.delete(id);
+
+      if (data.size === 0) {
+        ids.add(context);
```
sergeyteleshev: why not just `this.store.delete(context);`?

webapp/packages/core-data-context/src/DataContext/useDataContext.ts
```diff
@@ -5,16 +5,18 @@
...
 import type { IDataContext } from './IDataContext';
 import type { IDataContextProvider } from './IDataContextProvider';
-import { TempDataContext } from './TempDataContext';
 
 export function useDataContext(fallback?: IDataContextProvider): IDataContext {
-  const [context] = useState(() => new TempDataContext());
+  const [context] = useState(() => new DataContext());
 
-  context.setFallBack(fallback);
+  useLayoutEffect(() => {
```
sergeyteleshev, line 17: do we need to set it exactly right after all DOM manipulations being done? if it works correctly without the hook, I suggest to delete this

webapp/packages/core-data-context/src/DataContext/useDataContextLink.ts
```diff
@@ -0,0 +1,29 @@
...
+import { useLayoutEffect, useState } from 'react';
+
+import { uuid } from '@cloudbeaver/core-utils';
+
+import type { IDataContext } from './IDataContext';
+
+export function useDataContextLink(context: IDataContext | undefined, update: (context: IDataContext, id: string) => void): void {
+  const [id] = useState(() => uuid());
+
+  useLayoutEffect(() => {
```
sergeyteleshev: same for this one bad boy. I tried to remove both `useLayoutEffects` and at first glance the app works fine (did not invest many time to explore the whole app).

webapp/packages/core-data-context/src/DataContext/useDataContextLink.ts
```diff
@@ -0,0 +1,29 @@
...
+    if (context) {
+      update(context, id);
```
sergeyteleshev: in each case you use this hook you always pass this `id` and it seems like it is mandatory so `DataContext` works correctly

I suggest to incapsulate this id inside here and always pass it like this `context.set(..., id)`. Cause  it is easy to forget add it and then hard to find the bug

I suggest to create here a `localContext` with the same API interface `IDataContext`. Which is the same copy of `context` object but the setter is involves `id` from the `useState` something like this:

`function set() {
    context.set(..., id) // id from the useState 
}`

sergeyteleshev (changes requested):

webapp/packages/core-data-context/src/DataContext/DataContext.ts
```diff
@@ -37,119 +36,121 @@ export class DataContext implements IDataContext {
...
+    return value;
   }
 
   clear(): void {
-    this.map.clear();
-    this.versions.clear();
+    this.store.clear();
+  }
+
+  private internalGet<T>(context: DataContextGetter<T>): T | typeof NOT_FOUND {
```
devnaumov, line 134: _get ?

webapp/packages/core-data-context/src/DataContext/DataContextGetter.ts
```diff
@@ -5,9 +5,11 @@
  * Licensed under the Apache License, Version 2.0.
  * you may not use this file except in compliance with the License.
  */
-import type { IDataContextProvider } from './IDataContextProvider';
+
+const typescriptTypeLink = Symbol('typescript type link');
```
devnaumov, line 9: Looks strange. Is there a way to avoid this?

webapp/packages/core-data-context/src/DataContext/useDataContext.ts
devnaumov: Also its missing deps

webapp/packages/core-data-context/src/DataContext/DataContext.ts
Wroud: no, here is all fine, to access data mobx will track both Maps and when we change value all subscribers to that value will be notified

webapp/packages/core-data-context/src/DataContext/useDataContextLink.ts
```diff
@@ -0,0 +1,29 @@
...
+    }
+  });
+
+  useLayoutEffect(
```
devnaumov, line 26: Please fix lint errors

webapp/packages/core-navigation-tree/src/NodesManager/getNodesFromContext.ts
```diff
@@ -12,8 +12,8 @@ import { DATA_CONTEXT_NAV_NODES } from './DATA_CONTEXT_NAV_NODES';
 import type { NavNode } from './EntityTypes';
 
 export function getNodesFromContext(context: IDataContextProvider): NavNode[] {
-  const node = context.tryGet(DATA_CONTEXT_NAV_NODE);
-  const getNodes = context.tryGet(DATA_CONTEXT_NAV_NODES);
+  const node = context.get(DATA_CONTEXT_NAV_NODE);
+  const getNodes = context.get(DATA_CONTEXT_NAV_NODES);
```
devnaumov, line 16: Seems like getNodes is always defined now

webapp/packages/plugin-authentication-administration/src/Administration/Users/UsersTable/CreateUserBootstrap.ts
```diff
@@ -48,7 +48,7 @@ export class CreateUserBootstrap extends Bootstrap {
       },
       isDisabled: (context, action) => {
         if (action === ACTION_CREATE) {
-          const administrationItemRoute = context.tryGet(DATA_CONTEXT_ADMINISTRATION_ITEM_ROUTE);
+          const administrationItemRoute = context.get(DATA_CONTEXT_ADMINISTRATION_ITEM_ROUTE);
```
devnaumov, line 51: administrationItemRoute is defined now, not need for the ?.

webapp/packages/plugin-connections/src/ContextMenu/ConnectionMenuBootstrap.ts
```diff
@@ -68,13 +68,13 @@ export class ConnectionMenuBootstrap extends Bootstrap {
           return false;
         }
 
-        const connection = context.tryGet(DATA_CONTEXT_CONNECTION);
+        const connection = context.get(DATA_CONTEXT_CONNECTION);
```
devnaumov, line 71: Please check ?. here and all the way down where you change tryGet -> get

webapp/packages/core-data-context/src/DataContext/useDataContext.ts
Wroud: without hook it will lead to "bad setState during rendering of component"

webapp/packages/core-data-context/src/DataContext/useDataContextLink.ts
Wroud: This is too complicated, but we can discuss it in the future. This is not hard to implement, but we must support this code. I'm not sure this is justified.

webapp/packages/core-data-context/src/DataContext/DataContext.ts
Wroud: fixed

webapp/packages/plugin-user-profile/src/UserMenu/UserMenu.tsx
```diff
@@ -21,16 +22,19 @@ export const UserMenu = observer(function UserMenu() {
   const styles = useS(style);
   const authInfoService = useService(AuthInfoService);
   const menu = useMenu({ menu: MENU_USER_PROFILE });
+  const userInfo = authInfoService.userInfo;
```
devnaumov, line 25: Seems like this pattern is not stable as the developer can easily pass authInfoService.userInfo directly to the callback fn without reassigning this first

webapp/packages/core-data-context/src/DataContext/DataContext.ts
Wroud: fixed

devnaumov (changes requested):

webapp/packages/core-data-context/src/DataContext/DataContext.ts
Wroud: don't like this prefixes

sergeyteleshev (approved):

webapp/packages/core-navigation-tree/src/NodesManager/getNodesFromContext.ts
Wroud: type of `get` changed from `T` to `T | undefined` and equal to type of `tryGet`. so it's basically `tryGet` replaced `get`

webapp/packages/plugin-user-profile/src/UserMenu/UserMenu.tsx
Wroud: fixed

webapp/packages/core-data-context/src/DataContext/useDataContextLink.ts
Wroud: fixed

webapp/packages/core-data-context/src/DataContext/DataContextGetter.ts
Wroud: I don't know any way to keep the generic type without using it in the object

devnaumov (approved):

PR #2711, opened 2024-06-14
CB-4122. Rare example for gql multipart

DenisSinelnikov:


Wroud:
We did an investigation and are not planning to implement this approach

PR #2730, opened 2024-06-24
CB-5181 Display User Notification Pop-up When Copy/Paste Functionality is Disabled by Administrator

sergeyteleshev:


Wroud (changes requested):
it seems like copy data from the table is not covered by this PR

Wroud (approved):

webapp/packages/plugin-data-viewer/src/locales/it.ts
```diff
@@ -50,6 +50,8 @@ export default [
   ['data_viewer_total_count_canceled_message', 'Statement was cancelled due to user request'],
   ['data_viewer_total_count_failed', 'Failed to get total count'],
   ['data_viewer_model_not_loaded', 'Table model is not loaded'],
+  ['data_viewer_copy_not_allowed', 'An ability to copy data is disabled'],
+  ['data_viewer_copy_not_allowed_message', 'Please allow the "Preferences" to change it or ask your admin'],
```
devnaumov: If its an admin, he always can copy data, so no need to mention preferences here, i would suggest remove message completely or rephrase it like this: "If this is unexpected, contact the administrator"

devnaumov (changes requested):

webapp/packages/plugin-data-viewer/src/locales/it.ts
sergeyteleshev: This text was required by the ticket

devnaumov (approved):

PR #2733, opened 2024-06-25
Cb 5124 add a notification about the upcoming expiration of the license add to status

DenisSinelnikov:


Wroud (approved):

server/bundles/io.cloudbeaver.server/src/io/cloudbeaver/server/servlets/CBStatusServlet.java
```diff
@@ -44,6 +46,12 @@ protected void doGet(HttpServletRequest request, HttpServletResponse response) t
         infoMap.put("health", "ok");
         infoMap.put("product.name", GeneralUtils.getProductName());
         infoMap.put("product.version", GeneralUtils.getProductVersion().toString());
+        try {
+            CBApplication.getInstance().getStatusInfo(infoMap);
+        } catch (DBException e) {
+            log.error(e);
+            response.sendError(HttpServletResponse.SC_BAD_REQUEST, "Error additional status info : " + e.getMessage());
```
alexander-skoblikov: we should not send an error in this request

alexander-skoblikov (changes requested):

webapp/packages/core-bootstrap/package.json
```diff
@@ -52,6 +52,7 @@
     "@cloudbeaver/core-version": "^0",
     "@cloudbeaver/core-version-update": "^0",
     "@cloudbeaver/core-view": "^0",
+    "@cloudbeaver/core-website": "^0",
```
Wroud, line 55: it's not a core it's a plugin

webapp/packages/core-website/src/WebsiteLinksService.ts
```diff
@@ -0,0 +1,37 @@
...
+  }
+
+  getProductBuyPage(distributed: boolean) {
+    if (distributed) {
+      return WebsiteLinksService.TEAM_EDITION_BUY_PRODUCT_PAGE;
+    }
+
+    return WebsiteLinksService.ENTERPRISE_BUY_PRODUCT_PAGE;
+  }
+}
```
Wroud: seems like it's not a service but completely static class, you can make methods also static

webapp/packages/core-blocks/package.json
```diff
@@ -39,10 +39,13 @@
   },
   "peerDependencies": {},
   "devDependencies": {
+    "@cloudbeaver/core-di": "^0",
+    "@cloudbeaver/core-dialogs": "^0",
```
Wroud, line 43: a bit strange that we need `"@cloudbeaver/core-dialogs": "^0",` in core-blocks (because core-dialogs depends on core-blocks)

Wroud (changes requested):

webapp/packages/core-bootstrap/package.json
Wroud: now i see the problem, that we can't use it in our core-* so leave it as it is for now

Wroud (approved):

alexander-skoblikov (approved):

PR #2755, opened 2024-07-02
CB-5340 adds unit tests for basic hooks

sergeyteleshev:


webapp/packages/core-blocks/src/useObservableRef.ts
```diff
@@ -70,12 +70,12 @@ export function useObservableRef<T extends Record<any, any>>(
       update = undefined;
     }
 
-    state = makeObservable(state, observed, { deep: false, name });
-
     if (bind) {
       bindFunctions(state, bind as []);
     }
 
+    state = makeObservable(state, observed, { deep: false, name });
```
sergeyteleshev: this one moved cause `makeObservable` freeze methods (`Object.freeze`) in the `state`
so this makes jest falls in `bindFunction` due to the fact we literally reasign methods with its binded version

webapp/packages/core-blocks/src/useControlledScroll.test.ts
```diff
@@ -0,0 +1,98 @@
...
+  it('should not set scroll position if element is null', () => {
+    renderHook(() => useControlledScroll(null, scrollState));
+
+    jest.runAllTimers();
+
+    // No errors should be thrown
+    expect(true).toBe(true);
+  });
+
+  it('should remove event listener if element becomes null', () => {
```
Wroud, line 128: please add case for changing element

webapp/packages/core-blocks/src/useObjectRef.test.ts
```diff
@@ -0,0 +1,130 @@
...
+    const { result } = renderHook(() =>
+      useObjectRef({
+        count: 0,
+      }),
+    );
+
+    result.current.count = 123;
+
+    expect(result.current.count).toBe(123);
+  });
```
Wroud: not sure that this test is needed or if it's correct

webapp/packages/core-blocks/src/useObjectRef.test.ts
```diff
@@ -0,0 +1,130 @@
...
+    anotherContextObject.increment();
+
+    expect(result.current.count).toBe(0);
+    expect(anotherContextObject.count).toBe(124);
+  });
+
+  test('should update ref', () => {
+    const { result, rerender } = renderHook(
+      ({ update }: { update: { count: number; increment?: (this: { count: number }) => void } }) =>
+        useObjectRef(() => ({ count: 0 }), update, ['increment'] as const),
```
Wroud: impossible to read

webapp/packages/core-blocks/src/useObjectRef.test.ts
```diff
@@ -0,0 +1,130 @@
...
+    rerender({
+      update: {
+        count: 1,
+        increment: result.current.increment,
+      },
+    });
+
+    result.current?.increment?.();
+
+    expect(result.current.count).toBe(1);
```
Wroud: can't understand the logic of this test, probably here is mixed logic (not only for ref update, but also ???)

webapp/packages/core-blocks/src/useObservableRef.test.ts
```diff
@@ -0,0 +1,127 @@
...
+  test('updates the object', () => {
+    const init = () => ({ count: 0 });
+    const observed = { count: observable };
+    const update = { count: 1 };
+
+    const { result } = renderHook(() => useObservableRef(init, observed, update));
+
+    expect(result.current.count).toBe(1);
+    expect(isObservable(result.current)).toBe(true);
+  });
```
Wroud: same here not sure that this test is necessary

webapp/packages/core-blocks/src/useObservableRef.test.ts
```diff
@@ -0,0 +1,127 @@
...
+    const init = () => ({ count: 0, text: 'hello' });
+    const observed = { count: observable, text: observable };
+    const update = { count: 1 };
+
+    const { result } = renderHook(() => useObservableRef(init, observed, update));
+
+    expect(result.current.count).toBe(1);
+    expect(result.current.text).toBe('hello');
+    expect(isObservable(result.current)).toBe(true);
+  });
```
Wroud: this test is not about update but about merging init and update data on first render, so we can be sure that first time we access data its present and correct

webapp/packages/core-blocks/src/useObservableRef.test.ts
```diff
@@ -0,0 +1,127 @@
...
+    expect(result.current.count).toBe(1);
+
+    const anotherContext = { count: 0, increment: result.current.increment };
+
+    anotherContext.increment();
+
+    expect(anotherContext.count).toBe(0);
+    expect(result.current.count).toBe(2);
+  });
+});
```
Wroud: same here it's not an update logic, please add separate tests for update logic (with calling `rerender`)

webapp/packages/core-blocks/src/useObservableRef.ts
Wroud: it's incorrect because now if you will test it on update (`rerender`) it will fail and we will get different behaviour for init / update

Wroud (changes requested):
generally it's not practical to test hooks without `rerender` method because this will skip most of react logic and possible bugs

webapp/packages/core-blocks/src/TimerIcon.test.tsx
```diff
@@ -5,7 +5,8 @@
  * Licensed under the Apache License, Version 2.0.
  * you may not use this file except in compliance with the License.
  */
-import { queryByAttribute, waitFor } from '@testing-library/react';
+import { describe, expect, it } from '@jest/globals';
+import { queryByAttribute, screen, waitFor } from '@testing-library/react';
```
Wroud: screen is not used

webapp/packages/core-blocks/src/useControlledScroll.test.ts
```diff
@@ -0,0 +1,136 @@
...
+import { IScrollState, useControlledScroll } from './useControlledScroll';
+
+describe('useControlledScroll', () => {
+  let element: HTMLDivElement;
+  let scrollState: IScrollState;
+
+  beforeEach(() => {
+    jest.resetAllMocks();
+    jest.useFakeTimers();
+    element = document.createElement('div');
```
Wroud, line 25: is this element destroyed properly after each test?

webapp/packages/core-blocks/src/useExecutor.test.ts
```diff
@@ -0,0 +1,137 @@
...
+// import type { IExecutorHandler, IExecutorHandlersCollection } from '@cloudbeaver/core-executor';
+
+// import { useExecutor } from './useExecutor';
+// import { useObjectRef } from './useObjectRef';
+
+// jest.mock('./useObjectRef', () => ({
+//   useObjectRef: jest.fn(obj => obj),
+// }));
+
+describe('useExecutor', () => {
```
Wroud: describe.skip???

webapp/packages/core-cli/configs/jest.config.js
```diff
@@ -36,4 +36,6 @@ module.exports = {
     // https://mswjs.io/docs/migrations/1.x-to-2.x#cannot-find-module-mswnode-jsdom
     customExportConditions: [''],
   },
+  // unset all global variables
+  globals: {},
```
Wroud: it's `{}` by default, to disable globals you need to add `injectGlobals: false`

Wroud (changes requested):

webapp/packages/core-blocks/src/useControlledScroll.test.ts
sergeyteleshev: it kills the dom after all test cases in the file were executed
but after each test case it does not remove element from the dom. so I modified the test so it will remove this div element. thanks

webapp/packages/core-cli/configs/jest.config.js
sergeyteleshev: unfortunatelly it leads to 83 of 83 test to being fallen cause something is wrong with imported `expect`. Even with all cleared caches and other stuff. This moduled `jest` is really painful to use (impossible it just dont work like supposed to) and I hope we gonna move to the next framework during coming tech sprint

sergeyteleshev: for now IDE is not allowing to use globals. so I beleive we are good to go here

webapp/packages/core-blocks/src/TimerIcon.test.tsx
sergeyteleshev: fixed

webapp/packages/core-blocks/src/useExecutor.test.ts
sergeyteleshev: fixed

Wroud (approved):

PR #2774, opened 2024-07-10
CB-3823 refactor: introduce service provider

Wroud:


devnaumov (approved):

sergeyteleshev (approved):

webapp/packages/core-connections/src/ConnectionToolsResource.ts
```diff
@@ -0,0 +1,73 @@
...
+      });
+
+      if (connectionId && !connections.some(connection => connection.id === connectionId)) {
+        throw new Error(`Connection is not found (${connectionId})`);
+      }
+
+      connectionsList.push(...connections);
+    });
+
+    runInAction(() => {
```
sergeyteleshev: why `runInAction`? seems like `this.set(key, connectionsList);` is already an action

webapp/packages/core-di/src/IServiceProvider.ts
```diff
@@ -0,0 +1,16 @@
+/*
+ * CloudBeaver - Cloud Database Manager
+ * Copyright (C) 2020-2024 DBeaver Corp and others
+ *
+ * Licensed under the Apache License, Version 2.0.
+ * you may not use this file except in compliance with the License.
+ */
+import type { IServiceConstructor, IServiceInjector } from './IApp';
+
+export class IServiceProvider {
```
sergeyteleshev, line 10: `IServiceProvider` reads like an interface when in fact it is a class. This missleads a bit. Simply `ServiceProvider` seems okay as for me

webapp/packages/core-di/src/useService.ts
```diff
@@ -7,23 +7,21 @@
...
 
-  if (optional && !app.hasServiceByClass(ctor)) {
-    return undefined;
+  if (optional) {
+    try {
+      return serviceProvider.getService(ctor);
+    } catch {
+      return undefined;
+    }
   }
```
sergeyteleshev, line 24: I'm wondering what is this construction stands for?
It is used for `ThemeService` only.

It used like this:
`const filteredStyles = themeService
    ? // @ts-ignore
      themeService.mapStyles(componentStyles.flat(Infinity).filter(Boolean) as Style[], context)
    : (componentStyles.flat(Infinity).filter(Boolean) as Style[]);`

As I can see, we do not map styles to apply `Registries` if `ThemeService` is not available in the DI container yet.

Should we solve this issue with plugin order in `product-*type*` packages? I think if we delete this logic we can avoid some bugs here

webapp/packages/plugin-data-spreadsheet-new/src/DataGrid/DataGridContextMenu/DataGridContextMenuFilter/DataGridContextMenuFilterService.ts
```diff
@@ -80,7 +83,8 @@ export class DataGridContextMenuFilterService {
     isHidden?: (context: IMenuContext<IDataGridCellMenuContext>) => boolean,
   ): Array<IContextMenuItem<IDataGridCellMenuContext>> {
     const { model, resultIndex, key } = context.data;
-    const data = model.source.getAction(resultIndex, ResultSetDataAction);
+    const source = model.source as unknown as ResultSetDataSource;
```
sergeyteleshev, line 86: Now I can see that in a lot of places, you have to do the casting, which seems like a hack but more like a solution
Is there a way to fix types so you don't need to do that in a lot of places?

webapp/packages/plugin-data-viewer/src/DatabaseDataModel/DatabaseDataSource.ts
```diff
@@ -354,29 +333,25 @@ export abstract class DatabaseDataSource<TOptions, TResult extends IDatabaseData
...
-    return this.request(this.results).then(data => {
-      this.outdated = false;
-
-      if (data !== null) {
-        this.setResults(data);
-      }
-      return data;
-    });
+    return this.request(this.results)
+      .finally(() => {
```
sergeyteleshev, line 346: so if an error occured this is also gonna be outdated. is this expected?

webapp/packages/plugin-data-viewer/src/TableViewer/TableHeader/useWhereFilter.ts
```diff
@@ -24,32 +27,42 @@ interface IState {
...
-export function useWhereFilter(model: IDatabaseDataModel<IDatabaseDataOptions, any>, resultIndex: number): Readonly<IState> {
+export function useWhereFilter(model: IDatabaseDataModel, resultIndex: number): Readonly<IState> {
   return useObservableRef(
     () => ({
+      get supported() {
+        return isResultSetDataSource(this.model.source);
+      },
       get filter() {
+        const source = this.model.source;
+        if (!isResultSetDataSource<IDatabaseDataOptions>(source)) {
```
sergeyteleshev, line 38: `!this.supported`

webapp/packages/plugin-data-viewer/src/TableViewer/TableHeader/useWhereFilter.ts
```diff
@@ -24,32 +27,42 @@ interface IState {
...
         const supported = this.constraints?.supported ?? false;
         return !supported || this.model.isLoading() || this.model.isDisabled(resultIndex);
       },
       get applicableFilter() {
-        return (
-          this.model.source.prevOptions?.whereFilter !== this.model.source.options?.whereFilter ||
-          this.model.source.options?.whereFilter !== this.model.source.requestInfo.requestFilter
-        );
+        const source = this.model.source;
+        if (!isResultSetDataSource<IDatabaseDataOptions>(source)) {
```
sergeyteleshev, line 62: `!this.supported`

webapp/packages/core-di/src/IServiceProvider.ts
Wroud: it's intent to tell developer that it's not concrete implementation, but "type" of service that will be resolved based on context

server/bundles/io.cloudbeaver.model/src/io/cloudbeaver/model/WebConnectionInfo.java
```diff
@@ -244,6 +245,13 @@ public String[] getFeatures() {
 
         if (dataSourceContainer.isConnected()) {
             features.add("connected");
+            var contributedEditors = DBWorkbench.getPlatform().getDataSourceProviderRegistry().getContributedEditors(
+                DBPEditorContribution.MB_CONNECTION_EDITOR,
+                dataSourceContainer
+            );
+            if (SMUtils.isRMAdmin(session) && contributedEditors.length > 0) {
+                features.add("hasTools");
```
serge-rider: Move to constants

webapp/packages/core-di/src/useService.ts
Wroud: It's intended to fall to undefined in case we can't initialize the service, for example, when a component is used for displaying errors for the app initialization phase (where the theme service cannot be initialized yet)

server/bundles/io.cloudbeaver.model/src/io/cloudbeaver/model/WebConnectionInfo.java
```diff
@@ -475,4 +483,13 @@ public List<WebSecretInfo> getSharedSecrets() throws DBException {
             .collect(Collectors.toList());
     }
 
+    @Property
+    public List<String> getTools() {
```
serge-rider, line 494: getContributedEditors are editors for desktop products.
How does it work?

serge-rider (changes requested):

webapp/packages/plugin-data-viewer/src/DatabaseDataModel/DatabaseDataSource.ts
Wroud: in case we get data or exceptions we mark datasource as NOT outdated, because we now have results and don't need to load it again

webapp/packages/plugin-data-spreadsheet-new/src/DataGrid/DataGridContextMenu/DataGridContextMenuFilter/DataGridContextMenuFilterService.ts
Wroud: Screenshot 2024-07-15 at 9 42 19 PM
in part of these places i don't repeat checks because it's exists already

webapp/packages/plugin-data-viewer/src/TableViewer/TableHeader/useWhereFilter.ts
Wroud: needed as type guard

webapp/packages/plugin-data-viewer/src/TableViewer/TableHeader/useWhereFilter.ts
Wroud: needed as type guard

webapp/packages/core-connections/src/ConnectionToolsResource.ts
Wroud: fixed, it's from copy paste

sergeyteleshev (approved):

PR #2795, opened 2024-07-22
CB-5415. Added full name header for RP

DenisSinelnikov:


server/bundles/io.cloudbeaver.model/plugin.xml
```diff
@@ -15,6 +15,7 @@
             <propertyGroup label="Main">
                 <property id="firstName" label="First Name" type="string" description="User first name"/>
                 <property id="lastName" label="Last Name" type="string" description="User last name"/>
+                <property id="fullName" label="Full Name" type="string" description="Full name"/>
```
LonwoLonwo:
```suggestion
                <property id="fullName" label="Full Name" type="string" description="User full name"/>
```

LonwoLonwo: Optional

server/bundles/io.cloudbeaver.service.security/src/io/cloudbeaver/auth/provider/rp/RPAuthProvider.java
```diff
@@ -92,6 +94,11 @@ public DBWUserIdentity getUserIdentity(
             userMeta.put(SMStandardMeta.META_LAST_NAME, lastName);
         }
 
+        if (CommonUtils.isNotEmpty(fullName)) {
+            nameBuilder = new StringBuilder(fullName);
```
alexander-skoblikov, line 98: this doesn't make sense since we don't support this meta information on UI side, we should use only first/last name

alexander-skoblikov (changes requested):

alexander-skoblikov (approved):

PR #2797, opened 2024-07-22
CB-4823 Result set - no vertical and horizontal scrolls appear for long errors

sergeyteleshev:


webapp/packages/plugin-data-viewer/src/TableViewer/TableError.module.css
```diff
@@ -28,7 +28,6 @@
 .errorBody {
   display: flex;
   gap: 24px;
-  align-items: center;
```
sergeyteleshev, line 31: it looks weird if everything is centered. In case the error is huge and scrollable, the error icon is in the middle, which is not visible for the user to indicate it is an error. image with the fixed case attached

Снимок экрана 2024-07-22 в 21 51 59

webapp/packages/plugin-data-viewer/src/TableViewer/TableError.module.css
```diff
@@ -63,7 +62,6 @@
 }
 
 .animated {
-  overflow: hidden;
```
sergeyteleshev, line 66: Animation is saved its behavior

Wroud (approved):

devnaumov (approved):

PR #2808, opened 2024-07-29
Cb 5378 dm ability to use custom ssl certificate

sergeyteleshev:


webapp/packages/core-utils/src/getDomainFromUrl.ts
```diff
@@ -0,0 +1,17 @@
...
+ * you may not use this file except in compliance with the License.
+ */
+
+export function getDomainFromUrl(url: string): string {
+  try {
+    const urlObject = new URL(url);
+    return urlObject.hostname;
+  } catch (e) {
+    console.error('Invalid URL:', e);
+    return '';
```
sergeyteleshev, line 15: I suggest to return undefined if error appeared so we definatelly know outside of the scope that something can go wrong here

alexander-skoblikov (approved):

PR #2813, opened 2024-07-30
CB-5355 change panels naming

devnaumov:


webapp/packages/plugin-sql-editor/src/locales/en.ts
```diff
@@ -10,7 +10,8 @@ export default [
   ['sql_editor_settings_insert_table_aliases_name', 'Insert table aliases (in FROM clause)'],
   ['sql_editor_settings_insert_table_aliases_desc', 'Automatically inserts table aliases'],
   ['sql_editor_menu', 'SQL Editor'],
-  ['sql_editor_script_editor', 'SQL Editor'],
+  ['sql_editor_script_editor', 'Editor'],
```
sergeyteleshev: I agree with @Wroud that just "SQL" is a perfect fit here. The "Editor" is just too abstract. Simply "SQL" speaking for itself, and once the user clicks, it would be obvious that there is SQL Editor lives under this tab

sergeyteleshev (dismissed):

LonwoLonwo:
Not actual, for now.

PR #2818, opened 2024-08-01
CB-5437 use flex component instead of container

devnaumov:


sergeyteleshev (approved):

webapp/packages/core-blocks/src/Flex/Flex.tsx
```diff
@@ -0,0 +1,80 @@
...
+      className={clsx(
+        classes.flex,
+        overflow && classes.overflow,
+        gap && gapClasses[gap],
+        wrap && wrapClasses[wrap],
+        direction && directionClasses[direction],
+        align && alignClasses[align],
+        justify && justifyClasses[justify],
+        className,
+      )}
```
Wroud: seems like it will be easier to use `s` helper function together with `data-s-*` attribute

Wroud (changes requested):

Wroud (approved):

PR #2821, opened 2024-08-02
CB-5414 Import table - wrong description for canceled import

sergeyteleshev:


webapp/packages/plugin-data-import/src/DataImportService.ts
```diff
@@ -65,6 +71,11 @@ export class DataImportService {
         processorId,
         file,
         event => {
+          if (isCancelled) {
+            controller.setMessage('plugin_data_import_process_fail_cancel_message');
+            return;
+          }
```
Wroud: probable there will be enough to check `event` object or to just `return` without setting message cuz message will be set by error of cancellation returned by backend. There also can be a case when backend will return success after attempt of cancellation.

I suggest to check if it's possible to determine aborted upload via `event` object. Otherwise do not apply anything from event callback after getting a result (can be fixed in `CustomGraphQLClient.overrideFilesUpload`)

Wroud (changes requested):

webapp/packages/plugin-data-import/src/DataImportService.ts
```diff
@@ -41,9 +41,15 @@ export class DataImportService {
   async importData(connectionId: string, contextId: string, projectId: string, resultsId: string, processorId: string, file: File) {
     const abortController = new AbortController();
     let cancelImplementation: (() => void | Promise<void>) | null;
+    let isCancelled = false;
 
     function cancel() {
-      cancelImplementation?.();
+      if (!cancelImplementation) {
+        return;
```
sergeyteleshev, line 48: no need to do something here cause it is handled here. message about cancelation and other statuses provided by server added automatically by `controller.reject` and `ProcessNotificationController`
Снимок экрана 2024-08-05 в 11 38 46

Wroud (approved):

PR #2829, opened 2024-08-06
CB-4990 feat: add authentication timeout

Wroud:


webapp/packages/core-authentication/src/UserInfoResource.ts
```diff
@@ -140,6 +140,8 @@ export class UserInfoResource extends CachedDataResource<UserInfo | null, void,
             return authInfo as AuthInfo;
           },
           1000,
+          undefined,
+          300_000, // 5 minutes
```
devnaumov: we can use this approach to avoid comments

5 * 60 * 1000

webapp/packages/core-executor/src/whileTask.ts
```diff
@@ -6,35 +6,54 @@
...
-  function stop() {
+  const lockPromise = new Promise<T>((_resolve, _reject) => {
+    resolve = _resolve;
+    reject = _reject;
+  }).finally(() => {
+    if (intervalTimeoutId) {
+      clearTimeout(intervalTimeoutId);
+    }
+
     if (timeoutId) {
```
sergeyteleshev: what if we want to use this with `timeout = 0`?
I suppose it is better to use here:
```suggestion
    if (isNotNullDefined(timeoutId)) {
```

webapp/packages/core-executor/src/whileTask.ts
```diff
@@ -6,35 +6,54 @@
...
   let timeoutId: NodeJS.Timeout | null;
   let activeTask: Promise<T> | null;
   let stopped = false;
 
-  function stop() {
+  const lockPromise = new Promise<T>((_resolve, _reject) => {
+    resolve = _resolve;
+    reject = _reject;
+  }).finally(() => {
+    if (intervalTimeoutId) {
```
sergeyteleshev: same here as below

webapp/packages/core-executor/src/whileTask.ts
sergeyteleshev: please check all `if` statements for this

webapp/packages/core-executor/src/whileTask.ts
```diff
@@ -56,7 +79,7 @@ export function whileTask<T>(
         if (state) {
           resolve(value);
         } else if (!stopped) {
-          timeoutId = setTimeout(runTask, interval);
+          intervalTimeoutId = setTimeout(runTask, interval);
```
sergeyteleshev, line 82: should it be `setInterval` instead? if we have to check periodically `getAuthStatus` for current task it should be `setInterval` or am I missing something here?

Wroud: this may lead to concurrency:
10s interval:
A -> started not ended
10s interval:
B -> started not ended

current approach is like:
10s interval:
A -> started not ended
A -> ended (after 20s for example)
10s interval:
B -> started

webapp/packages/core-authentication/src/UserInfoResource.ts
Wroud: fixed

webapp/packages/core-executor/src/whileTask.ts
Wroud: fixed

webapp/packages/core-executor/src/whileTask.ts
Wroud: fixed

sergeyteleshev (approved):

PR #2830, opened 2024-08-06
CB-5393 fix: modal submenu

Wroud:


webapp/packages/core-ui/src/ContextMenu/MenuBar/MenuBar.tsx
```diff
@@ -220,7 +220,7 @@ const SubMenuItem = observer<ISubMenuItemProps>(function SubMenuItem({ children,
   const panelAvailable = subMenuData.itemCreators.length > 0;
 
   return (
-    <ContextMenu menu={subMenuData} className={className} rtl={rtl} disclosure {...nestedMenuSettings}>
+    <ContextMenu menu={subMenuData} className={className} rtl={rtl} modal disclosure {...nestedMenuSettings}>
```
sergeyteleshev, line 223: it affects a lot of other menus. not only auto-refresh one. does `nestedMenuSettings.modal = true` works for this case? may be better like this if it does so no side effects

Wroud: I think that we can make it the default behavior for simplicity

sergeyteleshev (approved):

PR #2831, opened 2024-08-06
CB-5123. Added backend part, added new parameter licenseStatus to ser…

DenisSinelnikov:
…ver config gql


webapp/packages/core-authentication/src/PasswordPolicyService.ts
```diff
@@ -25,22 +25,25 @@ type ValidationResult = { isValid: true; errorMessage: null } | { isValid: false
...
   ) {
     makeObservable(this, {
       config: computed,
     });
   }
 
+  /**
+   * PasswordPolicyResource should be loaded before calling this method
+   */
   validatePassword(password: string): ValidationResult {
```
sergeyteleshev, line 47:
```suggestion
    /**
   * PasswordPolicyResource should be loaded before calling this method
   */
  validatePassword(password: string): ValidationResult {
    if (!this.passwordPolicyResource.isLoaded()) {
      throw new Error('Password policy should be loaded before calling this method');
    }
```

sergeyteleshev, line 47: I believe we should add a check here. There is a chance a person would not have read the comment and used this function without the resource had been loaded before

```suggestion
  validatePassword(password: string): ValidationResult {
    if (!this.passwordPolicyResource.isLoaded()) {
      throw new Error('Password policy is not loaded yet');
    }
```

webapp/packages/core-events/src/NotificationService.ts
```diff
@@ -80,26 +80,41 @@ export class NotificationService {
...
+        title: observable.ref,
+        message: observable.ref,
+        details: observable.ref,
+        persistent: observable.ref,
+        isSilent: observable.ref,
+        customComponent: observable.ref,
+        extraProps: observable.ref,
+        autoClose: observable.ref,
+        type: observable.ref,
+        timestamp: observable.ref,
```
sergeyteleshev, line 114: should `close` to be included here? seems like potential bug if mobx does not know that some notification was closed

webapp/packages/core-root/src/DefaultNavigatorSettingsResource.ts
```diff
@@ -0,0 +1,107 @@
...
+    } else {
+      this.update = {
+        hideFolders: false,
+        hideSchemas: false,
+        hideVirtualModel: false,
+        mergeEntities: false,
+        showOnlyEntities: false,
+        showSystemObjects: false,
+        showUtilityObjects: false,
+      };
```
sergeyteleshev: to the `INITIAL_UPDATE_STATE` const please

webapp/packages/core-root/src/PermissionsService.ts
```diff
@@ -6,14 +6,41 @@
...
-  constructor(private readonly permissions: SessionPermissionsResource) {}
+  get publicDisabled(): boolean {
+    if (
+      this.serverConfigResource.configurationMode ||
+      (this.serverLicenseStatusResource?.licenseRequired && !this.serverLicenseStatusResource.licenseValid)
+    ) {
+      return true;
+    }
+
+    return false;
```
sergeyteleshev: Can be simplified

```suggestion
    return this.serverConfigResource.configurationMode ||
      (this.serverLicenseStatusResource?.licenseRequired && !this.serverLicenseStatusResource.licenseValid)
```

webapp/packages/core-root/src/QuotasService.ts
```diff
@@ -46,10 +46,13 @@ export class QuotasService {
     };
   }
 
-  constructor(private readonly serverConfigResource: ServerConfigResource) {}
+  constructor(private readonly serverResourceQuotasResource: ServerResourceQuotasResource) {}
 
+  /**
+   * Quotas should be manually loaded from ServerResourceQuotasResource before using this method
+   */
```
sergeyteleshev, line 53: same as for `PasswordPolicyResource` comment a bit above

webapp/packages/core-root/src/QuotasService.ts
```diff
@@ -46,10 +46,13 @@ export class QuotasService {
...
   getQuota(key: QuotaKey) {
-    const serverQuota = this.serverConfigResource.data?.resourceQuotas[key];
+    const serverQuota = this.serverResourceQuotasResource.data?.resourceQuotas[key];
 
     if (isNumber(serverQuota)) {
       return serverQuota;
```
sergeyteleshev, line 58: What if `serverQuota` might be a stringified number one day?

webapp/packages/plugin-authentication-administration/src/Administration/ServerConfiguration/ServerConfigurationAuthenticationBootstrap.ts
```diff
@@ -85,6 +86,7 @@ export class ServerConfigurationAuthenticationBootstrap extends Bootstrap {
       return validation.invalidate();
     }
 
+    await this.passwordPolicyResource.load();
```
sergeyteleshev, line 89: Would it be a better place for this in the `load(): void {}` method of this class?

webapp/packages/plugin-data-viewer/src/DatabaseDataModel/Actions/ResultSet/ResultSetDataContentAction.ts
```diff
@@ -34,17 +34,31 @@ interface ICacheEntry {
...
     private readonly format: ResultSetFormatAction,
     private readonly graphQLService: GraphQLService,
+    private readonly serverResourceQuotasResource: ServerResourceQuotasResource,
     private readonly quotasService: QuotasService,
     private readonly cache: ResultSetCacheAction,
   ) {
     super(source);
 
+    function loadQuotas() {
+      setTimeout(() => serverResourceQuotasResource.load(), 0);
```
sergeyteleshev, line 51: why do we need to run it as soon as possible?

you can also use by the way
`queueMicrotask(() => serverResourceQuotasResource.load());`

webapp/packages/plugin-data-viewer/src/DatabaseDataModel/Actions/ResultSet/ResultSetDataContentAction.ts
```diff
@@ -34,17 +34,31 @@ interface ICacheEntry {
...
+    }
+
+    this.serverResourceQuotasResource.onDataOutdated.addHandler(loadQuotas);
+
+    loadQuotas();
+
+    this.subscriptionDispose = () => {
+      this.serverResourceQuotasResource.onDataOutdated.removeHandler(loadQuotas);
+    };
```
sergeyteleshev, line 60: why do we add `removeHandler` only after `dispose` executed?

webapp/packages/core-root/src/ServerConfigResource.ts
```diff
@@ -240,31 +169,18 @@ export class ServerConfigResource extends CachedDataResource<ServerConfig | null
...
-          await this.graphQLService.sdk.setDefaultNavigatorSettings({ settings: this.navigatorSettingsUpdate });
-
-          if (this.data) {
-            this.data.defaultNavigatorSettings = { ...this.navigatorSettingsUpdate };
-          } else {
-            this.setData(await this.loader());
-          }
-        }
-
-        if (this.isChanged() && !skipConfigUpdate) {
```
sergeyteleshev, line 258: I suppose we can keep `this.isChanged()` here

webapp/packages/core-root/src/DefaultNavigatorSettingsResource.ts
```diff
@@ -0,0 +1,107 @@
...
+  async save(): Promise<void> {
+    await this.performUpdate(
+      undefined,
+      undefined,
+      async () => {
+        await this.graphQLService.sdk.setDefaultNavigatorSettings({ settings: this.update });
+
+        this.setData(await this.loader());
+
+        this.onDataOutdated.execute();
```
sergeyteleshev, line 74: I think you missed this check here
https://github.com/dbeaver/cloudbeaver/blob/afb4ea205c36f49485f236ea0196d0c09a677895/webapp/packages/core-root/src/ServerConfigResource.ts#L248

```suggestion
        if (!isChanged()) {
            return;
        }

        await this.graphQLService.sdk.setDefaultNavigatorSettings({ settings: this.update });

        this.setData(await this.loader());

        this.onDataOutdated.execute();
```

sergeyteleshev (approved):

webapp/packages/core-authentication/src/PasswordPolicyService.ts
Wroud: it can be outdated but loaded, so we need to know that we need to manage it at place

devnaumov (approved):

webapp/packages/core-root/src/DefaultNavigatorSettingsResource.ts
Wroud: No, I'm not cuz performUpdate will call this check again before running the callback

webapp/packages/core-root/src/ServerConfigResource.ts
Wroud: we don't need it (it's already in performUpdate arg)

webapp/packages/plugin-data-viewer/src/DatabaseDataModel/Actions/ResultSet/ResultSetDataContentAction.ts
Wroud: what do you mean by that?

webapp/packages/plugin-data-viewer/src/DatabaseDataModel/Actions/ResultSet/ResultSetDataContentAction.ts
Wroud: We don't need to run it as soon as possible, but we want to prevent onDataOutdated executor from blocking

webapp/packages/plugin-authentication-administration/src/Administration/ServerConfiguration/ServerConfigurationAuthenticationBootstrap.ts
Wroud: No, we need to keep in mind that resources can be invalidated at any time and `load` executed only once on app start, we also don't need this resource to be loaded in app starting process because we only need this data in certain forms

webapp/packages/core-events/src/NotificationService.ts
Wroud: We probably don't want users to change this field

webapp/packages/core-root/src/QuotasService.ts
Wroud: Actually, it can be anything even an object, we may consider it in the future

PR #2853, opened 2024-08-19
CB-5503 Login form - incorrect password is cleared from the form after pressing Enter

sergeyteleshev:


Wroud:
i suggest to revert `this.ref?.reset();` calls from `useForm` hook and check other ways to solve problem with validation highlighting
form's 'reset' have different purpose and we can't use it to cover only validation case

Wroud (approved):

PR #2854, opened 2024-08-20
CB-5048 ignore undo and redo on codemirror level

devnaumov:


webapp/packages/plugin-codemirror6/src/ReactCodemirror.tsx
```diff
@@ -126,6 +126,12 @@ export const ReactCodemirror = observer<IReactCodeMirrorProps, IEditorRef>(
 
         editorView.dom.addEventListener('keydown', event => {
           const newEvent = new KeyboardEvent('keydown', event);
+
+          // we handle undo/redo on the CaptureView level
+          if ((newEvent.metaKey || newEvent.ctrlKey) && ['z', 'y'].includes(newEvent.key.toLowerCase())) {
+            return;
+          }
+
```
Wroud, line 134: this is for purpose, cuz we can have different editors (for example VQB)
we have our own changes history implementation, please check `SqlDataSourceHistory`

Probably we have a collision with our own implementation and codemirror (we need to disable it for codemirror and use only ours)

Wroud (changes requested):

Wroud (approved):

webapp/packages/plugin-sql-editor/src/SqlDataSource/SqlDataSourceHistory/SqlDataSourceHistory.ts
```diff
@@ -13,12 +13,12 @@ import { createSqlDataSourceHistoryInitialState } from './createSqlDataSourceHis
 import type { ISqlDataSourceHistory } from './ISqlDataSourceHistory';
 import type { ISqlDataSourceHistoryState } from './ISqlDataSourceHistoryState';
 
-const HISTORY_DELAY = 1000;
+const HOT_HISTORY_SIZE = 30;
```
sergeyteleshev, line 16: i dont feel like it is enough to store 30 last operations. maybe like 200-300?

sergeyteleshev (approved):

webapp/packages/plugin-sql-editor/src/SqlDataSource/SqlDataSourceHistory/SqlDataSourceHistory.ts
Wroud: The problem here is that we store it on the user's disc, and if the script is 1mb+, then each edit will take 1mb (we don't support diffs and copy all values), so for 200-300, it will be 200-300mb and 2gb-3gb for scripts over 10MB

Wroud: Another moment that we don't lose all history after 30 records, we will start to compress it and leave only one edit in each 5 seconds interval

PR #2863, opened 2024-08-21
Fix/cb-5383/sync-grid-focus

Wroud:


webapp/packages/plugin-data-spreadsheet-new/src/DataGrid/DataGridTable.tsx
```diff
@@ -164,18 +166,49 @@ export const DataGridTable = observer<IDataPresentationProps>(function DataGridT
...
+          }
+          return;
+        }
+
+        const idx = tableData.getColumnIndexFromColumnKey(key.column!);
+        const rowIdx = tableData.getRowIndexFromKey(key.row!);
+
+        focusSyncRef.current = { idx, rowIdx };
+
+        this.selectCell({ idx, rowIdx });
```
sergeyteleshev:
```suggestion
        if (!key?.column || !key?.row) {
          if (initial) {
            const selectedElements = selectionAction.getSelectedElements();

            if (selectedElements.length > 0) {
              key = selectedElements[0];
            } else {
              key = { column: viewAction.columnKeys[0], row: viewAction.rowKeys[0] };
            }

            focusSyncRef.current = { idx: 0, rowIdx: -1 };
            this.selectCell(focusSyncRef.current);
          } else {
            focusSyncRef.current = null;
          }
        }

        if (!key?.column || !key?.row) {
          return; // TODO throw an error that cannot focus for some reason?
        }

        const idx = tableData.getColumnIndexFromColumnKey(key.column);
        const rowIdx = tableData.getRowIndexFromKey(key.row);

        focusSyncRef.current = { idx, rowIdx };

        this.selectCell({ idx, rowIdx });
```

sergeyteleshev: I suggest to simplify a bit if/else logic + let's forbid to use `!` operator in most cases. because this way we trick TS and make potential points for the bugs to appear. It is better to handle all edge cases then use this operator

Wroud: actually in current implementation `!` can be omitted also

Wroud: and your implementation is incorrect

devnaumov (approved):

PR #2876, opened 2024-08-28
CB-5391 remove controller from the plugin-connections-administration

devnaumov:


webapp/packages/plugin-connections-administration/src/Administration/Connections/ConnectionsTable/Connection.tsx
```diff
@@ -18,13 +18,12 @@ import { ConnectionEdit } from './ConnectionEdit';
...
   connectionKey: IConnectionInfoParams;
   connection: DatabaseConnection;
-  projectName?: string | null;
+  projectName?: string;
+  icon?: string;
 }
 
-export const Connection = observer<Props>(function Connection({ connectionKey, connection, projectName }) {
-  const driversResource = useService(DBDriverResource);
+export const Connection = observer<Props>(function Connection({ connectionKey, connection, projectName, icon }) {
```
sergeyteleshev: please remove `DBDriverResource` import

webapp/packages/core-blocks/src/Table/TableState.ts
```diff
@@ -9,30 +9,30 @@ import { action, computed, makeObservable, observable } from 'mobx';
 
 import { Executor, IExecutor } from '@cloudbeaver/core-executor';
 
-type Key = string | string[];
+type Key = any;
```
sergeyteleshev: please lets do proper types if possible. it would make debug easier if something about to happen here

webapp/packages/plugin-connections-administration/src/Administration/Connections/ConnectionsTable/ConnectionsTable.tsx
```diff
@@ -44,16 +30,17 @@ export const ConnectionsTable = observer<Props>(function ConnectionsTable({ keys
         <TableColumnHeader min />
         <TableColumnHeader>{translate('connections_connection_name')}</TableColumnHeader>
         <TableColumnHeader>{translate('connections_connection_address')}</TableColumnHeader>
-        {displayProjects && <TableColumnHeader>{translate('connections_connection_project')}</TableColumnHeader>}
+        {state.shouldDisplayProjects && <TableColumnHeader>{translate('connections_connection_project')}</TableColumnHeader>}
```
Wroud: It is bad practice to move all logic to a single place instead of splitting it between places where you use it. `shouldDisplayProjects` is not used outside of this component, but you keep it in a `state` shared among different components. Now, look at `useConnectionsTable`. It's almost 200 rows of code.

It's always better to keep non-reused code close to its uses. This will reduce the amount of logic in single logical units and improve the readability of code.

Wroud: same situation with `getProjectName` and `getConnectionIcon`

Wroud (changes requested):

webapp/packages/plugin-connections-administration/src/Administration/Connections/ConnectionsAdministrationService.ts
```diff
@@ -75,7 +71,6 @@ export class ConnectionsAdministrationService extends Bootstrap {
       isHidden: () => this.serverConfigResource.distributed,
       getContentComponent: () => ConnectionsAdministration,
       getDrawerComponent: () => ConnectionsDrawerItem,
```
sergeyteleshev, line 73: why `onActivate` logic was removed?

devnaumov: I think its better to load resources in places where there are used. In this case if we delete the component or hook that uses certain resource, the api call also will be deleted. When we load something on onActivate, we cant be sure where certain resources are used

Wroud (approved):

sergeyteleshev (approved):

PR #2885, opened 2024-09-02
CB-5478 Add an ability to new forms to edit formState id

sergeyteleshev:


webapp/packages/plugin-authentication-administration/src/Administration/Users/UsersTable/useAdministrationUserFormState.ts
```diff
@@ -0,0 +1,45 @@
...
+
+    return formState;
+  }
+
+  useEffect(() => {
+    if (ref.formState.id !== id) {
+      const formState = createFormState(id);
+      ref.formState = formState;
+    }
+  }, [id]);
```
Wroud: Seems a bit complicated. Can it be simplified like this?:

```ts
const ref = useRef<null | FormStateTypeHere>(null);

if(ref.current?.id !== id) {
  ref.current = new ...
}

return ref.current
```

the main difference:
form initialized and re-initialized same way (instead of init in render and re-initialization in the effect + re-render)
logic is straight forward
no need for mobx observable (cuz it will always produce correct state for provided id argument)

Wroud (changes requested):

webapp/packages/plugin-authentication-administration/src/Administration/Users/UsersTable/useAdministrationUserFormState.ts
sergeyteleshev: yeah it can be simplified that way. thanks

Wroud (approved):

PR #2906, opened 2024-09-10
CB-5511 fix: pagination

Wroud:


webapp/packages/core-authentication/src/UsersResource.ts
```diff
@@ -240,6 +242,7 @@ export class UsersResource extends CachedMapResource<string, AdminUser, UserReso
     }
 
     const usersList: AdminUser[] = [];
+    const pages: Parameters<typeof this.offsetPagination.setPage>[] = [];
```
sergeyteleshev, line 245: lets create a separate type for that

webapp/packages/core-resource/src/Resource/OffsetPagination/ResourceOffsetPage.ts
```diff
@@ -0,0 +1,83 @@
...
+
+  constructor() {
+    this.from = 0;
+    this.to = 0;
+    this.items = [];
+    this.outdated = false;
+
+    makeObservable(this, {
+      from: true,
+      to: true,
```
sergeyteleshev, line 26: why true?
may be `observable` also?

webapp/packages/core-resource/src/Resource/CachedResource.ts
```diff
@@ -316,17 +350,17 @@ export abstract class CachedResource<
...
-          ...metadata.offsetPage,
-          pages: expandOffsetPageRange(metadata.offsetPage?.pages || [], pageKey.options, false),
-        });
-      }
+      // if (pageKey) {
+      //   metadata.offsetPage = observable({
+      //     ...metadata.offsetPage,
+      //     pages: expandOffsetPageRange(metadata.offsetPage?.pages || [], pageKey.options, false),
+      //   });
+      // }
```
sergeyteleshev: what about this logic? should it be deleted at all or handled in a different way according to new api?
now I can see that pages wont be marked as loaded. is it correct?

webapp/packages/core-resource/src/Resource/CachedResourceOffsetPageKeys.ts
```diff
@@ -114,39 +114,48 @@ export function limitOffsetPages(pages: IResourceOffsetPage[], limit: number): I
   return result;
 }
```
sergeyteleshev: really hard to read what is going on here. please at least TODO to make it more readable

webapp/packages/core-resource/src/Resource/ResourceOffsetPagination.ts
```diff
@@ -47,29 +50,36 @@ export class ResourceOffsetPagination<TKey, TMetadata extends ICachedResourceMet
     return pageInfo.end === undefined || to < pageInfo.end;
   }
 
-  setPageEnd(key: ResourceAlias<TKey, Readonly<ICachedResourceOffsetPageOptions>>, hasNextPage: boolean): void {
-    const count = key.options.offset + key.options.limit;
+  setPage(key: ResourceAlias<TKey, Readonly<ICachedResourceOffsetPageOptions>>, items: any[], hasNextPage: boolean) {
+    items = [...items];
```
sergeyteleshev: can be removed

sergeyteleshev (approved):

webapp/packages/core-resource/src/Resource/OffsetPagination/ResourceOffsetPage.ts
devnaumov: Yeah, seems strange

webapp/packages/core-resource/src/Resource/OffsetPagination/ResourceOffsetPage.ts
```diff
@@ -0,0 +1,83 @@
...
+      items: observable.shallow,
+      outdated: observable,
```
devnaumov, line 28: observable.ref maybe? as its a boolean

devnaumov (approved):

webapp/packages/core-resource/src/Resource/OffsetPagination/ResourceOffsetPage.ts
Wroud: it's shorthand for from: observable

webapp/packages/core-resource/src/Resource/OffsetPagination/ResourceOffsetPage.ts
Wroud: there is no difference for primitives actually `observable` and `observable.ref` will have same effect (and true)

webapp/packages/core-authentication/src/UsersResource.ts
Wroud: im not sure that we need it, because we don't and can't use it anywhere to declare exact implementation, instead we have several places when we need to "delay" `setPage` call (we want to call it synchronously in same action with data mutation and after data mutation)

Probably we can want in future ability to delay setPage call via API but not with this array of arguments + for each + call

webapp/packages/core-resource/src/Resource/ResourceOffsetPagination.ts
Wroud: thanks, fixed

webapp/packages/core-resource/src/Resource/CachedResourceOffsetPageKeys.ts
Wroud: thanks, removed (it's not in use anymore)

webapp/packages/core-resource/src/Resource/CachedResource.ts
Wroud: we will delete it, pages have different logic

PR #2908, opened 2024-09-11
CB-4541 feat: tree filter

Wroud:


webapp/packages/plugin-navigation-tree/src/TreeNew/useTreeFilter.ts
```diff
@@ -0,0 +1,57 @@
...
+  options = useObjectRef(options);
+  return useObservableRef<ITreeFilter>(
+    () => ({
+      filter: '',
+      isNodeMatched(treeData: ITreeData, nodeId: string): boolean {
+        if (!this.filter) {
+          return true;
+        }
+
+        let isNodeMatched = treeData.getNode(nodeId).name.toLowerCase().includes(this.filter.toLowerCase());
```
sergeyteleshev: An idea for the future: we can implement a fuzzy search for all of our filters in the app. it is easy peasy to implement and improves UX experience while searching something

webapp/packages/plugin-navigation-tree/src/TreeNew/useTreeFilter.ts
```diff
@@ -0,0 +1,57 @@
...
+      },
+      transformer(treeData: ITreeData, nodeId: string, children: string[]): string[] {
+        return children.filter(child => this.isNodeMatched(treeData, child));
+      },
+      setFilter(filter: string): void {
+        this.filter = filter;
+      },
+    }),
+    {
+      filter: observable.ref,
```
sergeyteleshev, line 53: lets list actions here also

webapp/packages/plugin-navigation-tree/src/TreeNew/DataTransformers/rootTransformers.ts
```diff
@@ -10,7 +10,7 @@ import type { INodeState } from '../INodeState';
 import type { TreeDataTransformer } from './TreeDataTransformer';
 
 export function rootNodeStateTransformer(root: string): TreeDataTransformer<INodeState> {
-  return function rootNodeStateTransformer(nodeId, data) {
+  return function rootNodeStateTransformer(treeData, nodeId, data) {
```
sergeyteleshev, line 13: not in use here and below also. is this for future use or just forgot to delete it?

webapp/packages/plugin-settings-panel/src/SettingsPanel/SettingsList.tsx
```diff
@@ -36,17 +38,17 @@ export const SettingsList = observer<Props>(function SettingsList({ treeData, so
     groups.splice(0, 1, ...treeData.getChildren(groupId));
 
     const group = ROOT_SETTINGS_GROUP.get(groupId)!;
-    const groupSettings = settings.get(group);
 
-    list.push({ group, settings: groupSettings || [] });
+    list.push(group);
```
sergeyteleshev: Suggest to move this code to function getGroupList() and do it using for cylcle so there is less chance to get an infinite loop

I'm not a big fan of while cycles

webapp/packages/plugin-settings-panel/src/SettingsPanel/SettingsGroup.tsx
```diff
@@ -7,26 +7,33 @@
...
   source: ISettingsSource;
-  settings: ISettingDescription<any>[];
+  settings: Map<SettingsGroupType, ISettingDescription<any>[]>;
+  treeFilter: ITreeFilter;
 }
 
-export const SettingsGroup = observer<Props>(function SettingsGroup({ group, source, settings }) {
+export const SettingsGroup = observer<Props>(function SettingsGroup({ group, source, settings, treeFilter }) {
+  const translate = useTranslate();
+  const groupSettings = getComputed(() => settings.get(group)?.filter(settingsFilter(translate, treeFilter.filter)) || [], isArraysEqual);
```
sergeyteleshev, line 29: there is no need for current specific `SettingsGroup` to know something about all other `settings` map. I suggest to calculate this in parent component (in mapping function helper maybe idk) and pass here as a prop. so we wont mess up the logic here in future

webapp/packages/plugin-settings-panel/src/SettingsPanel/SettingsList.tsx
```diff
@@ -36,17 +38,17 @@ export const SettingsList = observer<Props>(function SettingsList({ treeData, so
...
-    <Container ref={ref} style={{ height: '100%' }} gap overflow>
-      {list.map(({ group, settings }) => (
-        <SettingsGroup key={group.id} group={group} source={source} settings={settings} />
+    <Container ref={ref} gap overflow>
+      {list.map(group => (
+        <SettingsGroup key={group.id} group={group} source={source} settings={settings} treeFilter={treeFilter} />
       ))}
-      <div style={{ height: 'calc(100% - 100px)' }} hidden />
+      {list.length === 0 && <TextPlaceholder>{translate('plugin_settings_panel_no_settings')}</TextPlaceholder>}
+      <div style={{ height: '25%' }} />
```
sergeyteleshev: to css file

webapp/packages/tests-runner/package.json
```diff
@@ -14,6 +14,8 @@
   },
   "dependencies": {
     "@cloudbeaver/core-di": "^0",
+    "@jest/globals": "^29",
+    "@testing-library/jest-dom": "^6",
     "@testing-library/react": "^16",
```
sergeyteleshev, line 20: isnt it all `devDependencies`?
including `"@testing-library/react": "^16",`

webapp/packages/plugin-navigation-tree/src/TreeNew/useTreeFilter.ts
Wroud: it's not an actions, it's map functions and it's important for correct mobx subscriptions (the only action is setFilter but it's mutates single field so there is no difference)

webapp/packages/plugin-navigation-tree/src/TreeNew/DataTransformers/rootTransformers.ts
Wroud: i personally prefer to left full name than _

webapp/packages/plugin-settings-panel/src/SettingsPanel/SettingsGroup.tsx
Wroud: this made for optimization, before this settings was created together with group in parent component, that lead to full re-render because of new props (always new array of settings)

current implementation shifts settings getter logic to component that renders it

webapp/packages/tests-runner/package.json
Wroud: for tests-runner it's not because it produces functions that actually depends on this, but other packages uses tests-runner as a devDependency

sergeyteleshev (approved):

webapp/packages/plugin-settings-panel/src/SettingsPanel/getGroupsFromTree.ts
```diff
@@ -0,0 +1,21 @@
...
+ * CloudBeaver - Cloud Database Manager
+ * Copyright (C) 2020-2024 DBeaver Corp and others
+ *
+ * Licensed under the Apache License, Version 2.0.
+ * you may not use this file except in compliance with the License.
+ */
+import { ROOT_SETTINGS_GROUP, type SettingsGroup as SettingsGroupType } from '@cloudbeaver/core-settings';
+import type { ITreeData } from '@cloudbeaver/plugin-navigation-tree';
+
+export function* getGroupsFromTree(treeData: ITreeData, groups: string[]): IterableIterator<SettingsGroupType> {
```
devnaumov, line 11: What is the reason for the generator here?

Wroud: it's fix for this comment https://github.com/dbeaver/cloudbeaver/pull/2908#discussion_r1756768170

PR #2922, opened 2024-09-17
CB-5658 CB-5659 fix: pagination resolving

Wroud:


webapp/packages/core-resource/src/Resource/ResourceAlias.ts
```diff
@@ -42,6 +42,10 @@ export abstract class ResourceAlias<TKey, TOptions extends ResourceAliasOptions>
     return undefined;
   }
 
+  copy(): this {
```
devnaumov: I think we should add some comments here to explain why we need this method

webapp/packages/core-navigation-tree/src/NodesManager/NavTreeResource.ts
```diff
@@ -461,8 +461,8 @@ export class NavTreeResource extends CachedMapResource<string, string[], Record<
   }
 
   protected async loader(originalKey: ResourceKey<string>): Promise<Map<string, string[]>> {
-    const pageKey =
-      this.aliases.isAlias(originalKey, CachedResourceOffsetPageKey) || this.aliases.isAlias(originalKey, CachedResourceOffsetPageListKey);
+    const pageListKey = this.aliases.isAlias(originalKey, CachedResourceOffsetPageListKey);
+    const pageKey = this.aliases.isAlias(originalKey, CachedResourceOffsetPageKey) || pageListKey;
```
sergeyteleshev: I can see a copy-paste here and a little bit below for all resources with pagination.
Maybe we should add a helper here so it returns the correct `pageKey` and the `page`? So we don't miss this bug when we create a new feature with the exact pagination mechanism

webapp/packages/core-resource/src/Resource/CachedResource.ts
```diff
@@ -91,41 +91,58 @@ export abstract class CachedResource<
...
 
-        for (const page of pageInfo.pages) {
-          if (page.isHasCommonSegment(from, to)) {
-            keys.push(...page.get(from, to));
+          for (const page of pageInfo.pages) {
+            if (page.isHasCommonSegment(from, to)) {
+              for (const pageKey of page.get(from, to)) {
+                keys.add(pageKey);
+              }
+            }
```
sergeyteleshev, line 140: Same here. we can define helper like `getKeysFromPages(pages)`

sergeyteleshev (approved):

webapp/packages/core-resource/src/Resource/ResourceAlias.ts
Wroud: removed

devnaumov (approved):

webapp/packages/core-resource/src/Resource/CachedResource.ts
Wroud: let's keep it for now

webapp/packages/core-navigation-tree/src/NodesManager/NavTreeResource.ts
Wroud: now you can use `getOffsetPageKeyInfo` helper

sergeyteleshev (approved):

PR #2927, opened 2024-09-19
CB-5625 Show only enabled identity providers in configuration section

sergeyteleshev:


webapp/packages/core-authentication/src/AuthProvidersResource.ts
```diff
@@ -20,19 +20,19 @@ import { ServerConfigResource } from '@cloudbeaver/core-root';
...
-import { AuthSettingsService } from './AuthSettingsService';
 
 export type AuthProvider = NonNullable<AuthProviderInfoFragment>;
 export type AuthProviderConfiguration = NonNullable<AuthProviderConfigurationInfoFragment>;
 
 @injectable()
 export class AuthProvidersResource extends CachedMapResource<string, AuthProvider> {
   get configurable(): AuthProvider[] {
-    return this.values.filter(provider => provider.configurable);
+    const enabledAuthProviders = new Set(this.serverConfigResource.enabledAuthProviders);
```
devnaumov: We dont want this logic in resource, please move it to the plugin-identity-providers-administration and filter on the component level, also we want to show some placeholder (we need to thing about the ui) if there is no available providers

devnaumov (dismissed):

webapp/packages/core-authentication/src/AuthProvidersResource.ts
```diff
@@ -50,6 +54,7 @@ export class AuthProvidersResource extends CachedMapResource<string, AuthProvide
 
     makeObservable(this, {
       configurable: computed,
+      enabledAuthProviders: computed,
```
Wroud: anytime this `computed` triggered it will lead to subscribers re-validation, because it returns new object every time
please consider to remove `computed` or pass `equals` function (make sure that this computed is heavy in computations to do so, because comparing can be overhead)

webapp/packages/core-ui/src/Form/FormBaseService.ts
```diff
@@ -36,11 +36,11 @@ export class FormBaseService<TState, TProps extends IFormProps<TState> = IFormPr
     this.parts = new TabsContainer(name);
     this.actionsContainer = new PlaceholderContainer();
     this.onConfigure = new ExecutorHandlersCollection();
-    this.onFillDefaultConfig = new ExecutorHandlersCollection();
```
Wroud: do we still use `onConfigure`?
i think it was part of `load` method of FormState

Wroud (changes requested):

webapp/packages/core-ui/src/Form/FormBaseService.ts
sergeyteleshev: we do not. thanks!

webapp/packages/core-ui/src/Form/FormBaseService.ts
```diff
@@ -20,12 +20,11 @@ export class FormBaseService<TState, TProps extends IFormProps<TState> = IFormPr
   readonly parts: TabsContainer<TProps>;
   readonly actionsContainer: PlaceholderContainer<TProps>;
 
-  readonly onConfigure: IExecutorHandlersCollection<IFormState<TState>>;
-  readonly onFillDefaultConfig: IExecutorHandlersCollection<IFormState<TState>>;
   readonly onPrepareConfig: IExecutorHandlersCollection<TState>;
```
Wroud: seems like we don't use `onPrepareConfig` also

webapp/packages/core-ui/src/Form/IFormState.ts
```diff
@@ -30,8 +30,8 @@ export interface IFormState<TState> {
   readonly statusMessage: string | string[] | null;
   readonly statusType: ENotificationType | null;
 
+  readonly loadedStateTask: IExecutor<IFormState<TState>>;
```
Wroud: please rename to `loadedTask`

webapp/packages/core-ui/src/Form/FormBaseService.ts
Wroud: please also check other executors

sergeyteleshev: This was the last one

Wroud (approved):

Wroud (changes requested):
please revert all related to `setOnShowDetails` please use components for such logic

Wroud (approved):

PR #2933, opened 2024-09-24
dbeaver/dbeaver-vscode#16 fix SQL processor, server port cli arg

alexander-skoblikov:


server/bundles/io.cloudbeaver.model/src/io/cloudbeaver/model/session/WebSession.java
```diff
@@ -129,16 +129,11 @@ public WebSession(
         @NotNull WebAuthApplication application,
         @NotNull Map<String, DBWSessionHandler> sessionHandlers
     ) throws DBException {
-        super(requestInfo.getId(), application);
-        this.lastAccessTime = this.createTime;
```
serge-rider, line 133: Why remove all the init sequence?

alexander-skoblikov: It is a duplicated code replaced by a call this(..) constructor

PR #2935, opened 2024-09-26
CB-5446 chore: keyboard navigation improvements

Wroud:


webapp/packages/core-blocks/src/useListKeyboardNavigation.ts
```diff
@@ -0,0 +1,100 @@
...
+
+    // Function to reset tabindex on all elements and set it to 0 on aria-selected="true"
+    const resetTabindex = () => {
+      const focusableElements = getFocusableElements();
+      focusableElements.forEach(el => {
+        if (el.getAttribute('aria-selected') === 'true') {
+          el.setAttribute('tabindex', '0');
+        } else {
+          el.setAttribute('tabindex', '-1');
+        }
```
sergeyteleshev, line 35: I suppose this set of elements can include elements without `aria-selected` attribute. Is this okay to set `tabIndex` for them also to `-1`? Or it is supposed to be only for elements with `aria-selected` attribute?

webapp/packages/core-blocks/src/useListKeyboardNavigation.ts
```diff
@@ -0,0 +1,100 @@
...
+
+      let newIndex = currentIndex;
+
+      EventContext.set(e, EventKeyboardNavigationFlag);
+      e.preventDefault();
+
+      switch (e.key) {
+        case 'ArrowRight':
+        case 'ArrowDown':
+          newIndex = (currentIndex + 1) % focusableElements.length; // Move to next element
```
sergeyteleshev, line 59: lets move to `next/prev` element into functions

webapp/packages/plugin-data-viewer/src/TableViewer/TableViewer.module.css
```diff
@@ -51,6 +51,16 @@
...
 }
+.paneContent.grid:focus-visible::after {
+  content: '';
+  position: absolute;
+  top: 0;
+  left: 0;
+  right: 0;
+  bottom: 0;
+  outline-offset: -1px;
+  outline: -webkit-focus-ring-color auto 1px;
```
sergeyteleshev: this one is not working for Firefox

sources:
https://stackoverflow.com/questions/7538771/what-is-webkit-focus-ring-color
https://css-tricks.com/copy-the-browsers-native-focus-styles/

webapp/packages/plugin-data-viewer/src/TableViewer/TableViewer.tsx
```diff
@@ -69,6 +71,10 @@ export const TableViewer = observer<TableViewerProps, HTMLDivElement>(
     const loading = dataModel?.isLoading() ?? true;
     const dataFormat = result?.dataFormat || ResultDataFormat.Resultset;
     const splitState = useSplitUserState('table-viewer');
+    const navRef = useListKeyboardNavigation(
+      '[data-presentation][tabindex]:not(:disabled), [data-presentation-tools][tabindex]:not(:disabled), [data-presentation-header][tabindex]:not(:disabled), [data-presentation-bar][tabindex]:not(:disabled)',
```
sergeyteleshev, line 75: just an idea (not required to implement right now): we should implement some kind of selector builder with chain of responsibility pattern cause now it is easy to miss-type something and get a bug

sergeyteleshev (approved):

webapp/packages/core-blocks/src/useListKeyboardNavigation.ts
Wroud: this is okay, if element don't have this attribute then tabindex=-1 is okay, but here more problems over -1 and 0 because sometimes you can't restore initial value (when you don't have aria-selected or when element was tabindex=0 initially)

but for now seems that it works, maybe we will found bugs related to this function later

webapp/packages/plugin-data-viewer/src/TableViewer/TableViewer.module.css
Wroud: fixed

PR #2941, opened 2024-09-26
dbeaver/pro#1441 Universal Database Tunnel POC

ShadelessFox:


server/bundles/io.cloudbeaver.server/src/io/cloudbeaver/server/jetty/CBJettyServletContext.java
```diff
@@ -21,16 +21,27 @@
...
 import org.eclipse.jetty.ee10.servlet.ServletContextHandler;
 import org.eclipse.jetty.ee10.servlet.ServletHolder;
+import org.eclipse.jetty.server.Server;
 
 public class CBJettyServletContext implements DBWServletContext {
+    private final Server server;
     private final ServletContextHandler contextHandler;
 
-    public CBJettyServletContext(ServletContextHandler contextHandler) {
+    public CBJettyServletContext(Server server, ServletContextHandler contextHandler) {
```
serge-rider: I don't like the idea of adding server reference here.
If you really need it then you can keep it as variable in CBApplication.

And we definitely do not want to patch VSCode app which has nothing to do with web sockets at all.

serge-rider (changes requested):

PR #2958, opened 2024-10-02
CB-5693 update protobuf version

yagudin10:


server/drivers/mysql/pom.xml
```diff
@@ -20,6 +20,11 @@
             <artifactId>mysql-connector-j</artifactId>
             <version>8.2.0</version>
         </dependency>
+        <dependency>
+            <groupId>com.google.protobuf</groupId>
```
LonwoLonwo, line 24: Please, move version to dependency management in dbeaver-common

alexander-skoblikov (approved):

PR #2960, opened 2024-10-02
CB-5749 fix: metadata table loading

Wroud:


sergeyteleshev (approved):

devnaumov (approved):

webapp/packages/core-resource/src/Resource/CachedResourceOffsetPageKeys.ts
```diff
@@ -108,41 +108,64 @@ export function expandOffsetPageRange(
...
+      i++;
     }
+  } else {
+    // Since hasNextPage is false, we remove all pages after newPage
+    // No action needed here as we simply don't add them
   }
 
-  const lastPage = pages[pages.length - 1];
+  // Remove zero-length ranges
+  const filteredPages = mergedPages.filter(range => range.to - range.from > 0);
```
sergeyteleshev, line 164: do you think we can simplify the logic upper this line? cause this overlapping logic seems a bit unstable and easy to get a bug here someday

I mean can we do it in 3 cycles:
- Add all pages including `newPage` and do `current.setSize()` (add stage)
- Filter zero-length ranges (filter stage)
- Sort pages by `from` filter (sort stage)

sergeyteleshev: I mean that filtering is gonna handle this overlapping logic cause it is ordered

sergeyteleshev (approved):

PR #2972, opened 2024-10-08
CB-5710 return anonymous user info

yagudin10:


sergeyteleshev (approved):

webapp/packages/core-authentication/src/UserInfoResource.ts
```diff
@@ -55,6 +55,10 @@ export class UserInfoResource extends CachedDataResource<UserInfo | null, void,
     return this.data !== null;
   }
 
+  get isAnonymous() {
+    return !this.data || this.data.isAnonymous;
+  }
```
Wroud: If we don't have this.data we don't know and can't say that the user is anonymous

better to use `this.data && this.data.isAnonymous`

webapp/packages/core-authentication/src/UserInfoResource.ts
```diff
@@ -73,9 +77,14 @@ export class UserInfoResource extends CachedDataResource<UserInfo | null, void,
 
     makeObservable(this, {
       parametersAvailable: computed,
+      isAnonymous: computed,
     });
   }
 
+  isAuthenticated(): this is { data: UserInfo } {
+    return !this.isAnonymous;
+  }
```
Wroud: this check should `!!this.data`

webapp/packages/plugin-authentication/src/PluginBootstrap.ts
```diff
@@ -28,7 +28,7 @@ export class PluginBootstrap extends Bootstrap {
     this.menuService.addCreator({
       menus: [TOP_NAV_BAR_SETTINGS_MENU],
       getItems: (context, items) => {
-        if (this.serverConfigResource.enabledAuthProviders.length > 0 && !this.authInfoService.userInfo) {
+        if (this.serverConfigResource.enabledAuthProviders.length > 0 && this.userInfoResource.isAnonymous) {
```
Wroud: `!this.userInfoResource.isAnonymous && !this.userInfoResource.isAuthenticated`

if we use this check multiple times it's better to add third flag to `UserInfoResource` like `isRealUser` or something like this, so we will have three states:

anonymous - authenticated as anonymous user
real user - authenticated as real user with credentials
authenticated - anonymous || real user

Wroud (changes requested):

webapp/packages/plugin-authentication/src/PluginBootstrap.ts
Wroud: please check and update cases where `serverConfigResource.anonymousAccessEnabled` is used together with check for user, it can be completely replaced with this new flags

webapp/packages/core-authentication/src/UserInfoResource.ts
```diff
@@ -81,8 +89,8 @@ export class UserInfoResource extends CachedDataResource<UserInfo | null, void,
     });
   }
 
-  isAuthenticated(): this is { data: UserInfo } {
-    return !this.isAnonymous;
+  isData(): this is { data: UserInfo } {
+    return !!this.data;
   }
```
Wroud: kinda strange, what difference with `if (resource.data)` and `if (resource.isData())`?

webapp/packages/plugin-connections/src/PublicConnectionForm/PublicConnectionFormService.ts
```diff
@@ -60,7 +60,7 @@ export class PublicConnectionFormService {
       executorHandlerFilter(
         () => !!this.formState && this.optionsPanelService.isOpen(formGetter),
         async (event, context) => {
-          if (event === 'before' && this.userInfoResource.data === null) {
+          if (event === 'before' && !this.userInfoResource.isData()) {
```
Wroud: for this place it's better to use `this.userInfoResource.isAnonymous()` because we want to display this message only if user is anonymous and this action can lead to login as a user

webapp/packages/plugin-user-profile/src/UserMenu/UserMenu.tsx
```diff
@@ -27,7 +27,7 @@ export const UserMenu = observer(function UserMenu() {
     context.set(DATA_CONTEXT_USER, userInfoResource.data, id);
   });
 
-  if (!userInfoResource.isAuthenticated()) {
+  if (!userInfoResource.isData() || !userInfoResource.isAuthenticated) {
```
Wroud: seems like we can make `isAuthenticated` and `isAnonymous` type guards cuz they both guarantee that data is presented

webapp/packages/plugin-user-profile/src/UserProfileOptionsPanelService.ts
```diff
@@ -82,7 +82,7 @@ export class UserProfileOptionsPanelService {
       return;
     }
 
-    if (this.userInfoResource.data === null) {
+    if (!this.userInfoResource.isData()) {
```
Wroud: here is better to use `!this.userInfoResource.hasAccess` so we will close this panel if user now not anonymous and not real user

Wroud (changes requested):

Wroud (approved):

PR #2973, opened 2024-10-08
CB-5707 Change font size in dialogs 

sergeyteleshev:


webapp/packages/core-blocks/src/CommonDialog/CommonDialog/CommonDialogHeader.module.css
```diff
@@ -27,6 +27,7 @@
 }
 
 .headerTitle {
+  font-size: 20px;
```
Wroud: please use proper headline class from MDC: https://github.com/material-components/material-components-web/tree/v4.0.0/packages/mdc-typography

probably headline5 or headline6

Wroud (changes requested):

webapp/packages/core-blocks/src/CommonDialog/CommonDialog/CommonDialogHeader.module.css
Wroud: you can use under our namespace like `theme-typography--headline6`

sergeyteleshev: I was about to push it with your suggestion. But it adds an additional `line-height` style, which was not on the previous screenshots before the font had changed

Wroud (approved):

PR #2981, opened 2024-10-08
CB-5563 [TE] Change credentials - 'save credentials' is not marked for auth even if it is saved

sergeyteleshev:


Wroud (approved):

webapp/packages/plugin-connections/src/DatabaseAuthDialog/DatabaseCredentialsAuthDialog/useDatabaseCredentialsAuthDialog.ts
```diff
@@ -96,7 +96,7 @@ export function useDatabaseCredentialsAuthDialog(
           for (const id of this.networkHandlers) {
             const handler = connection.networkHandlersConfig.find(handler => handler.id === id);
 
-            if (handler && (handler.userName || handler.authType !== NetworkHandlerAuthType.Password)) {
```
devnaumov: What changed? Why don't we want this check anymore?

sergeyteleshev: Cause of this case:
https://dbeaver.atlassian.net/browse/CB-5563?focusedCommentId=23902

sergeyteleshev: cause it does not apply for SSL (no `username` and `NetworkHandlerAuthType.Password`)
see the last attached case to the ticket in the comments

sergeyteleshev: reverted this logic. we need to groom this form in general cause it has several issues with SSH and SSL fields (created tickets for that)

PR #2987, opened 2024-10-09
CB-5424 QM cleanup job

alexander-skoblikov:


server/bundles/io.cloudbeaver.model/src/io/cloudbeaver/model/app/WebServerConfiguration.java
```diff
@@ -27,4 +31,9 @@ default String getRootURI() {
         return "";
     }
 
+    @NotNull
```
serge-rider, line 38: Please add comment what is it for

serge-rider (changes requested):

serge-rider (approved):

PR #3017, opened 2024-10-21
CB-5645 [EE\AWS] Change the instruction in the Version update tab

sergeyteleshev:


webapp/packages/core-links/src/GithubLinks.ts
```diff
@@ -0,0 +1,21 @@
+/*
+ * CloudBeaver - Cloud Database Manager
+ * Copyright (C) 2020-2024 DBeaver Corp and others
+ *
+ * Licensed under the Apache License, Version 2.0.
+ * you may not use this file except in compliance with the License.
+ */
+export class GithubLinks {
```
devnaumov: Seems like its all static, we could convert it to const and function

webapp/packages/core-version-update/src/BaseUpdateInstruction.tsx
```diff
@@ -0,0 +1,27 @@
...
+ * Copyright (C) 2020-2024 DBeaver Corp and others
+ *
+ * Licensed under the Apache License, Version 2.0.
+ * you may not use this file except in compliance with the License.
+ */
+import { Link, useTranslate } from '@cloudbeaver/core-blocks';
+
+import type { InstructionComponent } from './VersionUpdateService.js';
+
+const SPLITTER = '*splitter*';
```
devnaumov: This splitter is confusing, why do we need it? Can we use \n or translate args?

webapp/packages/core-links/package.json
```diff
@@ -1,5 +1,5 @@
 {
-  "name": "@cloudbeaver/core-website",
```
devnaumov, line 2: Why do we need to rename the package? I think both has the similar meaning

devnaumov (changes requested):

webapp/packages/core-links/src/GithubLinks.ts
sergeyteleshev: done

webapp/packages/core-version-update/src/BaseUpdateInstruction.tsx
sergeyteleshev: split to 2 locale keys. the best we can do here

webapp/packages/core-links/package.json
sergeyteleshev: it was about our website dbeaver.com, but now I can see it can be any links we have: website, github, social media, marketplaces etc. so left it as it is right now

devnaumov (approved):

PR #3019, opened 2024-10-22
CB-5833 Fix e-mails recognition for SSO

sergeyteleshev:


webapp/packages/plugin-authentication-administration/src/Administration/Users/UserForm/Info/UserFormInfoPart.ts
```diff
@@ -96,6 +96,7 @@ export class UserFormInfoPart extends FormPart<IUserFormInfoState, IUserFormStat
         authRole: getTransformedAuthRole(this.state.authRole),
         enabled: this.state.enabled,
       });
+      this.state.userId = user.userId;
```
Wroud, line 108: please add comment that userId is modified by backend and may not match value we sent, so we need to update state

Wroud (approved):

PR #3023, opened 2024-10-23
CB-5738 api for tokens

yagudin10:


server/bundles/io.cloudbeaver.service.security/db/cb_schema_update_22.sql
```diff
@@ -0,0 +1,11 @@
+CREATE TABLE {table_prefix}CB_ACCESS_TOKEN
+(
+    TOKEN_ID        VARCHAR(128) NOT NULL,
+    USER_ID         VARCHAR(128) NOT NULL,
+    TOKEN_NAME      VARCHAR(128) NOT NULL,
+    PERIOD          INTEGER NOT NULL DEFAULT 0,
```
alexander-skoblikov:
```suggestion
    EXPIRATION_TIME               TIMESTAMP    NULL
```

alexander-skoblikov: or not null if needed

server/bundles/io.cloudbeaver.service.auth/schema/service.auth.graphqls
```diff
@@ -148,6 +148,13 @@ type UserTeamInfo {
     teamRole: String
 }
 
+type APITokenInfo @since(version: "24.2.4") {
```
alexander-skoblikov: move to a separate plugin

server/bundles/io.cloudbeaver.service.auth/schema/service.auth.graphqls
```diff
@@ -148,6 +148,13 @@ type UserTeamInfo {
...
+    tokenName: String!
+    token: String
+    periodDays: Int!
```
alexander-skoblikov: expiredAt/expirationTime/etc instead of period

server/bundles/io.cloudbeaver.service.auth/schema/service.auth.graphqls
alexander-skoblikov: and create separate model for create/list requests

server/bundles/io.cloudbeaver.service.security/db/cb_schema_update_22.sql
```diff
@@ -0,0 +1,11 @@
+CREATE TABLE {table_prefix}CB_ACCESS_TOKEN
+(
+    TOKEN_ID        VARCHAR(128) NOT NULL,
+    USER_ID         VARCHAR(128) NOT NULL,
+    TOKEN_NAME      VARCHAR(128) NOT NULL,
+    CREATE_TIME     TIMESTAMP DEFAULT CURRENT_TIMESTAMP NOT NULL,
+    EXPIRATION_TIME TIMESTAMP NULL,
+
+    PRIMARY KEY (USER_ID, TOKEN_ID),
```
alexander-skoblikov, line 9: also add key for USER_ID, TOKEN_NAME

Wroud (approved):

devnaumov (approved):

webapp/packages/plugin-authentication/src/Dialog/useAuthDialogState.ts
```diff
@@ -67,7 +67,7 @@ export function useAuthDialogState(accessRequest: boolean, providerId: string |
   const providers = authProvidersResource.data.filter(notEmptyProvider).sort(compareProviders);
 
   const activeProviders = providers.filter(provider => {
-    if (provider.federated || provider.trusted || provider.private) {
+    if (provider.federated || provider.trusted || provider.private || provider.authHidden) {
```
Wroud, line 70: why we can't use private? or trusted

sergeyteleshev: private does not fit here due to the fact we won't see the setting in the server config
trusted does not fit also here, but not sure why - the backend said so

we had a call with all of the flags which can fit to our needs, so we decided to extend the scheme cause there was no best candidate to fit our needs

PR #3028, opened 2024-10-23
Cb 5368 create api for kill session sm

devnaumov:


webapp/packages/plugin-data-viewer/src/DataViewerContextMenuService.ts
```diff
@@ -20,17 +24,18 @@ export interface IDataGridCellMenuContext {
 }
 
 @injectable()
-export class DataGridContextMenuService {
-  onRootMenuOpen: IExecutor<IDataGridCellMenuContext>;
-  static cellContext = 'data-grid-cell-context-menu';
-  private static readonly menuToken = 'dataGridCell';
+export class DataViewerContextMenuService {
```
Wroud: we really don't want to move this service outside of `plugin-data-viewer` because it's too wired with data presentation and not abstract from data model we also need to rewrite it to our new context menu api. I suggest to use bottom bar instead to avoid this changes

Wroud (changes requested):

PR #3033, opened 2024-10-27
Cb 5819 administration tables redesign

devnaumov:


webapp/packages/core-blocks/src/Containers/GroupTitle.tsx
```diff
@@ -15,16 +18,32 @@ import elementsSizeStyles from './shared/ElementsSize.module.css';
...
+      {onClose ? (
+        <Flex gap="xs" align="center">
+          <ActionIconButton className={s(styles, { closeButton: true })} name="angle" title={translate('ui_close')} onClick={onClose} />
+          {children}
+        </Flex>
+      ) : (
+        children
+      )}
+    </h2>
+  );
```
Wroud: please remove, use composition instead, another question we have `GroupClose` component what is wrong with it?

Wroud (changes requested):

webapp/packages/plugin-authentication-administration/src/Administration/Users/UsersTable/UsersTableOptionsPanelService.ts
```diff
@@ -0,0 +1,73 @@
...
+      return;
+    }
+
+    const result = await this.optionsPanelService.close();
+
+    if (result) {
+      this.userId = null;
+    }
+  }
+}
```
sergeyteleshev: I can see that this is exact same class as `TeamTableOptionsPanelService`. Do you think we can do something about logic duplication?

sergeyteleshev (approved):

Wroud (approved):

PR #3039, opened 2024-11-04
CB-5892 removes min fetch size setting

sergeyteleshev:


webapp/packages/plugin-data-viewer/src/DataViewerSettingsService.ts
```diff
@@ -21,14 +21,13 @@ import { schema, schemaExtra } from '@cloudbeaver/core-utils';
 
 import { DATA_EDITOR_SETTINGS_GROUP } from './DATA_EDITOR_SETTINGS_GROUP.js';
 
-const FETCH_MIN = 10;
+const FETCH_MIN = 100;
```
devnaumov: Why 100? Cant see this in ticket

Wroud (approved):

webapp/packages/plugin-data-viewer/src/DataViewerSettingsService.ts
sergeyteleshev: The case: We have 11k rows in the table and set the fetch size to 10. Now, we cannot load other data because the data loads in lazy mode when we scroll. We should probably add a button for this case. But for now, I set the minimum size to 100 so the user is not confused until we solve this issue.

Снимок экрана 2024-11-05 в 11 45 59

devnaumov (approved):

PR #3042, opened 2024-11-04
CB-5878 add title for inline editor button

SychevAndrey:


webapp/packages/core-localization/src/locales/en.ts
```diff
@@ -40,6 +40,7 @@ export default [
   ['ui_custom', 'Custom'],
   ['ui_configure', 'Configure'],
   ['ui_apply', 'Apply'],
+  ['ui_reject', 'Reject'],
```
devnaumov, line 44: We probably could use just ui_cancel?

devnaumov (changes requested):

webapp/packages/core-localization/src/locales/en.ts
SychevAndrey: We discussed it with @Wroud and he was for adding Reject option. Anyway there is no cancel token in locales, there is only `processing_cancel`, so we need to add either Reject or Cancel

devnaumov: Then we need to change ru locale to "Отклонить"

devnaumov (approved):

webapp/packages/core-localization/src/locales/en.ts
SychevAndrey: Personally I like more cancel option with "Отменить" translation. Can we discuss together and decide?

sergeyteleshev (approved):

PR #3051, opened 2024-11-06
СB-5796 Metadata/data editor - error when changing server settings

sergeyteleshev:


webapp/packages/plugin-connections/src/ConnectionAuthService.ts
```diff
@@ -104,7 +104,7 @@ export class ConnectionAuthService extends Dependency {
       if (resetCredentials && isConnectedInitially && result === DialogueStateResult.Rejected) {
         await this.connectionInfoResource.init(key);
       }
-    } else {
+    } else if (!connection.connected) {
       await this.connectionInfoResource.init({ projectId: key.projectId, connectionId: key.connectionId });
     }
```
Wroud: this is not true fix i think, because this change is hides case but not solves the problem.
I suppose that problem is that one of extra requests to inner database resources silently re-opens connection on the backend. We need to investigate this case and solve the problem properly

Wroud (changes requested):

devnaumov (approved):

webapp/packages/plugin-connections/src/ConnectionShield.tsx
```diff
@@ -23,10 +23,11 @@ export const ConnectionShield = observer<PropsWithChildren<IConnectionShieldProp
   const notificationService = useService(NotificationService);
 
   const connection = useResource(ConnectionShield, ConnectionInfoResource, connectionKey);
-  const connecting = getComputed(() => connectionKey && connection.resource.isConnecting(connectionKey));
+  const connecting = getComputed(() => (connectionKey && connection.resource.isConnecting(connectionKey)) || connection.loading);
+  const isConnectionReady = getComputed(() => !connecting && connection.data?.connected && connection.loaded && !connection.outdated);
```
devnaumov, line 27: Seems like we should implement some general solution for checking whether resource entity is fully available or not. Not now, but maybe in a tech sprint

devnaumov (approved):

Wroud (approved):

PR #3056, opened 2024-11-06
CB-5891 User - error after disabling and deleting

sergeyteleshev:


webapp/packages/plugin-authentication-administration/src/Administration/Users/UserForm/AdministrationUserForm.tsx
```diff
@@ -59,6 +59,10 @@ export const AdministrationUserForm = observer<Props>(function AdministrationUse
     },
   });
 
+  function onDisableUser() {
+    userFormInfoPart.state.enabled = false;
+  }
```
Wroud: seems like we need to create `disable` method in the form part and use it to disable user and sync form:
(pseudocode)
```
this.state.enabled = false;
await this.save()
```

Wroud (changes requested):

webapp/packages/plugin-authentication-administration/src/Administration/Users/UserForm/Info/UserFormInfoPart.ts
```diff
@@ -74,6 +74,11 @@ export class UserFormInfoPart extends FormPart<IUserFormInfoState, IUserFormStat
     return this.loaded;
   }
 
+  async disableUser() {
+    this.state.enabled = false;
+    await this.saveChanges();
```
Wroud: You need to call `save` to handle update and error handling logic (for example, if the user made some changes that will not pass validation, calling saveChanges directly will skip this logic, and it also will make it possible to run multiple concurrent save requests to the backend

Wroud (changes requested):

devnaumov (approved):

Wroud (approved):

PR #3061, opened 2024-11-07
Cb 5584 provide more information in the debug log for auditing

DenisSinelnikov:


server/bundles/io.cloudbeaver.server/plugin.xml
```diff
@@ -75,4 +75,12 @@
         </eventHandler>
     </extension>
 
+    <extension point="org.jkiss.dbeaver.settings">
+        <propertyGroup id="log" label="Logger">
+            <property id="log.on.show.variables" label="Show all variables in log"
```
alexander-skoblikov: strange naming

server/bundles/io.cloudbeaver.server/src/io/cloudbeaver/server/graphql/GraphQLLoggerUtil.java
```diff
@@ -0,0 +1,77 @@
...
+import java.util.Set;
+
+public class GraphQLLoggerUtil {
+
+    private static final Set<String> PROHIBITED_VARIABLES = GraphQLProhibitedVariables.getAllProhibitedVariables();
+
+    public static String getUserId(HttpServletRequest request) {
+        String userId = null;
+        WebSession session = (WebSession) CBApplication.getInstance().getSessionManager().getSession(request.getSession().getId());
+        if (session != null) {
```
serge-rider: request.getsession may be null

server/bundles/io.cloudbeaver.server/src/io/cloudbeaver/server/graphql/GraphQLLoggerUtil.java
```diff
@@ -0,0 +1,77 @@
...
+            }
+        }
+        return userId;
+    }
+
+    public static String buildLoggerMessage(String sessionId, String userId, Map<String, Object> variables) {
+        StringBuilder loggerMessage = new StringBuilder(" [user: ").append(userId)
+            .append(", sessionId: ").append(sessionId).append("]");
+
+        if (CBPlatform.getInstance().getPreferenceStore().getBoolean("log.debug") && variables != null) {
```
serge-rider: Why use pref store for that? Where `log.debug` is declared elsewhere?

server/bundles/io.cloudbeaver.server/src/io/cloudbeaver/server/graphql/GraphQLProhibitedVariables.java
```diff
@@ -0,0 +1,48 @@
...
+
+import java.util.Arrays;
+import java.util.List;
+import java.util.Set;
+import java.util.stream.Collectors;
+
+public enum GraphQLProhibitedVariables {
+
+
+    PASSWORD("password"),
```
serge-rider: It don't think that global enum will work. We need some explicit annotations for parameters which we do not want to log (for example as comments).
However we need to move log invocation into WebExecutionStrategy to get information about GQL endpoint.

If this is too complicated then we can keep it as is for now. In this case add "credentials" and "username" too. 
However I don't understand why we need enum but simple Set<String>

serge-rider (changes requested):

server/bundles/io.cloudbeaver.server/src/io/cloudbeaver/server/graphql/GraphQLProhibitedVariables.java
DenisSinelnikov: let's leave Set for now, added additional parameters, let's start a task on improve with annotations, need more time for realization

PR #3088, opened 2024-11-20
CB-5926 refactor: sort settings by translation

SychevAndrey:


webapp/packages/plugin-settings-panel/src/SettingsPanel/useSettings.ts
```diff
@@ -27,7 +28,13 @@ export function useSettings(accessor?: string[]): ISettings {
...
         const settings = this.settingsManagerService.activeSettings
           .filter(setting => accessor?.some(value => setting.access.scope.includes(value)))
-          .sort((a, b) => a.name.localeCompare(b.name));
+          .sort((a, b) => {
+            if (a.type === b.type) {
+              return translate(a.name).localeCompare(translate(b.name));
+            } else {
+              return a.type - b.type;
+            }
+          });
```
Wroud, line 37: I'm really not sure that we need to group settings by type because it can look better in a small amount of settings but will be unreadable with many

SychevAndrey: image
image

For me settings of the same type is more likely to be closer by meaning. See how it looks w/o sorting by type

PR #3089, opened 2024-11-20
CB-5960 sort features

devnaumov:


SychevAndrey (approved):

webapp/packages/core-root/src/FeaturesResource.ts
```diff
@@ -19,6 +19,11 @@ export type ApplicationFeature = WebFeatureSet;
 @injectable()
 export class FeaturesResource extends CachedDataResource<ApplicationFeature[]> {
   private baseFeatures: string[];
+
+  get features(): ApplicationFeature[] {
+    return this.data.slice().sort(sortFeature);
+  }
```
Wroud: not sure that sorting on resource level is good idea, it's better to sort it in UI component instead

PR #3096, opened 2024-11-22
CB-5125 do not trigger logo in configuration stage

devnaumov:


SychevAndrey (changes requested):
plugin-top-bar-app cannot use the administration package, it's better to branch out from CB-5918 and make changes in plugin-app-logo-administration

webapp/packages/plugin-top-app-bar/src/TopNavBar/Logo.tsx
```diff
@@ -26,5 +28,5 @@ export const Logo = observer(function Logo() {
 
   const title = isSameVersion ? backendVersionTitle : commonVersionTitle;
 
-  return <AppLogo title={title} onClick={() => screenService.navigateToRoot()} />;
+  return <AppLogo title={title} onClick={administrationScreenService.isConfigurationMode ? undefined : () => screenService.navigateToRoot()} />;
```
Wroud: it's better to fix somehow in `ScreenService` or in `AppScreenBootstrap` (so basically we need to block public screen activation in case we in configuration mode (we can check it from server config for not administration pages))

devnaumov: Plugins can use any core* packages. 
I am not sure about adding this logic to the screen service. It should not check itself whether it can be navigate or not, it should be checked outside of this service. Cause screen service has single responsibility, we want to keep it this way

Wroud (approved):

SychevAndrey (approved):

PR #3101, opened 2024-11-27
CB-4070 Migrate Project form to form template

sergeyteleshev:


webapp/packages/core-blocks/src/StatusMessage.tsx
```diff
@@ -66,10 +66,10 @@ export const StatusMessage = observer<Props>(function StatusMessage({
...
             {onShowDetails ? (
               <Link className={s(styles, { link: !multipleRows })} onClick={onShowDetails}>
-                {message}
+                {translate(message)}
               </Link>
             ) : (
-              message
+              translate(message)
             )}
           </div>
```
Wroud: probably we need to translate it in div's title too?

webapp/packages/core-ui/src/Form/FormPart.ts
```diff
@@ -17,6 +17,7 @@ export abstract class FormPart<TPartState, TFormState = any> implements IFormPar
   state: TPartState;
   initialState: TPartState;
   isSaving: boolean;
+  saved: boolean;
```
Wroud: Why do we need this flag? Our design was based on comparing the initial and current states, so the initial always needs to be synchronized with the server. When the initial state and current state are equal, then the  form is saved

How is this flag managed? What if the form gets a new initial state, and the current state will not equal the initial state? Will the form be marked as saved?

Wroud (changes requested):

webapp/packages/core-ui/src/Form/FormPart.ts
sergeyteleshev: removed

Wroud (approved):

PR #3132, opened 2024-12-12
CB-6000 handle task info events from WS

SychevAndrey:
1. `AsyncTaskService` was moved to core-root to avoid circular dependencies
2. Task is added to tasks map only when getter brings data from backend to save real task ID (`AsyncTask.create`)
3. `updateStatus` method was added to update task status on websocket message if task is running
4. When `running` prop is changing to `false`, we request gql endpoint
5. Naming improvements (onEvent  return type Subscription -> Unsubscribe) to avoid confusion with Rx Subscription plus it's basically returns cancel function, not subscription in any sense.

### DIFF
--- a/webapp/packages/core-root/src/AsyncTask/AsyncTask.ts
+++ b/webapp/packages/core-root/src/AsyncTask/AsyncTask.ts
@@ -8,7 +8,7 @@
 import { computed, makeObservable, observable } from 'mobx';
 
 import { type ISyncExecutor, SyncExecutor } from '@cloudbeaver/core-executor';
-import { type AsyncTaskInfo, ServerInternalError } from '@cloudbeaver/core-sdk';
+import { type AsyncTaskInfo, ServerInternalError, type WsAsyncTaskInfo } from '@cloudbeaver/core-sdk';
 
 export class AsyncTask {
   get id(): string {
@@ -125,9 +125,11 @@ export class AsyncTask {
     }
   }
 
-  public updateStatus(info: AsyncTaskInfo): void {
-    this.taskInfo = info;
-    this.onStatusChange.execute(this.taskInfo);
+  public updateStatus(info: WsAsyncTaskInfo): void {
+    if (this.taskInfo) {
+      this.taskInfo.status = info.statusName;
+      this.onStatusChange.execute(this.taskInfo);
+    }
   }
 
   private updateInfo(info: AsyncTaskInfo): void {
--- a/webapp/packages/core-root/src/AsyncTask/AsyncTask.ts
+++ b/webapp/packages/core-root/src/AsyncTask/AsyncTask.ts
@@ -8,10 +8,8 @@
 import { computed, makeObservable, observable } from 'mobx';
 
 import { type ISyncExecutor, SyncExecutor } from '@cloudbeaver/core-executor';
+import { type AsyncTaskInfo, ServerInternalError } from '@cloudbeaver/core-sdk';
 import { uuid } from '@cloudbeaver/core-utils';
-
-import type { AsyncTaskInfo } from '../sdk.js';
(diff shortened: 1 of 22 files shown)


webapp/packages/core-root/src/AsyncTask/AsyncTask.ts
```diff
@@ -8,13 +8,18 @@
...
-  readonly id: string;
+  get id(): string {
+    return this._id;
+  }
+
+  set id(id: string) {
+    if (!this._id) {
+      this._id = id;
+    }
+  }
```
Wroud: why it's private then? it can be just `id: string` instead of `_id` and get + set

webapp/packages/core-root/src/AsyncTask/AsyncTask.ts
```diff
@@ -34,18 +39,19 @@ export class AsyncTask {
...
   private updatingAsync: boolean;
   private readonly init: () => Promise<AsyncTaskInfo>;
-  private readonly cancel: (info: AsyncTaskInfo) => Promise<void>;
+  private readonly cancel: (id: string) => Promise<void>;
   private initPromise: Promise<void> | null;
 
-  constructor(init: () => Promise<AsyncTaskInfo>, cancel: (info: AsyncTaskInfo) => Promise<void>) {
-    this.id = uuid();
+  constructor(init: () => Promise<AsyncTaskInfo>, cancel: (id: string) => Promise<void>) {
+    this._id = '';
```
Wroud: kinda incorrect, cuz attempt to access id field after object creation will return incorrect data

webapp/packages/core-root/src/AsyncTask/AsyncTaskInfoService.ts
```diff
@@ -0,0 +1,117 @@
...
+      if (info.id && !this.tasks.has(info.id)) {
+        this.tasks.set(info.id, task);
+        task.id = info.id;
+        if (this.tasks.size === 1) {
+          this.connection = this.asyncTaskInfoEventHandler.eventsSubject.connect();
+        }
+      }
+
+      return info;
+    }, this.cancelTask.bind(this));
```
Wroud: please subscribe to `onStatusChange` instead to update task and AsyncTaskInfo.id linking

Wroud (changes requested):

webapp/packages/core-root/src/AsyncTask/AsyncTaskInfoService.ts
```diff
@@ -0,0 +1,155 @@
...
+  create(getter: () => Promise<AsyncTaskInfo>): AsyncTask {
+    const task = new AsyncTask(getter, this.cancelTask.bind(this));
+
+    this.tasks.set(task.id, task);
+    task.onStatusChange.addHandler(info => {
+      const pendingEvent = this.pendingEvents.get(info.id);
+      if (pendingEvent) {
+        const changes = { ...pendingEvent };
+        this.pendingEvents.delete(info.id);
+        this.updateTask(task, changes);
```
Wroud: you don't need to create a object copy cuz it will be used only once so we don't care about possible mutations

webapp/packages/core-root/src/AsyncTask/AsyncTaskInfoService.ts
```diff
@@ -0,0 +1,155 @@
...
+      }
+      if (!this.taskIdAliases.get(info.id)) {
+        this.taskIdAliases.set(info.id, task.id);
+      }
```
Wroud: it's better to isolate this logic to "only first time"
1. check if we have task alias already, if we do: do nothing
2. else set task alias, then check for pending events and process if we have some

Wroud (changes requested):

Wroud (approved):

PR #3140, opened 2024-12-17
Cb 5448 sql assist doesnt work after alias with enabled use long object names

sergeyteleshev:


webapp/packages/plugin-sql-editor-new/src/SQLEditor/useSqlDialectAutocompletion.ts
```diff
@@ -80,7 +80,7 @@ export function useSqlDialectAutocompletion(data: ISQLEditorData): [Compartment,
         const limitIsMet = optionsRef.data.hintsLimitIsMet;
 
         const result: CompletionResult = {
-          from: word.from,
+          from: getFrom(proposals, word.from),
```
devnaumov: We should use built in codemirror tools to apply proposal the way we want. It could be done by passing custom function
https://codemirror.net/docs/ref/#autocomplete.Completion.apply

devnaumov (changes requested):

devnaumov (approved):

PR #3145, opened 2024-12-20
#3143: Add number conversion

### LINKED ISSUE #3143: Add thousands separator to numbers
It would be immensely useful if cloudbeaver displayed numbers in the data view pane with a thousands separator (maybe with an optional toggle in preferences if it bothers someone).
Currently it does not:
image
Whereas dbeaver does:
image
Otherwise seeing scales of numbers is very difficult

JFYI I'm aware of a [similar issue](https://github.com/dbeaver/cloudbeaver/issues/2997) but it's for data not for row count

phantom943:
Closes #3143

### DIFF
--- a/webapp/packages/plugin-data-viewer/src/DatabaseDataModel/Actions/ResultSet/ResultSetFormatAction.ts
+++ b/webapp/packages/plugin-data-viewer/src/DatabaseDataModel/Actions/ResultSet/ResultSetFormatAction.ts
@@ -108,6 +108,19 @@ export class ResultSetFormatAction
     return false;
   }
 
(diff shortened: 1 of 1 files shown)


LonwoLonwo:
Hello @phantom943 

Thanks for your contribution.
We will look at it and give you the answer.

LonwoLonwo:
Hello. 
Unfortunately, your version is too local ("en-US").
Our application is for the whole world, not just for some regions.
We also need to consider different types of numbers splitting in different regions. Probably, we should add some visual settings in the Administration menu. 
We will implement your request on our end.
Thanks again for the contribution.

phantom943:
Hello
Sure. Waiting for your fix, then

PR #3151, opened 2024-12-23
CB-3753-1809-fr-add-field-hints-in-the-filter-field

sergeyteleshev:


webapp/packages/core-blocks/src/FormControls/useInputAutocomplete.ts
```diff
@@ -0,0 +1,154 @@
...
+import { debounce, isFuzzySearchable, isNotNullDefined } from '@cloudbeaver/core-utils';
+
+import { useObservableRef } from '../useObservableRef.js';
+
+export type InputAutocompleteStrategy = 'startsWith' | 'contains' | 'fuzzy';
+
+interface InputAutocompleteOptions {
+  sourceHints: InputAutocompleteProposal[];
+  matchStrategy?: InputAutocompleteStrategy;
+  filter?: (suggestion: InputAutocompleteProposal, lastWord?: string) => boolean;
```
SychevAndrey: I believe `predicate` is more descriptive word for that prop, then filter.

webapp/packages/core-blocks/src/FormControls/useInputAutocomplete.ts
```diff
@@ -0,0 +1,154 @@
...
+        const input = this.inputRef.current;
+
+        if (!input) {
+          return;
+        }
+
+        const cursorPosition = this.selectionStart;
+        const words = this.input?.split(' ');
+
+        if (!isNotNullDefined(words) || !isNotNullDefined(cursorPosition) || !isNotNullDefined(this.currentWord)) {
```
SychevAndrey: Do we need to check words and cursor here if currentWord getter should return falsy value if there is no words or cursorPosition?

webapp/packages/core-blocks/src/FormControls/useInputAutocomplete.ts
```diff
@@ -0,0 +1,154 @@
...
+            }
+
+            return (
+              (this.matchStrategy === 'startsWith' &&
+                values.some(value => isNotNullDefined(this.currentWord) && value.startsWith(this.currentWord))) ||
+              (this.matchStrategy === 'contains' && values.some(value => isNotNullDefined(this.currentWord) && value.includes(this.currentWord))) ||
+              (this.matchStrategy === 'fuzzy' &&
+                values.some(value => isNotNullDefined(this.currentWord) && isFuzzySearchable(this.currentWord, value)))
+            );
+          })
```
SychevAndrey:
```suggestion
const isMatch = () => {
  const { matchStrategy, currentWord } = this;

  if (!isNotNullDefined(currentWord)) {
    return false;
  }

  const matchFunctions: { [key: InputAutocompleteStrategy]: (value: string) => boolean } = {
    startsWith: value => value.startsWith(currentWord),
    contains: value => value.includes(currentWord),
    fuzzy: value => isFuzzySearchable(currentWord, value),
  };

  return values.some(matchFunctions[matchStrategy]);
};
```


Maybe like that? And we checked thus.currentWord just above, so maybe we can skip this check:
```
 if (!isNotNullDefined(currentWord)) {
    return false;
  }
```

SychevAndrey (approved):
👍🏻

webapp/packages/core-utils/src/isFuzzySearchable.ts
```diff
@@ -0,0 +1,28 @@
+/*
+ * CloudBeaver - Cloud Database Manager
+ * Copyright (C) 2020-2024 DBeaver Corp and others
+ *
+ * Licensed under the Apache License, Version 2.0.
+ * you may not use this file except in compliance with the License.
+ */
+export function isFuzzySearchable(pattern: string, text: string): boolean {
```
devnaumov: Maybe we should use fuzzy library?

webapp/packages/core-blocks/src/FormControls/InputAutocompletionMenu.tsx
```diff
@@ -0,0 +1,104 @@
...
+  useEffect(() => {
+    if (menuRef.current?.visible && (autocompleteState.filteredSuggestions === null || autocompleteState.filteredSuggestions.length === 0)) {
+      menuRef.current?.hide();
+    }
+    if (!menuRef.current?.visible && autocompleteState.filteredSuggestions !== null && autocompleteState.filteredSuggestions.length !== 0) {
+      menuRef.current?.show();
+    }
+  }, [sourceHints, menuRef, autocompleteState.filteredSuggestions]);
+
+  return (
```
devnaumov, line 53: We should probably use combobox component as its based on the input and supports search

devnaumov (changes requested):

webapp/packages/core-blocks/src/FormControls/InputAutocompletionMenu.tsx
sergeyteleshev: seems redundant for me. we should implement it once we migrate from reakit to something better. cause for now matching our combobox + reakit menus with our wrapper over it `Menu.tsx` seems like a nightmare

webapp/packages/core-blocks/src/FormControls/useInputAutocomplete.ts
sergeyteleshev: done

webapp/packages/core-blocks/src/FormControls/useInputAutocomplete.ts
sergeyteleshev: done

webapp/packages/core-blocks/src/FormControls/useInputAutocomplete.ts
sergeyteleshev: done

webapp/packages/core-blocks/package.json
```diff
@@ -30,6 +30,7 @@
     "@cloudbeaver/core-sdk": "^0",
     "@cloudbeaver/core-theming": "^0",
     "@cloudbeaver/core-utils": "^0",
+    "fuse.js": "^7",
```
Wroud: What options do we have besides this library? (I'm curious because it's not actively maintained; maybe other options have more active support)

webapp/packages/core-blocks/src/FormControls/useInputAutocomplete.ts
```diff
@@ -0,0 +1,182 @@
...
+
+  const handleInput = debounce((event: Event) => {
+    const target = event.target as HTMLInputElement;
+
+    state.selectionStart = target.selectionStart;
+    state.selectionEnd = target.selectionEnd;
+    state.input = target?.value;
+    setFuzzySearchValue(state.currentWord);
+    state.isFound = false;
+  }, INPUT_DELAY);
```
Wroud: Without memo this function will not work as expected in case its call trigger component re-render

SychevAndrey (approved):

webapp/packages/core-blocks/src/FormControls/InputAutocompletionMenu.module.css
```diff
@@ -0,0 +1,37 @@
...
+
+  & .itemTitle {
+    position: relative;
+  }
+
+  & .iconOrImage {
+    width: 16px;
+    height: 16px;
+  }
+}
```
Wroud, line 11: please reuse `BaseDropdownStyles` instead

webapp/packages/core-blocks/src/Menu/Menu.tsx
```diff
@@ -95,34 +100,61 @@ export const Menu = observer<IMenuProps, HTMLButtonElement>(
...
-        menu.hide();
-        return;
-      }
+        const span = document.createElement('span');
+        span.style.position = 'absolute';
+        span.style.visibility = 'hidden';
+        span.style.whiteSpace = 'pre';
+        span.textContent = contextInputRef.current.value.substring(0, caretPosition);
+
+        document.body.appendChild(span);
```
Wroud: Please use the menu anchor (invisible menu trigger / button) instead (you can position it as you need)

webapp/packages/core-ui/src/InlineEditor/InlineEditor.tsx
```diff
@@ -28,6 +39,7 @@ export interface InlineEditorProps extends Omit<React.InputHTMLAttributes<HTMLIn
   autofocus?: boolean;
   active?: boolean;
   loading?: boolean;
+  autoCompleteProposals?: InputAutocompleteProposal[];
```
Wroud: please do not extend `InlineEditor` functionality cuz its deprecated and we want to use our base controls instead

webapp/packages/plugin-data-viewer/src/TableViewer/TableHeader/useWhereFilter.ts
```diff
@@ -20,13 +22,48 @@ interface IState {
   resultIndex: number;
   readonly supported: boolean;
   readonly filter: string;
+  readonly columns: SqlResultColumn[];
+  readonly hintProposals: InputAutocompleteProposal[];
   readonly constraints: DatabaseDataConstraintAction | null;
```
Wroud: please abstract where filter from columns or structure of the data

Wroud (changes requested):

webapp/packages/core-blocks/src/FormControls/InputAutocompletionMenu.tsx
```diff
@@ -0,0 +1,107 @@
...
+  matchStrategy,
+  inputRef,
+  onSelect,
+}: AutocompletionProps) {
+  const styles = useS(style, BaseDropdownStyles);
+  const menuRef = useRef<IMenuState>();
+  const autocompleteState = useInputAutocomplete(inputRef, {
+    sourceHints,
+    matchStrategy,
+  });
```
Wroud: please move this hook outside from the component, so you can use it and use input ref in the same component

webapp/packages/core-blocks/src/FormControls/InputAutocompletionMenu.tsx
```diff
@@ -0,0 +1,107 @@
...
+        break;
+    }
+  }
+
+  useEffect(() => {
+    inputRef.current?.addEventListener('keydown', handleKeyDown);
+    return () => {
+      inputRef.current?.removeEventListener('keydown', handleKeyDown);
+    };
+  }, [inputRef.current]);
```
Wroud: this all can be part of the hook probably

webapp/packages/core-blocks/src/Menu/Menu.tsx
```diff
@@ -22,6 +24,7 @@ import type { IMouseContextMenu } from './useMouseContextMenu.js';
 
 interface IMenuProps extends React.ButtonHTMLAttributes<any> {
   mouseContextMenu?: IMouseContextMenu;
+  contextInputRef?: React.RefObject<HTMLInputElement | HTMLTextAreaElement>;
```
Wroud: looks like it's not a part of the menu

Wroud (changes requested):

webapp/packages/core-blocks/src/useFuzzySearch.ts
```diff
@@ -0,0 +1,51 @@
...
+
+interface UseSearchProps<T> extends Options {
+  sourceProposals: T[];
+}
+
+const DEFAULT_THRESHOLD = 0.4;
+
+function createSearchInstance<T>(options: Options) {
+  return new MiniSearch<T>(options);
+}
```
Wroud: strange function

webapp/packages/core-blocks/src/useFuzzySearch.ts
```diff
@@ -0,0 +1,51 @@
...
+
+export function useFuzzySearch<T extends object>({ sourceProposals, ...options }: UseSearchProps<T>) {
+  const minisearchOptions = { ...options };
+  const dataSetKeys = sourceProposals.reduce((acc, item) => {
+    Object.keys(item).forEach(key => acc.add(key));
+
+    return acc;
+  }, new Set<string>());
+  const searchInstance = useRef(createSearchInstance<T>(minisearchOptions));
```
Wroud: it will create new instance every render, the better solution is to use `useState` or:
```
const searchInstance = useRef(null);
if (searchInstance === null) {
  searchInstance.current = new MiniSearch<T>(options);
}
```

webapp/packages/core-blocks/src/useFuzzySearch.ts
```diff
@@ -0,0 +1,51 @@
...
+  }
+
+  if (!minisearchOptions.storeFields) {
+    minisearchOptions.storeFields = Array.from(dataSetKeys);
+  }
+
+  useEffect(() => {
+    searchInstance.current = createSearchInstance<T>(minisearchOptions);
+    searchInstance.current.addAll(sourceProposals);
+  }, [sourceProposals]);
```
Wroud: 1. why we re-create instance?
2. why we use `useEffect`?

webapp/packages/core-blocks/src/useFuzzySearch.ts
```diff
@@ -0,0 +1,51 @@
...
+  return new MiniSearch<T>(options);
+}
+
+export function useFuzzySearch<T extends object>({ sourceProposals, ...options }: UseSearchProps<T>) {
+  const minisearchOptions = { ...options };
+  const dataSetKeys = sourceProposals.reduce((acc, item) => {
+    Object.keys(item).forEach(key => acc.add(key));
+
+    return acc;
+  }, new Set<string>());
```
Wroud: it seems that logic is overcomplicated in this hook and can be simplifyed:
```
export function useFuzzySearch<T extends object>({ sourceProposals, ...options }: UseSearchProps<T>) {
  const minisearchRef = useRef(null);
  const currentOptions = useRef({});

  if (minisearchRef.current === null || !isObjectEqual(currentOptions.current, options)) {
    minisearchRef.current = new Minisearch(options);
    currentOptions.current = options;
  }
  
  useMemo(() => autorun(() => { // here we ensure that mutable props from mobx object will be traced correctly
    minisearchRef.current.removeAll();
    minisearchRef.current.addAll(sourceProposals);
  }), [sourceProposals]) // possible we can tune how we determine changes in sourceProposals (we can write our own logic for example with use of isArrayEqual)
  
  ....
```

Wroud: i'm not sure that we need to index all fields passed  as proposals it's better to declare fields manually with type: `fields: Array<keyof T>`

webapp/packages/core-blocks/src/useFuzzySearch.ts
```diff
@@ -0,0 +1,51 @@
...
+}
+
+const DEFAULT_THRESHOLD = 0.4;
+
+function createSearchInstance<T>(options: Options) {
+  return new MiniSearch<T>(options);
+}
+
+export function useFuzzySearch<T extends object>({ sourceProposals, ...options }: UseSearchProps<T>) {
+  const minisearchOptions = { ...options };
```
Wroud: in case this hook is specific binding of minisearch for react, then we probably want to keep original interface as much as possible

webapp/packages/core-blocks/src/FormControls/useInputAutocomplete.ts
```diff
@@ -0,0 +1,141 @@
...
+export type InputAutocompleteStrategy = 'startsWith' | 'contains' | 'fuzzy';
+
+interface InputAutocompleteOptions {
+  sourceHints: InputAutocompleteProposal[];
+  matchStrategy?: SearchStrategy;
+  predicate?: (suggestion: InputAutocompleteProposal, lastWord?: string) => boolean;
+}
+
+export interface InputAutocompleteProposal {
+  id: string;
```
Wroud: i suggest to remove this 'id' fields because:
1. it's not in use
2. it's just a "always" random string

webapp/packages/core-blocks/src/Menu/useContextMenuPosition.ts
```diff
@@ -10,17 +10,15 @@ import { observable } from 'mobx';
...
-  x: number;
-  y: number;
-}
-
-export interface IMouseContextMenu {
-  position: IContextMenuPosition | null;
+  position: {
+    x: number;
+    y: number;
+  } | null;
```
Wroud: I don't personally like inlining public types cuz it's impossible to use them if needed only with `SomeType['someKey']`

Wroud (changes requested):

Wroud:
please take a look at this: https://github.com/lucaong/react-minisearch
and this: https://github.com/lucaong/react-minisearch/issues/38#issuecomment-1619129761

webapp/packages/core-blocks/src/FormControls/InputAutocompletionMenu.tsx
```diff
@@ -0,0 +1,156 @@
...
+
+  useEffect(() => {
+    inputRef.current?.addEventListener('keydown', handleKeyDown);
+    inputRef.current?.addEventListener('input', handleInput);
+
+    return () => {
+      inputRef.current?.removeEventListener('keydown', handleKeyDown);
+      inputRef.current?.removeEventListener('input', handleInput);
+    };
+  }, [inputRef.current]);
```
Wroud: `React.RefObject` can't be used this way because:
1. it's passed as a component property
2. it's mutated by react and will not cause child component (this component) to re-render when the `current` changed

if you need some logic related to some input ref please place this logic in some hook and use this hook in the same component that passes ref to input to avoid synchronization bugs

webapp/packages/core-blocks/src/FormControls/useInputAutocomplete.ts
```diff
@@ -0,0 +1,139 @@
...
+    searchFields: SEARCH_FIELDS,
+    matchStrategy,
+    predicate,
+  });
+
+  const state = useObservableRef(
+    () => ({
+      isFound: false,
+      input: inputRef.current?.value as string | undefined,
+      selectionStart: inputRef.current?.selectionStart ?? null,
```
Wroud: this initial value will be incorrect because at the moment of initialization there may be no `current` object yet

webapp/packages/core-blocks/src/useFuzzySearch.ts
```diff
@@ -0,0 +1,51 @@
...
+ */
+import { autorun, observable } from 'mobx';
+import { useMemo } from 'react';
+import { useMiniSearch } from 'react-minisearch';
+
+import { useObservableRef } from './useObservableRef.js';
+
+interface UseSearchProps<T> {
+  sourceProposals: T[];
+  fields: string[];
```
Wroud: why not to use `Array<keyof T>` for `fields`?

webapp/packages/core-blocks/src/useFuzzySearch.ts
```diff
@@ -0,0 +1,51 @@
...
+  });
+  const state = useObservableRef(
+    () => ({}),
+    {
+      engine: observable.ref,
+    },
+    {
+      engine: miniSearch,
+    },
+  );
```
Wroud: I will suggest removing this part for now. We may implement it later when we have different engines (if we have them), and we can't be sure that implementation will be like this if we imagine 3-4 engines written similarly. Then it may be performance issues (right now, we indexing every re-render miniseach even if the engine will be some other one)

webapp/packages/core-blocks/src/useFuzzySearch.ts
```diff
@@ -0,0 +1,51 @@
...
+const DEFAULT_THRESHOLD = 0.4;
+
+export function useFuzzySearch<T extends object>({ sourceProposals, fields, threshold = DEFAULT_THRESHOLD, prefix = true }: UseSearchProps<T>) {
+  const miniSearch = useMiniSearch(sourceProposals, {
+    fields,
+    searchOptions: {
+      fuzzy: threshold,
+      prefix,
+    },
+  });
```
Wroud: this way documents are already added on render, so please replace `sourceProposals` with `[]` or remove the `useMemo` block

webapp/packages/core-blocks/src/useSearch.ts
```diff
@@ -0,0 +1,107 @@
...
+
+              if (!suggestion) {
+                return false;
+              }
+
+              return filterBase(suggestion, this.searchFields, searchWord, this.predicate);
+            },
+          });
+          return;
+        }
```
Wroud: Why do we use early return here? Is it expected that with a `fuzzy` strategy, `setSeach` will not change the `search` state?

webapp/packages/plugin-data-viewer/src/TableViewer/TableHeader/useTableViewerHeaderData.ts
```diff
@@ -0,0 +1,95 @@
...
+        return [...BASE_HINTS].concat(
+          this.columns.map(column => ({
+            id: uuid(),
+            title: column.label || '',
+            displayString: column.label || '',
+            replacementString: column.label || '',
+            icon: column.icon || '',
+            score: 1,
+          })),
+        );
```
Wroud: please remove extra fields

Wroud (changes requested):

webapp/packages/core-blocks/src/FormControls/InputAutocompletionMenu.tsx
```diff
@@ -0,0 +1,81 @@
...
+  className,
+  menuRef,
+  proposals,
+  onSelect,
+}: AutocompletionProps) {
+  const styles = useS(style, BaseDropdownStyles);
+  const contextMenuPosition = getComputed(() => ({
+    position,
+    handleContextMenuOpen: () => {},
+  }));
```
Wroud: getComputed will not help cuz function always returns new object (getComputed works if you can return stable value, for example for primitives)

webapp/packages/core-blocks/src/FormControls/useInputAutocomplete.ts
```diff
@@ -0,0 +1,187 @@
...
+    const input = state.inputRef.current!;
+
+    input.addEventListener('input', handleInput);
+    input.addEventListener('keydown', handleKeyDown);
+
+    return () => {
+      input.removeEventListener('keydown', handleKeyDown);
+      input.removeEventListener('input', handleInput);
+    };
+  }, []);
```
Wroud: This will lead to unexpected behavior if the input is unmounted because the hook is executed only once at the first render. I will suggest to remove deps ([]) to run this hook every re-render

Wroud (approved):

devnaumov (approved):

PR #3155, opened 2024-12-24
CB-6051 ability to download drivers

yagudin10:


sergeyteleshev (approved):

webapp/packages/plugin-connection-custom/src/DriverSelector/Driver.module.css
```diff
@@ -5,8 +5,24 @@
...
   max-height: 24px;
-};
+}
+
+.indicator {
+  position: absolute;
+  bottom: 2px;
+  right: 0;
+  width: 12px;
+  height: 12px;
```
sergeyteleshev: may be we need a bit more larger icon? this one seems too small for me
image

SychevAndrey: Yep, If we expect users to hover over that icon, it's better to have it at least 14px or 16px. [WCAG recommends minimum 24px](https://www.w3.org/WAI/WCAG22/Understanding/target-size-minimum.html), however in our case it's not a button and maybe not so critical, so we can make it smaller.
And I personally would place it on top. For users it will make more sense in terms of catching their attention since such approach is commonly used among many applications for different type of notifications.

SychevAndrey: Actually, I have a better Idea. Why don't we just make the whole driver icon hoverable? I assume this is how it should work. Icon itself doesn't need to be a target

sergeyteleshev (approved):

webapp/packages/plugin-connections/src/ConnectionForm/Options/Options.tsx
```diff
@@ -187,7 +191,12 @@ export const Options: TabContainerPanelComponent<IConnectionFormProps> = observe
                 tiny
                 fill
               >
-                {translate('connections_connection_driver')}
+                <Flex gap="xs">
+                  {isAdmin && !driver?.driverInstalled && (
+                    <IconOrImage icon="/icons/info_icon_sm.svg" title={translate('plugin_connections_connection_driver_not_installed')} />
+                  )}
+                  {translate('connections_connection_driver')}
+                </Flex>
```
Wroud, line 205: Do we display uninstalled drivers in the list of available drivers? What will happen if I create a connection for such a driver?

sergeyteleshev (approved):

Wroud (changes requested):
please don't show uninstalled drivers in the list for not administrator users

Wroud (approved):

PR #3188, opened 2025-01-10
CB-6063 feat: add ability to import data without PK

SychevAndrey:
[video](https://github.com/user-attachments/assets/df7ac05e-6aa0-4f83-8ab0-e82dfd97c6e5)


webapp/packages/plugin-data-import/src/DataImportBootstrap.ts
```diff
@@ -103,7 +103,7 @@ export class DataImportBootstrap extends Bootstrap {
         const presentation = context.get(DATA_CONTEXT_DV_PRESENTATION);
         const isContainer = model.source instanceof ContainerDataSource;
         return (
-          !model.isReadonly(resultIndex) &&
```
devnaumov: Fix is not clear for me. Why do we need this? Previous logic seems fine

devnaumov: You probably need to use hasRowIdentifier flag to check whether table has pk

devnaumov (changes requested):

webapp/packages/plugin-data-import/src/DataImportBootstrap.ts
SychevAndrey: this isReadonly method uses hasRowIdentifier inside, which we should ignore. How do you see this c check here?

devnaumov: We probably want to add new flag something like isAppendAllowed as adding new row from ui is not allowed but import is allowed that basically does the same

SychevAndrey: No, this is not true

SychevAndrey:
So we discussed with @Wroud and agreed that the only option to add the hasRowIdentifier flag because it is used in different places, not only related to insertion. This is how it works now:
https://github.com/user-attachments/assets/df7ac05e-6aa0-4f83-8ab0-e82dfd97c6e5

webapp/packages/plugin-data-spreadsheet-new/src/DataGrid/DataGridContextMenu/DataGridContextMenuCellEditingService.ts
```diff
@@ -86,29 +86,32 @@ export class DataGridContextMenuCellEditingService {
         const isComplex = format.isBinary(key) || format.isGeometry(key);
         const isTruncated = content.isTextTruncated(key);
         const selectedElements = select?.getSelectedElements() || [];
+        const hasRowIdentifier = model.hasRowIdentifier(resultIndex);
+        // we can't edit table cells if table doesn't have row identifier, but we can edit new created rows before insert (CB-6063)
+        const canEdit = hasRowIdentifier || (!hasRowIdentifier && editor.getElementState(key) === DatabaseEditChangeType.add);
```
Wroud: this is equal to `hasRowIdentifier || editor.getElementState(key) === DatabaseEditChangeType.add`

webapp/packages/plugin-data-viewer/src/DatabaseDataModel/DatabaseDataSource.ts
```diff
@@ -206,6 +206,10 @@ export abstract class DatabaseDataSource<TOptions, TResult extends IDatabaseData
     return !this.isLoading() && !this.disabled;
   }
 
+  hasRowIdentifier(resultIndex: number): boolean {
+    return this.getResult(resultIndex)?.data?.hasRowIdentifier === true;
+  }
```
Wroud: we need to abstract it from Row because we have documents also where we don't have rows but have documents (json), i suggest to use `hasElementIdentifier` or `hasIdentifiers`

Wroud (changes requested):

Wroud (approved):

PR #3209, opened 2025-01-24
CB-6152 restrict editing bool and hide actions in datasets

SychevAndrey:


webapp/packages/plugin-data-viewer/src/TableViewer/TableFooter/TableFooterMenu/TableFooterMenuService.ts
```diff
@@ -61,7 +62,7 @@ export class TableFooterMenuService {
         const model = context.get(DATA_CONTEXT_DV_DDM)!;
         const resultIndex = context.get(DATA_CONTEXT_DV_DDM_RESULT_INDEX)!;
 
-        if (model.isReadonly(resultIndex)) {
+        if (model.isReadonly(resultIndex) || !(model.source instanceof ContainerDataSource)) {
```
Wroud: why not `ResultSetDataSource`? why we restrict for all non SQL based tables?

Wroud (changes requested):

webapp/packages/plugin-data-viewer/src/TableViewer/TableFooter/TableFooterMenu/TableFooterMenuService.ts
SychevAndrey: removed that check, added check that we have one of any edit buttons. Actually it's only needed for "select 1" case

https://github.com/user-attachments/assets/72f2963a-3be5-4bd1-80e1-f0160f368134

sergeyteleshev (approved):

Wroud (approved):

devnaumov (approved):

PR #3214, opened 2025-01-27
CB-6168 map user role by team

alexander-skoblikov:


server/bundles/io.cloudbeaver.model/src/io/cloudbeaver/auth/SMAutoAssign.java
```diff
@@ -47,4 +50,13 @@ public List<String> getExternalTeamIds() {
     public void addExternalTeamId(String externalRoleId) {
         this.externalTeamIds.add(externalRoleId);
     }
+
+    @Nullable
+    public String getAuthRoleChangedReason() {
```
serge-rider: -> authRoleAssignReason

serge-rider (changes requested):

PR #3216, opened 2025-01-27
CB-5992 fix: limit maximum fraction digits in chart legend

SychevAndrey:
image

Do we want to use locale formatting maybe?


webapp/packages/core-utils/src/formatNumber.ts
```diff
@@ -8,7 +8,7 @@
 
 export function formatNumber(n: number, d: number) {
   if (n < 1000) {
-    return n.toString();
+    return n.toLocaleString(undefined, { maximumFractionDigits: 2 });
```
devnaumov, line 11: I think toFixed(2) will be enough

devnaumov (changes requested):

webapp/packages/core-utils/src/formatNumber.ts
SychevAndrey: toFixed will make 85.00 from 85, why do we need it?

Wroud (approved):

devnaumov (approved):

webapp/packages/core-utils/src/formatNumber.ts
devnaumov: In case of the toLocaleString we depend on the user local machine setup, it will create different behaviour for users. But I think its not that important for now

PR #3220, opened 2025-01-28
CB-6175 fixes eslint check on github

sergeyteleshev:


.github/workflows/frontend-lint.yml
```diff
@@ -39,7 +39,7 @@ jobs:
           git fetch origin "${{ github.base_ref }}" --depth=1
           FILES=$(git diff --name-only 'origin/${{ github.base_ref }}' ${{ github.sha }} -- . | sed 's|^webapp/||')
           if [ -n "$FILES" ]; then
-            yarn lint --pass-on-no-patterns --no-error-on-unmatched-pattern $FILES
+            yarn lint --parser-options --no-error-on-unmatched-pattern $FILES
```
Wroud, line 42: according documentation this option still exists and must work https://eslint.org/docs/latest/use/command-line-interface#options
but there is maybe a chance that it's now broken with legacy eslint config (not flat config)

Wroud: `--parser-options` itself absolutely different option and probably should not be used at all

Wroud:
fixed by eslint version update

PR #3221, opened 2025-01-28
CB-6128 add tooltips to tabs

devnaumov:
useService is used instead of useResource intentionally. It will decrease amount of calls to the resources. We dont need to force data to be presented (connection and project core entities).


webapp/packages/plugin-sql-editor-navigation-tab/src/SqlEditorTab.tsx
```diff
@@ -59,8 +61,30 @@ export const SqlEditorTab: TabHandlerTabComponent<ISqlEditorTabState> = observer
...
+    }
+
+    if (executionContext.defaultSchema) {
+      tooltip += `\n${translate('ui_schema')}: ${executionContext.defaultSchema}`;
+    }
+
+    if (project) {
+      tooltip += `\n${translate('ui_project')}: ${project.name}`;
+    }
+  }
```
Wroud: this part of logic is duplicated, can we move it to separate function-helper that will return tooltip string (localized)
something like `getObjectInfoTooltip(translate, connection?, catalog?, schema?, project?)`

webapp/packages/core-blocks/src/useObjectInfoTooltip.ts
```diff
@@ -0,0 +1,33 @@
...
+ * CloudBeaver - Cloud Database Manager
+ * Copyright (C) 2020-2024 DBeaver Corp and others
+ *
+ * Licensed under the Apache License, Version 2.0.
+ * you may not use this file except in compliance with the License.
+ */
+
+import { useTranslate } from './localization/useTranslate.js';
+
+export function useObjectInfoTooltip(connection?: string, catalog?: string, schema?: string, project?: string) {
```
sergeyteleshev, line 11: it is already almost a lot arguments. in such cases better to provide 1 argument as an object so it is easier to handle

sergeyteleshev (approved):

PR #3233, opened 2025-02-05
dbeaver/dbeaver-vscode#77 fix: eslint

Wroud:


webapp/packages/eslint-config/flat.js
```diff
@@ -0,0 +1,241 @@
...
+      'react/jsx-sort-props': [
+        'error',
+        {
+          callbacksLast: true,
+          shorthandLast: true,
+          ignoreCase: true,
+          noSortAlphabetically: true,
+          reservedFirst: true,
+        },
+      ],
```
sergeyteleshev, line 236: I also suggest to add this configs:
'no-console': ["warn", { allow: ["warn", "error"] }]
'no-unsafe-finally': 'error'
'require-await': 'error'
'@typescript-eslint/await-thenable': 'error'
'@typescript-eslint/explicit-module-boundary-types': 'warn'
'@typescript-eslint/no-unnecessary-condition': 'warn'
'react/no-array-index-key': 'warn'

sergeyteleshev (approved):

webapp/packages/eslint-config/flat.js
Wroud: `@typescript-eslint/no-unnecessary-condition` and `@typescript-eslint/await-thenable` can't be used because we don't use types information in eslint (because of performance issues on big projects)
other rules were added

PR #3235, opened 2025-02-05
CB-6119 await promise

devnaumov:


webapp/packages/core-root/src/PermissionsService.ts
```diff
@@ -43,7 +43,7 @@ export class PermissionsService {
   }
 
   async hasAsync(id: string): Promise<boolean> {
-    return this.permissions.hasAsync(id);
+    return await this.permissions.hasAsync(id);
   }
```
Wroud: there is no difference in runtime

SychevAndrey: agree, what is the motivation to do that? anyway we need to await hasAsync in the place of executing

Wroud (approved):

SychevAndrey (approved):

PR #3240, opened 2025-02-06
CB-6200 disconnect data source on changing config

yagudin10:


alexander-skoblikov (dismissed):

webapp/packages/core-connections/src/ConnectionInfoResource.ts
```diff
@@ -291,6 +293,16 @@ export class ConnectionInfoResource extends CachedMapResource<IConnectionInfoPar
     return this.get(key).every(connection => connection?.connected ?? false);
   }
 
+  async isConnectedAsync(key: IConnectionInfoParams): Promise<boolean>;
+  async isConnectedAsync(key: ResourceKeyList<IConnectionInfoParams>): Promise<boolean>;
+  async isConnectedAsync(key: ResourceKey<IConnectionInfoParams>): Promise<boolean>;
+  async isConnectedAsync(key: ResourceKey<IConnectionInfoParams>): Promise<boolean> {
+    key = ResourceKeyUtils.toList(this.aliases.transformToKey(key));
+    const connections = await this.refresh(key);
```
Wroud: Refresh is too expensive an operation, especially when you transform the key before executing it. In the worst case, it will make N requests to the backend for each connection and refresh related data.

Why is it not enough to have a synchronous check?

Wroud (changes requested):

webapp/packages/core-connections/src/ConnectionInfoResource.ts
sergeyteleshev: reverted

Wroud (approved):

PR #3241, opened 2025-02-06
CB-5023 Add vizualization for focus on switchers

sergeyteleshev:
https://github.com/user-attachments/assets/a8c74c76-7d42-479e-89db-90771ff1f562

### DIFF
--- a/webapp/packages/core-theming/src/styles/_switch.scss
+++ b/webapp/packages/core-theming/src/styles/_switch.scss
@@ -13,11 +13,6 @@
 
   .theme-switch {
     @extend .mdc-switch;
-
-    &:focus-within {
-      outline: 2px solid var(--theme-primary);
-      outline-offset: 12px;
-    }
   }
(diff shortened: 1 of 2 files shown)


webapp/packages/core-theming/src/styles/_switch.scss
```diff
@@ -13,6 +13,11 @@
 
   .theme-switch {
     @extend .mdc-switch;
+
+    &:focus-within {
+      outline: 2px solid var(--theme-primary);
+      outline-offset: 12px;
```
sergeyteleshev: There are no classes to implement from the material library, so I decided to go with this

https://m2.material.io/develop/web/components/switches

Wroud: There is probably a default style that is not applied correctly

Wroud (changes requested):

PR #3251, opened 2025-02-10
Feat/dbeaver/vscode#86/storybook integration

SychevAndrey:


common-react/@dbeaver/ui-kit/package.json
```diff
@@ -0,0 +1,32 @@
...
+    "@types/react": "^19",
+    "@types/react-dom": "^19",
+    "rimraf": "^6",
+    "typescript": "^5"
+  },
+  "dependencies": {
+    "@ariakit/react": "^0.4.15",
+    "@ladle/react": "^5.0.1",
+    "@tailwindcss/vite": "^4.0.3",
+    "@wroud/vite-plugin-asset-resolver": "^0.1.5",
```
Wroud: please keep only major version like `^5`
also check for dependencies that it's added correctly to `devDependencies` and `dependencies`

common-react/@dbeaver/ui-kit/src/Button/Button.tsx
```diff
@@ -0,0 +1,35 @@
...
+import './Button.css';
+
+export interface UiKitButtonProps extends Omit<ButtonProps, 'clickOnEnter' | 'clickOnSpace'> {
+  variant?: 'primary' | 'secondary' | 'danger';
+  size?: 'small' | 'medium' | 'large' | 'xlarge';
+  loading?: boolean;
+}
+
+export function Button({ className, variant = 'primary', size = 'medium', loading, children, onClick, ...props }: UiKitButtonProps) {
+  const classToApply = `dbv-ui-button dbv-ui-button--${variant} dbv-ui-button--${size}` + (className ? ` ${className}` : '');
```
Wroud: maybe it's better to use a prefix like `dbv-kit`

common-react/@dbeaver/ui-kit/src/Button/Button.tsx
```diff
@@ -0,0 +1,35 @@
...
+  return (
+    <AriaButton aria-label={loading ? 'Loading' : ''} onClick={loading ? () => null : onClick} className={classToApply} {...props}>
+      {loading && <span className="dbv-ui-button-loader" />}
+      {children}
+    </AriaButton>
+  );
+}
+
+export interface ButtonIconProps extends React.HTMLAttributes<HTMLSpanElement> {
+  position?: 'left' | 'right';
```
Wroud: this is composition element, doesn't position controlled by inserting component in the right position?:
```
<Button>
  some text
  <Button.Icon />
</Button>
```
```
<Button>
  <Button.Icon />
  some text
</Button>
```

common-react/@dbeaver/ui-kit/src/Button/button.css
```diff
@@ -0,0 +1,164 @@
+/*
+ * CloudBeaver - Cloud Database Manager
+ * Copyright (C) 2020-2025 DBeaver Corp and others
+ *
+ * Licensed under the Apache License, Version 2.0.
+ * you may not use this file except in compliance with the License.
+ */
+
+@layer components {
```
Wroud: file name is incorrect (must be **B**utton)

common-react/@dbeaver/ui-kit/src/Button/button.css
```diff
@@ -0,0 +1,164 @@
...
+    font-size: calc(var(--btn-font-size) * 1.125);
+    height: var(--control-height-large);
+  }
+
+  .dbv-ui-button--xlarge {
+    padding-inline: calc(var(--btn-padding-inline) * 1.5);
+    padding-block: calc(var(--btn-padding-block) * 1.5);
+    font-size: calc(var(--btn-font-size) * 1.25);
+    height: var(--control-height-xlarge);
+  }
```
Wroud: why can't we adjust css variable instead, like:
```
  .dbv-ui-button--large {
    --btn-padding-inline: calc(var(--btn-padding-inline) * 1.25);
    --btn-padding-block: calc(var(--btn-padding-block) * 1.25);
    --btn-font-size: calc(var(--btn-font-size) * 1.125);
    --btn-height: var(--control-height-large);
  }
  ```
(this code may not work as expected but shows an idea)

Comment (Wroud) [reply common-react/@dbeaver/ui-kit/src/Button/button.css]: It also may be good to use a global multiplier (so you need to change only one variable)

File: common-react/@dbeaver/ui-kit/src/Checkbox/Checkbox.css [hunk 6]
```diff
@@ -0,0 +1,20 @@
...
+ * Licensed under the Apache License, Version 2.0.
+ * you may not use this file except in compliance with the License.
+ */
+.checkbox {
+  accent-color: var(--checkbox-background);
+  width: 16px;
+  height: 16px;
+}
+
+.checkbox-label {
```
Wroud: probably we need to add prefixes to classes and variables here

common-react/@dbeaver/ui-kit/src/Input/Input.tsx
```diff
@@ -0,0 +1,28 @@
...
+}
+
+export const Input: React.FC<UIKitInputProps> = function Input({ error, warning, size, ...props }) {
+  const classNameToApply =
+    `dbv-ui-input dbv-ui-input--${size ?? 'medium'} ${error ? 'dbv-ui-input-validation--error' : ''} ${warning ? 'dbv-ui-input-validation--warning' : ''}`.trim();
+  return (
+    <div className="dbv-ui-input-wrapper">
+      <input className={classNameToApply} {...props} />
+      {error && <div className="dbv-ui-input-validation-message dbv-ui-input-validation-message--error">{error}</div>}
+      {warning && <div className="dbv-ui-input-validation-message dbv-ui-input-validation-message--warning">{warning}</div>}
```
Wroud: Let's not add validation primitives right now, but we need labels and descriptions (in composite components way)

webapp/packages/core-blocks/src/Button.tsx
```diff
@@ -25,7 +26,7 @@ export type ButtonProps = (React.ButtonHTMLAttributes<HTMLButtonElement | HTMLAn
   icon?: string;
   viewBox?: string;
   mod?: ButtonMod;
-  tag?: 'button' | 'a' | 'div';
+  render?: UiKitButtonProps['render'];
```
Wroud: i suggest renaming it later to avoid big diffs in this PR, right now we need to affect the public interface of base components minimally

common-react/@dbeaver/ui-kit/src/Button/Button.tsx
```diff
@@ -0,0 +1,35 @@
...
+ * CloudBeaver - Cloud Database Manager
+ * Copyright (C) 2020-2025 DBeaver Corp and others
+ *
+ * Licensed under the Apache License, Version 2.0.
+ * you may not use this file except in compliance with the License.
+ */
+import { Button as AriaButton, type ButtonProps } from '@ariakit/react';
+import './Button.css';
+
+export interface UiKitButtonProps extends Omit<ButtonProps, 'clickOnEnter' | 'clickOnSpace'> {
```
Wroud: just ButtonProps

webapp/packages/core-blocks/src/ToolsPanel/ToolsAction.module.css
```diff
@@ -17,7 +17,7 @@
 }
 .iconOrImage {
   display: block;
-  width: 16px;
+  width: 24px;
```
Wroud, line 20: strange changes

webapp/packages/core-ui/src/Tabs/Tab/TabActions.module.css
```diff
@@ -24,12 +24,11 @@
 
 .tabAction {
   position: relative;
-  height: 8px;
-  width: 8px;
+  height: 16px;
+  width: 16px;
   align-self: flex-start;
```
Wroud, line 29: strange changes

webapp/packages/core-ui/src/Tabs/Tab/TabIcon.module.css
```diff
@@ -1,21 +1,22 @@
...
+ * Copyright (C) 2020-2025 DBeaver Corp and others
  *
  * Licensed under the Apache License, Version 2.0.
  * you may not use this file except in compliance with the License.
  */
 
 .tabIcon {
   display: flex;
-  height: 24px;
+  height: 32px;
```
Wroud, line 11: strange changes

webapp/packages/core-ui/src/Tabs/Tab/TabIcon.module.css
```diff
@@ -1,21 +1,22 @@
...
   padding-left: 8px;
   justify-content: center;
   align-items: center;
 
   & .staticImage,
   & .iconOrImage,
   & img {
+    max-width: 16px;
```
Wroud, line 19: strange changes

Wroud (changes requested):

common-react/@dbeaver/ui-kit/src/Button/Button.tsx
SychevAndrey: yes, you are basically not controlling the position with this props, but styles (subtract padding to balance the button). You may define the prop or keep it empty. It's not ideal, let's discuss

common-react/@dbeaver/ui-kit/src/Button/Button.tsx
SychevAndrey: kit has less meaning for me than ui, plus ui is one symbol shorter means less html and css. So why should we change it? Can we vote in the team maybe?

common-react/@dbeaver/ui-kit/src/Button/button.css
SychevAndrey: git didn't want to change it) Fixed.

common-react/@dbeaver/ui-kit/src/Button/button.css
SychevAndrey: sorry didn't understand

webapp/packages/core-blocks/src/Button.tsx
SychevAndrey: it's the only way to use ariakit button properly, not so much components were touched by that changes

common-react/@dbeaver/ui-kit/src/Button/button.css
SychevAndrey: understood, working on that

common-react/@dbeaver/ui-kit/package.json
```diff
@@ -0,0 +1,32 @@
...
+    "@tailwindcss/vite": "^4",
+    "@types/react": "^19",
+    "@types/react-dom": "^19",
+    "@wroud/vite-plugin-asset-resolver": "^0",
+    "rimraf": "^6",
+    "typescript": "^5"
+  },
+  "dependencies": {
+    "@ariakit/react": "^0",
+    "@ladle/react": "^5",
```
Wroud: `ladle` probably is the dev dependency the same as `tailwindcss`

common-react/@dbeaver/ui-kit/src/Button/Button.css
```diff
@@ -0,0 +1,182 @@
...
+      @media (hover: hover) {
+        background-color: var(--btn-background-hover);
+      }
+    }
+  }
+
+  .dbv-kit-button--primary:active,
+  .dbv-kit-button--primary[data-active] {
+    background-color: var(--btn-background-active);
+  }
```
Wroud: Please prefer nesting or a consistent style of deep style declaration:
```
.dbv-kit-button--primary {
  &:hover { ... }
  &:active,
  &[data-active] { ... }
```

common-react/@dbeaver/ui-kit/src/Button/Button.css
```diff
@@ -0,0 +1,182 @@
...
+
+  .dbv-kit-button--secondary {
+    --btn-loader-color: color-mix(in srgb, var(--btn-secondary-foreground) 75%, var(--color-gray-600));
+    --btn-loader-background: color-mix(in srgb, var(--color-white) 75%, transparent);
+
+    padding-block: calc(var(--btn-padding-block) - 1px);
+
+    background-color: var(--btn-secondary-background);
+    border: 1px solid var(--btn-secondary-border-color);
+    color: var(--btn-secondary-foreground);
```
Wroud: The only difference between primary and secondary is colors. Why do we have different `padding-block` and full `border` styles instead of `border-color`

common-react/@dbeaver/ui-kit/src/Button/Button.css
```diff
@@ -0,0 +1,182 @@
...
+  }
+
+  @keyframes spin {
+    0% {
+      transform: rotate(0deg);
+    }
+    100% {
+      transform: rotate(360deg);
+    }
+  }
```
Wroud: Can we use tailwind animations? https://tailwindcss.com/docs/animation

common-react/@dbeaver/ui-kit/src/Checkbox/Checkbox.css
```diff
@@ -0,0 +1,20 @@
...
+  accent-color: var(--checkbox-background);
+  width: 16px;
+  height: 16px;
+}
+
+.dbv-kit-checkbox-label {
+  display: inline-flex;
+  align-items: center;
+  gap: var(--checkbox-gap);
+  margin-left: 0.5rem;
```
Wroud: Probably we want to use some global spacing variable (probably in pixels)

common-react/@dbeaver/ui-kit/src/Input/Input.tsx
```diff
@@ -0,0 +1,26 @@
...
+ * you may not use this file except in compliance with the License.
+ */
+
+import type { ComponentPropsWithRef } from 'react';
+import './Input.css';
+
+export interface InputProps extends Omit<ComponentPropsWithRef<'input'>, 'size'> {
+  size?: 'small' | 'medium' | 'large';
+  error?: string;
+  warning?: string;
```
Wroud: let's remove this at all

common-react/@dbeaver/ui-kit/src/index.css
```diff
@@ -0,0 +1,84 @@
...
+
+@import 'tailwindcss';
+
+@layer base {
+  :root {
+    /* General */
+    --font-size-base: 1rem;
+
+    /* Sizes */
+    --control-height-base: 2rem;
```
Wroud: I don't think using rem is a good idea for control sizes; usually, you will have some options like small, medium, and large and use them in controls, sometimes multiplying with some value. However, control sizes should not be dependent on the font size.

common-react/@dbeaver/ui-kit/src/index.css
```diff
@@ -0,0 +1,84 @@
...
+    --input-validation-warning-foreground: var(--color-warning-text);
+    --input-validation-warning-border: var(--color-warning-border);
+
+    /* Checkbox */
+    --checkbox-background: var(--color-primary-500);
+    --checkbox-foreground: var(--color-text);
+    --checkbox-border: var(--color-orange-500);
+    --checkbox-selectBackground: var(--color-orange-500);
+    --checkbox-selectBorder: var(--color-primary-500);
+    --checkbox-gap: --spacing(1);
```
Wroud: why not to move it to separate component related style files like you did with button?

webapp/packages/plugin-help/package.json
```diff
@@ -36,6 +36,7 @@
     "@cloudbeaver/plugin-sql-editor": "workspace:*",
     "@cloudbeaver/plugin-sql-editor-navigation-tab-script": "workspace:*",
     "@cloudbeaver/plugin-top-app-bar": "workspace:*",
+    "@dbeaver/ui-kit": "workspace:*",
```
Wroud: it seems that this dependency is not used in `plugin-help`

Wroud (changes requested):

Wroud:
we also need to add prefix for our ui-kit CSS variables

common-react/@dbeaver/ui-kit/src/index.css
SychevAndrey: please don't review input and checkbox parts because they will be done in separate tickets. This code will be refactored, all props will be moved to dedicated folders

common-react/@dbeaver/ui-kit/src/Checkbox/Checkbox.css
SychevAndrey: let's skip, component in WIP status and not in use, it might change drastically, we need to discuss the approach first.

common-react/@dbeaver/ui-kit/src/index.css
SychevAndrey: I get your point about controls not always needing to scale with font size, but I think in our case, it's the right choice.

Tailwind uses rem not only for text classes, but also for widths in classes like `w-sm` or `w-lg`.  If our controls are in `px`, they'll look totally wrong when users tweak their default font size.  Our buttons could end up tiny next to huge text, or vice versa – not a good look. We can see it now in cloudbeaver and I assume this is one of the things UI Kit should solve.

Screenshot 2025-02-18 at 14 45 59
Screenshot 2025-02-18 at 14 46 18
Screenshot 2025-02-18 at 14 46 27
Screenshot 2025-02-18 at 14 46 39

Users do change their font size, and we want our UI to adapt.  It's a big accessibility win. Since Tailwind is already using rem everywhere else, sticking with it for controls keeps things consistent and maintainable.  MDN Website is a good example of how controls should behave.
Screenshot 2025-02-18 at 15 15 50
Screenshot 2025-02-18 at 15 16 01

SychevAndrey:
> we also need to add prefix for our ui-kit CSS variables

It's not a must, for example: **github**, **mdn**, **radix**, and even **tailwind** don't use any prefix for their css props. However, **spectrum**, **jira**, and **ant design** use a prefix. I believe we need to discuss it.

sergeyteleshev (dismissed):

common-react/@dbeaver/ui-kit/src/index.css
Wroud: I disagree. For example, in the context of accessibility, we have some standards for click areas, and in this case, if you change the font to lover one, you may get controls that do not follow accessibility standards.

I've also checked this behavior in GitHub, and not any control that changes size depending on font size
this may work fine for text-orienting websites where you work with a lot of text (like magazines or newspapers) but not work at all for tools like IDE, where you have layouts and a place for controls

Wroud:
We need to use prefixes because we have different products and already have prefixes in the vs. code, so we can easily navigate to the correct part of the codebase; it's the same reason I suggested adding prefixes for class names. We can have multiple controls in our codebase (like one from UI Kit, and we may have one in each product as well)

common-react/@dbeaver/ui-kit/src/index.css
Wroud: Even on your screenshot from MDN, you can see that the  buttons and filter did not change size based on the font size

Wroud: Even more in the vs code, if you change font size, you will get controls that will be bigger or smaller than the same native controls from the vs code

common-react/@dbeaver/ui-kit/package.json
```diff
@@ -0,0 +1,33 @@
...
+  "version": "0.1.0",
+  "description": "",
+  "license": "Apache-2.0",
+  "exports": {
+    ".": "./dist/index.js"
+  },
+  "scripts": {
+    "build": "tsc -b",
+    "clean": "rimraf --glob dist",
+    "lint": "eslint ./src/ --ext .ts,.tsx"
```
Wroud: --ext will not work in eslint 9+

common-react/@dbeaver/ui-kit/src/Button/_base.css
```diff
@@ -0,0 +1,21 @@
...
+    --btn-padding-inline: --spacing(2.5);
+    --btn-gap: calc(var(--btn-padding-inline) / 4);
+    --btn-foreground: var(--color-white);
+    --btn-background: var(--color-primary-600);
+    --btn-background-hover: var(--color-primary-700);
+    --btn-background-active: var(--color-primary-800);
+    --btn-border-width: 1px;
+    --btn-border-color: transparent;
+    --btn-border-style: solid;
+    --btn-radius: var(--radius-md);
```
Wroud: i think we need some base variables on the theme layout for global things like:
1. controls radius - it may be used for buttons, inputs, text areas (examples)
2. maybe padding (like spacing between elements)

Wroud (changes requested):

common-react/@dbeaver/ui-kit/src/index.css
Wroud: ok let's use rem

common-react/@dbeaver/ui-kit/package.json
SychevAndrey: removed lint command for now. By the way, all our packages use exact same lint command, should we create a ticket to refactor it?

common-react/@dbeaver/ui-kit/src/Button/_base.css
SychevAndrey: I added vars for border, but I'm not sure about spacing right now. First of all, we can use tailwind directly, they have spacing units like p(padding) and m(margin) based on rem, they have container units like w- and h- based on rem.
Plus, creating the layout and grid units is more about product, not the UI-kit itself until it's very opinionated and have to be used as is which is not true in our case. We can create components / recommendation/ guidelines of how to use spacing **in the product**.

webapp/packages/plugin-administration/src/ConfigurationWizard/Welcome/WelcomePage.module.css
```diff
@@ -1,14 +1,8 @@
-/*
- * CloudBeaver - Cloud Database Manager
- * Copyright (C) 2020-2024 DBeaver Corp and others
- *
- * Licensed under the Apache License, Version 2.0.
- * you may not use this file except in compliance with the License.
- */
-
-.p {
+p {
```
Wroud: why has class now changed to a global style? this is CSS module we should not use global selectors like that

Wroud (changes requested):

webapp/packages/plugin-administration/src/ConfigurationWizard/Welcome/WelcomePage.module.css
SychevAndrey: fixed

Wroud (approved):

PR #3252, opened 2025-02-10
CB-6219 refactor: migrate to react 19

Wroud:


sergeyteleshev (dismissed):

webapp/packages/core-cli/configs/excludedFromVendor.js
```diff
@@ -1,7 +1,6 @@
 module.exports = [
   'react-data-grid',
   'react-markdown',
-  'react-virtualized',
```
SychevAndrey, line 4: How do we use this list and why it's safe to exclude `react-virtualized` from it?

SychevAndrey (dismissed):

webapp/packages/core-cli/configs/excludedFromVendor.js
SychevAndrey: Ok, I've seen in EE repo we removed that lib

Wroud: We don't use it right now. We used it for Webpack to exclude some libraries from the "vendor" bundle to let them be bundled with components that rely on them, but right now, we have not implemented proper chunk splitting, so this list is not in use, but we may use it later

sergeyteleshev (approved):

PR #3254, opened 2025-02-10
CB-6135 Database objects do not open after disconnecting

sergeyteleshev:
I tested this well, and I think this logic with `freeze` props is not working in this case. According to @devnaumov, this code should cover the case when the children node is loaded before the parent node is loaded during refresh. However, this case may not be replicated
https://github.com/dbeaver/cloudbeaver/pull/3129 - the cause PR

See the attached video for 2 cases: fixing the issue and showing that refresh is working just fine

https://github.com/user-attachments/assets/ea3812d6-bd2d-4d04-83de-20add7cc58af


Wroud:
probably we need to test it with different nav tree refresh cases

Wroud (approved):

webapp/packages/plugin-object-viewer/src/ObjectViewerPanel/ObjectViewerPanel.tsx
```diff
@@ -55,7 +53,6 @@ export const ObjectViewerPanel: TabHandlerPanelComponent<IObjectViewerTabState>
         tab.handlerState.tabTitle = data.name;
       });
     },
-    freeze: navTreeResource.isOutdated(resourceKeyList(navNodeInfoResource.getParents(objectId))),
```
devnaumov, line 58: it will bring bug back

PR #3263, opened 2025-02-12
CB-6171 show alert instead of icon

devnaumov:


webapp/packages/core-blocks/src/Alert.tsx
```diff
@@ -0,0 +1,44 @@
...
+  message: string;
+  title?: string;
+}
+
+export const Alert = observer<Props>(function Alert({ title, message }) {
+  const translate = useTranslate();
+  const styles = useS(classes);
+
+  return (
+    <div className={s(styles, { alert: true })} role="alert">
```
sergeyteleshev, line 36: probably you can use `<Container/>` here

webapp/packages/plugin-connection-custom/src/DriverSelector/Driver.tsx
```diff
@@ -35,7 +35,7 @@ export const Driver = observer<Props>(function Driver({ driver, onSelect }) {
       <ListItemIcon className={s(styles, { icon: true })}>
         <StaticImage icon={driver.icon} className={s(styles, { staticImage: true })} />
         {!driver.driverInstalled && (
-          <div className={s(styles, { indicator: true })} title={translate('plugin_connection_custom_drivers_driver_not_installed')}>
+          <div className={s(styles, { indicator: true })} title={translate('connections_connection_driver_not_installed')}>
```
sergeyteleshev: the idea of reusing one translation across several places where it is needed seems great from optimization perspective
but in fact an interfaces are dynamic things and can change all the time. so somewhen this would put us in situation when in one place translation has a perfect fit, and in another one is not. plus if someone changes translation in one place, it also changes in another one and we meet the same issue - translation is not fit in this place anymore

so in my experience it is better to go with this: 1 place - 1 translation (unless it is something general like `ui_delete` and etc.). since it is our base practice right now I suggest to keep with it

.vscode/settings.json
```diff
@@ -37,7 +37,7 @@
     "editor.defaultFormatter": "esbenp.prettier-vscode"
   },
   "[typescriptreact]": {
-    "editor.defaultFormatter": "esbenp.prettier-vscode"
+    "editor.defaultFormatter": "vscode.typescript-language-features"
```
Wroud: something strange, please revert

webapp/packages/core-connections/src/locales/zh.ts
```diff
@@ -94,4 +94,5 @@ export default [
   ['core_connections_settings_disable', 'Disable'],
   ['core_connections_settings_disable_description', 'Disable the ability to create new connections'],
   ['connections_templates_deprecated_message', 'Template connections are deprecated and will be removed in future releases'],
+  ['connections_connection_driver_not_installed', 'Driver is not installed'],
```
Wroud: please follow a naming convention for localization tokens

webapp/packages/plugin-connection-custom/src/DriverSelector/Driver.tsx
Wroud: i think here it's okay, but please fix linting warning on line 30 in this file, it seems that it may create a bug (remove useCallback or specify all required deps)

Wroud (changes requested):

webapp/packages/core-blocks/src/Alert.tsx
devnaumov: It will apply extra styles to children that we dont want

Wroud (approved):

PR #3266, opened 2025-02-13
CB-3771 Incorrect autocomplete of the Host field while connecting to the local db in the local env (via Search)

sergeyteleshev:


SychevAndrey (dismissed):

webapp/packages/plugin-connections/src/ConnectionForm/Options/ConnectionOptionsTabService.ts
```diff
@@ -160,12 +162,14 @@ export class ConnectionOptionsTabService extends Bootstrap {
     }
 
     const driver = await this.dbDriverResource.load(state.config.driverId, ['includeProviderProperties']);
+    const driverDatabases = await this.connectionInfoResource.searchDatabases(driver.defaultDatabase || 'localhost');
```
Wroud: you can't run this function that way cuz it will:
1. make many requests to provided hosts (that may take long time)
2. will not be cached (and this is expected)

please check `select` in the webapp/packages/plugin-connection-search/src/Search/ConnectionSearchService.ts`
we use `setConfig` to provide settings from selected driver, but it seems like this settings is not honored or overwrited

Wroud (changes requested):

Wroud (approved):

PR #3267, opened 2025-02-13
CB-6249 No icons in the app for windows builds

sergeyteleshev:


Wroud (dismissed):

SychevAndrey (dismissed):

devnaumov (dismissed):

webapp/packages/product-base/src/copy-assets/copyAssetsPlugin.ts
```diff
@@ -16,8 +16,7 @@ import { getAssets } from './getAssets.js';
 
 export function copyAssetsPlugin(): Plugin {
   // let rootPath = fileURLToPath(import.meta.url);
-
-  return DynamicPublicDirectory(getAssets(fs.join(process.cwd(), 'package.json')).map(v => `${v}/**`)) as any;
+  return DynamicPublicDirectory(getAssets(fs.join(process.cwd(), 'package.json')).map(v => fs.normalize(`${v}/**`))) as any;
```
sergeyteleshev: https://nodejs.org/api/path.html#pathnormalizepath

Wroud (dismissed):

webapp/packages/product-base/src/copy-assets/copyAssetsPlugin.ts
```diff
@@ -16,8 +16,15 @@ import { getAssets } from './getAssets.js';
 
 export function copyAssetsPlugin(): Plugin {
   // let rootPath = fileURLToPath(import.meta.url);
+  const assets = getAssets(fs.join(process.cwd(), 'package.json'));
+  const normalizedAssets = assets.map(asset =>
+    fs
+      .normalize(`${asset}/**`)
+      // the DynamicPublicDirectory plugin expects paths to be in posix format (to include images and fonts for windows)
```
sergeyteleshev, line 23: Well, our library has some problems with the paths. Here are some notes from the author that pushed me to the idea of replacing all double backslashes with a single slash

https://github.com/nguyenbatranvan/vite-multiple-assets/blob/0a2216b5341d42622adfb015956d0d89385c5bf2/DEVELOPMENT.md

Wroud: It shouldn't be an issue with the library, but this is a unique behavior of rollup (used in vite as bundler) where all paths use posix style.

But this definitely should be handled by the plugin itself.

Wroud (approved):

PR #3270, opened 2025-02-17
CB-6208 adjust z index

devnaumov:


Wroud (changes requested):
I think that it's better to decrease the z-index in the menus / selectors / combobox instead

Wroud (approved):

PR #3278, opened 2025-02-18
CB-6264 Filter field hint list is shown when filter is added manually and focus is out of the field

sergeyteleshev:
…suggestions

### DIFF
--- a/webapp/packages/core-blocks/src/FormControls/useInputAutocomplete.ts
+++ b/webapp/packages/core-blocks/src/FormControls/useInputAutocomplete.ts
@@ -150,6 +150,8 @@ export const useInputAutocomplete = (
       case 'ArrowDown':
       case 'ArrowUp':
         state.menuRef.current?.first();
+      case 'Tab':
+        state.resetState();
         break;
       default:
         break;
--- a/webapp/packages/core-blocks/src/FormControls/useInputAutocomplete.ts
+++ b/webapp/packages/core-blocks/src/FormControls/useInputAutocomplete.ts
@@ -83,6 +83,7 @@ export const useInputAutocomplete = (
         }
 
         this.isFound = true;
+        this.resetState();
       },
       get currentWord(): string {
         const cursorPosition = this.selectionStart;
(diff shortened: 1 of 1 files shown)


webapp/packages/core-blocks/src/FormControls/useInputAutocomplete.ts
```diff
@@ -142,19 +142,31 @@ export const useInputAutocomplete = (
...
+
+    state.selectionStart = null;
+  }
+
   useEffect(() => {
     const input = state.inputRef.current!;
 
     input.addEventListener('input', handleInput);
     input.addEventListener('keydown', handleKeyDown);
+    document.addEventListener('click', handleClickOutsideOfInput);
```
Wroud: Is it better to keep only keyboard events? Cuz seems that the issue is because of the `input` event

devnaumov: I think the issue is related to the input event too

sergeyteleshev: It is not like this. The input handler works only when typing on a keyboard. And during the case, it is not involved at all
The main issue is that input doesn't know whether it is focused. We can identify that using `positionStart`. It was not reset during user actions, so `currentWord` existed but it shouldn't, cause it should exist only when a user types something using keyboard
Frankly speaking, `positionStart` just leaked. That's it

webapp/packages/core-blocks/src/FormControls/useInputAutocomplete.ts
```diff
@@ -83,6 +83,7 @@ export const useInputAutocomplete = (
         }
 
         this.isFound = true;
+        this.resetState();
```
Wroud: Probably, we need to use `finally` cuz we have `early return`.

Wroud (changes requested):

Wroud (dismissed):

devnaumov (dismissed):

Wroud (approved):

PR #3288, opened 2025-02-24
CB-4067 Migrate Connection form to form template

sergeyteleshev:


webapp/packages/core-connections/src/locales/de.ts
```diff
@@ -62,4 +62,6 @@ export default [
   ['connections_templates_deprecated_message', 'Template connections are deprecated and will be removed in future releases'],
   ['core_connections_connection_driver_not_installed', 'Driver is not installed'],
   ['core_connections_connection_temporary', 'Temporary connection'],
+  ['connections_connection_create_success', 'Connection is created'],
+  ['connections_connection_update_success', 'Connection is updated'],
```
Wroud: please follow naming pattern: `{plugin_name}_...`
in this case is: `core_connections_connection_create_success`

webapp/packages/plugin-connections-administration/src/ConnectionForm/ConnectionAccess/ConnectionFormAccessPart.ts
```diff
@@ -0,0 +1,100 @@
...
+import { createConnectionParam, type ConnectionInfoResource } from '@cloudbeaver/core-connections';
+import { action, makeObservable } from 'mobx';
+import type { IConnectionFormState } from '@cloudbeaver/plugin-connections';
+import { getSubjectDifferences } from '@cloudbeaver/core-utils';
+
+const getDefaultState = () =>
+  ({
+    grantedSubjects: [],
+    editing: false,
+  }) as IConnectionFormAccessState;
```
Wroud: This is kinda hard to read, maybe:
```ts
function getDefaultState(): IConnectionFormAccessState {
  return {
    grantedSubjects: [],
    editing: false,
  };
}
```

webapp/packages/plugin-connections-administration/src/ConnectionForm/ConnectionAccess/ConnectionFormAccessPart.ts
```diff
@@ -0,0 +1,100 @@
...
+    const connectionId = this.formState.state.config.connectionId;
+
+    if (this.formState.state.submitType === 'test' || !data.state.projectId || !status.saved || !connectionId || !this.loaded) {
+      return;
+    }
+
+    const key = createConnectionParam(data.state.projectId, connectionId);
+
+    const currentGrantedSubjects = await this.connectionInfoResource.loadAccessSubjects(key);
+    const currentGrantedSubjectIds = currentGrantedSubjects.map(subject => subject.subjectId);
```
Wroud: You don't need to call this function because the initial state will be updated by loader fn before saving (the form part calls it)
so use `this.initialState.grantedSubjects `

webapp/packages/plugin-connections-administration/src/ConnectionForm/ConnectionAccess/ConnectionFormAccessPart.ts
```diff
@@ -0,0 +1,100 @@
...
+
+    const { subjectsToRevoke, subjectsToGrant } = getSubjectDifferences(currentGrantedSubjectIds, this.state.grantedSubjects);
+
+    if (subjectsToRevoke.length === 0 && subjectsToGrant.length === 0) {
+      return;
+    }
```
Wroud: This logic is duplicated by the code below. I don't think that we need this code

webapp/packages/plugin-connections-administration/src/ConnectionForm/ConnectionAccess/ConnectionFormAccessPart.ts
```diff
@@ -0,0 +1,100 @@
...
+    if (subjectsToGrant.length > 0) {
+      promises.push(this.connectionInfoResource.addConnectionsAccess(key, subjectsToGrant));
+    }
+
+    await Promise.all(promises);
+  }
+
+  edit() {
+    this.state.editing = !this.state.editing;
+  }
```
Wroud: Is this a view state?

Wroud (changes requested):

webapp/packages/plugin-connections/src/ConnectionForm/ConnectionFormState.ts
```diff
@@ -1,342 +1,61 @@
...
-      this.service.formSubmittingTask,
-    );
-  }
-
-  async checkFormState(): Promise<IFormStateInfo | null> {
-    await this.loadConnectionInfo();
-    return this.stateInfo;
+    resource.onItemUpdate.removeHandler(this.syncInfo.bind(this));
+    projectInfoResource.onDataUpdate.removeHandler(this.syncProject.bind(this));
+    projectsService.onActiveProjectChange.removeHandler(this.syncProject.bind(this));
```
Wroud: Why is this code in the constructor? How does it work?

webapp/packages/plugin-connections/src/ConnectionForm/ConnectionFormState.ts
```diff
@@ -1,342 +1,61 @@
...
+      !this.state.config.connectionId ||
+      this.state.projectId === null ||
+      !resource.isIntersect(key, createConnectionParam(this.state.projectId, this.state.config.connectionId))
     ) {
       return;
     }
 
-    this.loadConnectionInfo();
+    await this.reload();
   }
```
Wroud: This code should not be here. If you use some data from the connection info resource in any part, you should add tracking logic to the form part's `isOutdated` or `isLoaded`. The synchronization mechanism uses these methods

webapp/packages/plugin-connections/src/ConnectionForm/ConnectionFormState.ts
```diff
@@ -1,342 +1,61 @@
...
+    const projectInfoResource = this.serviceProvider.getService(ProjectInfoResource);
+    const projectsService = this.serviceProvider.getService(ProjectsService);
 
-  private async loadInfo(data: IConnectionFormState, contexts: IExecutionContextProvider<IConnectionFormState>) {
-    if (!data.config.connectionId || data.projectId === null) {
-      return;
+    const project = projectInfoResource.get(this.state.projectId);
+    if (!project?.canEditDataSources || !projectsService.activeProjects.includes(project)) {
+      await this.dispose();
     }
```
Wroud: This brings inversion of control. Probably, this logic must be handled by some service or code that creates a connection form

webapp/packages/plugin-connections/src/ConnectionForm/DriverProperties/ConnectionFormDriverPropertiesPart.ts
```diff
@@ -0,0 +1,72 @@
...
+
+  protected override async loader(): Promise<void> {
+    if (!this.formState.state.config.connectionId || !this.formState.state.projectId) {
+      this.setInitialState(getDefaultState());
+      return;
+    }
+
+    const connection = this.connectionInfoResource.get(
+      createConnectionParam(this.formState.state.projectId, this.formState.state.config.connectionId),
+    );
```
Wroud: this form part uses `connectionInfoResource` you should add `isOutdated`method to ensure that part's state synchronized

webapp/packages/plugin-connections/src/ConnectionForm/DriverProperties/ConnectionFormDriverPropertiesPart.ts
```diff
@@ -0,0 +1,72 @@
...
+
+        trimmedProperties[key] = typeof value === 'string' ? value.trim() : value;
+      }
+
+      this.state = trimmedProperties;
+      optionsPart.state.properties = {
+        ...optionsPart.state.properties,
+        ...this.state,
+      };
+    }
```
Wroud: Hm, should we trim only changed properties? If so, we can compare it with the initial state and trim only changed values

webapp/packages/plugin-connections/src/ConnectionForm/DriverProperties/ConnectionFormDriverPropertiesPart.ts
```diff
@@ -0,0 +1,72 @@
...
+    if (!this.formState.state.config.connectionId || !this.formState.state.projectId) {
+      this.setInitialState(getDefaultState());
+      return;
+    }
+
+    const connection = this.connectionInfoResource.get(
+      createConnectionParam(this.formState.state.projectId, this.formState.state.config.connectionId),
+    );
+
+    this.setInitialState(connection?.properties ?? getDefaultState());
```
Wroud: i will recommend to create a copy of the properties object to avoid direct mutations

webapp/packages/plugin-connections/src/ConnectionForm/IConnectionFormState.ts
```diff
@@ -0,0 +1,26 @@
...
+ */
+
+import { schema } from '@cloudbeaver/core-utils';
+import { CONNECTION_CONFIG_SCHEMA } from './Options/IConnectionConfig.js';
+import type { IFormProps } from '@cloudbeaver/core-ui';
+
+export const CONNECTION_FORM_STATE_SCHEMA = schema
+  .object({
+    projectId: schema.string(),
+    config: CONNECTION_CONFIG_SCHEMA,
```
Wroud: I will not recommend using this property (`config`) here, instead, when you needed you can always access the state of a part to change the initial state or state of it

webapp/packages/plugin-connections/src/ConnectionForm/Options/ConnectionFormOpitonsPart.ts
```diff
@@ -0,0 +1,462 @@
...
+      return;
+    }
+
+    if (!this.formState.state.config.connectionId || !this.formState.state.projectId) {
+      console.error('Connection id and project id should be defined');
+      return;
+    }
+
+    const [info] = await Promise.all([
+      this.connectionInfoResource.load(createConnectionParam(this.formState.state.projectId, this.formState.state.config.connectionId), [
```
Wroud: please add synchronization for resources via `isOutdated`

webapp/packages/plugin-connections/src/ConnectionForm/Options/useOptions.ts
```diff
@@ -8,32 +8,37 @@
...
 import { getDefaultConfigurationType } from './getDefaultConfigurationType.js';
+import type { IFormState } from '@cloudbeaver/core-ui';
+import type { IConnectionFormState } from '../IConnectionFormState.js';
+
+export function useOptions(formState: IFormState<IConnectionFormState>, config: ConnectionConfig) {
+  const mode = formState.mode;
+  const connectionInfoService = useService(ConnectionInfoResource);
+  const info = connectionInfoService.get(createConnectionParam(formState.state.projectId, formState.state.config.connectionId!));
 
-export function useOptions(state: IConnectionFormState) {
```
Wroud, line 19: this hook probably completely may be replaced by form's API

Wroud (changes requested):

webapp/packages/plugin-connections/src/ConnectionForm/OriginInfo/ConnectionFormOriginInfoFormPart.ts
```diff
@@ -0,0 +1,105 @@
...
+    const state = defaultStateGetter();
+
+    if (!this.formState.state.config.connectionId || !this.formState.state.projectId || !this.formState.state.config.driverId) {
+      this.setInitialState(state);
+      return;
+    }
+
+    const info = this.connectionInfoResource.get(createConnectionParam(this.formState.state.projectId, this.formState.state.config.connectionId));
+    const driver = await this.dbDriverResource.load(this.formState.state.config.driverId);
+    await this.databaseAuthModelsResource.load(this.formState.state.config.authModelId ?? info?.authModel ?? driver?.defaultAuthModel ?? null);
```
Wroud: probably we also need to load user info here

webapp/packages/plugin-connections/src/ConnectionForm/OriginInfo/ConnectionFormOriginInfoFormPart.ts
```diff
@@ -0,0 +1,105 @@
...
+import { AUTH_PROVIDER_LOCAL_ID, type UserInfoResource } from '@cloudbeaver/core-authentication';
+import { computed, makeObservable } from 'mobx';
+
+const defaultStateGetter = () => ({}) as IConnectionFormOriginInfoState;
+
+@injectable()
+export class ConnectionFormOriginInfoFormPart extends FormPart<IConnectionFormOriginInfoState, IConnectionFormState> {
+  constructor(
+    formState: IFormState<IConnectionFormState>,
+    private readonly connectionInfoOriginDetailsResource: ConnectionInfoOriginDetailsResource,
```
Wroud, line 31: you need to add connection form options part (and track it's `isOutdated` also

webapp/packages/plugin-connections/src/ConnectionForm/OriginInfo/OriginInfoTab.tsx
```diff
@@ -10,13 +10,18 @@ import { observer } from 'mobx-react-lite';
...
 import { Tab, type TabContainerTabComponent, TabTitle } from '@cloudbeaver/core-ui';
 
-import type { IConnectionFormProps } from '../IConnectionFormProps.js';
+import type { IConnectionFormProps } from '../IConnectionFormState.js';
+import { ConnectionInfoOriginResource, createConnectionParam } from '@cloudbeaver/core-connections';
+import { useService } from '@cloudbeaver/core-di';
+
+export const OriginInfoTab: TabContainerTabComponent<IConnectionFormProps> = observer(function OriginInfoTab({ formState, ...rest }) {
+  const connectionInfoOriginService = useService(ConnectionInfoOriginResource);
+  const info = connectionInfoOriginService.get(createConnectionParam(formState.state.projectId, formState.state.config.connectionId!));
```
Wroud: it's better to use `useResource` in such cases

webapp/packages/plugin-connections-administration/src/ConnectionForm/ConnectionAccess/ConnectionAccess.tsx
```diff
@@ -23,58 +23,64 @@ import {
...
 
   if (!selected) {
     return null;
   }
 
-  const loading = users.isLoading() || teams.isLoading() || state.state.loading;
-  const cloud = formState.info && formState.originInfo?.origin ? isCloudConnection(formState.originInfo.origin) : false;
-  const disabled = loading || !state.state.loaded || formState.disabled || cloud;
+  const connectionInfoService = useService(ConnectionInfoResource);
+  const originInfoService = useService(ConnectionInfoOriginResource);
```
devnaumov: Please dont call hooks conditionally

webapp/packages/plugin-connections/src/ConnectionForm/SSH/SSHPanel.tsx
```diff
@@ -7,18 +7,21 @@
...
 
-import type { IConnectionFormProps } from '../IConnectionFormProps.js';
 import { SSH } from './SSH.js';
+import { useAutoLoad } from '@cloudbeaver/core-blocks';
+import { getConnectionFormSSHPart } from './getConnectionFormSSHPart.js';
+import type { IConnectionFormProps } from '../IConnectionFormState.js';
 
 export const SSHPanel: TabContainerPanelComponent<IConnectionFormProps> = observer(function SSHPanel(props) {
-  const state = props.state.config.networkHandlersConfig?.find(state => state.id === SSH_TUNNEL_ID);
+  const SSHPart = getConnectionFormSSHPart(props.formState);
```
Wroud: sshPart

webapp/packages/plugin-connections/src/ConnectionForm/SSL/SSLPanel.tsx
```diff
@@ -7,31 +7,28 @@
...
+import { getConnectionFormSSLPart } from './getConnectionFormSSLPart.js';
+import type { IConnectionFormProps } from '../IConnectionFormState.js';
 
 export const SSLPanel: TabContainerTabComponent<IConnectionFormProps> = observer(function SSLPanel(props) {
   const networkHandlerResource = useResource(SSLPanel, NetworkHandlerResource, CachedMapAllKey);
-  const dbDriverResource = useResource(SSLPanel, DBDriverResource, props.state.config.driverId ?? null);
+  const dbDriverResource = useResource(SSLPanel, DBDriverResource, props.formState.state.config.driverId ?? null);
 
   const handler = getSSLDriverHandler(networkHandlerResource.resource.values, dbDriverResource.data?.applicableNetworkHandlers ?? []);
+  const SSLPart = getConnectionFormSSLPart(props.formState);
```
Wroud: sslPart

webapp/packages/plugin-connections-administration/src/ConnectionForm/ConnectionAccess/ConnectionAccess.tsx
```diff
@@ -23,58 +23,64 @@ import {
...
+  const connectionInfo = connectionInfoService.get(createConnectionParam(formState.state.projectId, formState.state.config.connectionId!));
+  const originInfo = originInfoService.get(createConnectionParam(formState.state.projectId, formState.state.config.connectionId!));
```
devnaumov: Why dont we use useResource hook here if we need this data?

webapp/packages/plugin-connections/src/ConnectionForm/useConnectionFormState.ts
```diff
@@ -1,38 +1,49 @@
...
+  if (ref.current?.state.config.connectionId !== params.connectionId || ref.current?.state.projectId !== params.projectId) {
+    ref.current = new ConnectionFormState(serviceProvider, service, {
+      ...EMPTY_CONNECTION_INFO_PARAMS,
+      projectId: params.projectId,
+      config: {
+        ...EMPTY_CONNECTION_INFO_PARAMS.config,
+        connectionId: params.connectionId,
+      },
+    });
+    configure?.(ref.current);
```
Wroud: you need to dispose previous ref if any

Wroud (changes requested):

webapp/packages/plugin-connections/src/ConnectionForm/ConnectionForm.tsx
```diff
@@ -6,93 +6,103 @@
...
-      <TabsState container={service.tabsContainer} localState={state.partsState} state={state} onCancel={onCancel}>
+      <TabsState container={service.parts} localState={formState.parts} formState={formState}>
         <div className={s(styles, { box: true }, className)}>
           <div className={s(styles, { connectionTopBar: true })}>
             <div className={s(styles, { connectionTopBarTabs: true })}>
               <div className={s(styles, { connectionStatusMessage: true })}>
-                <StatusMessage type={ENotificationType.Info} message={state.statusMessage} />
+                <StatusMessage
+                  multipleRows
+                  type={exception ? ENotificationType.Error : ENotificationType.Info}
```
devnaumov, line 92: Please fix eslint rules

webapp/packages/plugin-connections/src/ConnectionForm/ConnectionFormState.ts
```diff
@@ -1,342 +1,61 @@
...
-import {
-  ConnectionInfoOriginResource,
-  ConnectionInfoResource,
-  createConnectionParam,
-  type DatabaseConnection,
-  type IConnectionInfoParams,
-} from '@cloudbeaver/core-connections';
-import { Executor, type IExecutionContextProvider, type IExecutor } from '@cloudbeaver/core-executor';
-import type { ProjectInfoResource, ProjectsService } from '@cloudbeaver/core-projects';
+import { ConnectionFormService } from './ConnectionFormService.js';
```
devnaumov, line 11: Please place relative imports after workspace imports

webapp/packages/plugin-connections/src/ConnectionForm/DriverProperties/ConnectionFormDriverPropertiesPart.ts
devnaumov: Why do we need to trim key here?

webapp/packages/plugin-connections/src/ConnectionForm/Options/IConnectionNetworkHanler.ts
```diff
@@ -0,0 +1,23 @@
...
+ *
+ * Licensed under the Apache License, Version 2.0.
+ * you may not use this file except in compliance with the License.
+ */
+import { NetworkHandlerAuthType } from '@cloudbeaver/core-sdk';
+import { schema } from '@cloudbeaver/core-utils';
+
+export const CONNECTION_NETWORK_HANDLER_SCHEMA = schema.object({
+  id: schema.string(),
+  authType: schema.enum([NetworkHandlerAuthType.Agent, NetworkHandlerAuthType.Password, NetworkHandlerAuthType.PublicKey]).optional(),
```
devnaumov: Probably could use nativeEnum(NetworkHandlerAuthType)

webapp/packages/plugin-connections/src/ConnectionForm/SSH/ConnectionFormSSHPart.ts
```diff
@@ -0,0 +1,176 @@
...
+
+    if (!this.isChanged) {
+      return;
+    }
+
+    if (!this.state.enabled) {
+      return;
+    }
+
+    if (this.isChanged) {
```
devnaumov: Seems like an extra check, you checked if part was changed before

devnaumov (dismissed):

webapp/packages/core-ui/src/Form/FormState.ts
```diff
@@ -65,10 +65,13 @@ export class FormState<TState> implements IFormState<TState> {
...
     this.formStateTask = new Executor<TState>(state, () => true);
-    this.formStateTask.addCollection(service.onState).addPostHandler(this.updateFormState.bind(this));
 
     this.loadedTask = new Executor(this as IFormState<TState>, () => true);
-    this.loadedTask.addCollection(service.onLoaded).next(this.formStateTask).addPostHandler(this.onLoadedHandler.bind(this));
+    this.loadedTask
+      .addCollection(service.onLoaded)
+      .next(this.formStateTask)
+      .addPostHandler(this.onLoadedHandler.bind(this))
+      .addPostHandler(this.updateFormState.bind(this));
```
Wroud: These changes may be incorrect, `updateFormState` now will not be triggered when `formStateTask` executed
I will suggest to revert this changes

webapp/packages/core-ui/src/Form/FormState.ts
```diff
@@ -219,6 +229,6 @@ export class FormState<TState> implements IFormState<TState> {
       await part.dispose();
     }
 
-    await this.disposeTask.execute(this);
+    this.disposeTask.execute(this);
```
Wroud: Why are we not awaiting it anymore?

webapp/packages/plugin-connection-search/src/Search/ConnectionSearchService.ts
```diff
@@ -156,33 +151,28 @@ export class ConnectionSearchService {
...
-    if (!this.formState) {
-      this.formState = new ConnectionFormState(
-        this.projectsService,
-        this.projectInfoResource,
-        this.connectionFormService,
-        this.connectionInfoResource,
-        this.connectionInfoOriginResource,
-      );
+    this.formState?.dispose();
+    this.formState?.disposeTask.removeHandler(this.goBack.bind(this));
```
Wroud: this will not work `bind` will return new function every time

webapp/packages/plugin-connection-search/src/Search/ConnectionSearchService.ts
```diff
@@ -156,33 +151,28 @@ export class ConnectionSearchService {
...
+      requiredNetworkHandlersIds: [],
+    });
 
-      this.formState.closeTask.addHandler(this.goBack.bind(this));
-    }
+    runInAction(() => {
+      this.optionsPart!.state.host = database.host;
+      this.optionsPart!.state.port = String(database.port);
+      this.optionsPart!.state.driverId = database.defaultDriver;
+    });
```
Wroud: just mark `select` method as mobx action

webapp/packages/plugin-connection-search/src/Search/ConnectionSearchService.ts
```diff
@@ -156,33 +151,28 @@ export class ConnectionSearchService {
...
-      })
-      .setAvailableDrivers(database.possibleDrivers);
-
-    this.formState.load();
+    this.formState.disposeTask.addHandler(this.goBack.bind(this));
   }
 
   private clearFormState() {
     this.formState?.dispose();
+    this.formState?.disposeTask.removeHandler(this.goBack.bind(this));
```
Wroud: This will not work as expected, even more as we plan to remove the form state object at all; we probably don't need to remove handlers (GC should handle it)

webapp/packages/plugin-connections-administration/src/Administration/Connections/CreateConnectionService.ts
```diff
@@ -109,26 +104,31 @@ export class CreateConnectionService {
...
+    this.data = new ConnectionFormState(this.serviceProvider, this.connectionFormService, {
+      projectId,
+      availableDrivers,
+      submitType: null,
+      type: 'admin',
+      requiredNetworkHandlersIds: [],
+    });
 
-    this.data.load();
+    runInAction(() => {
```
Wroud: just mark setConnectionTemplate as an mobx action

webapp/packages/plugin-connections-administration/src/Administration/Connections/CreateConnectionService.ts
```diff
@@ -109,26 +104,31 @@ export class CreateConnectionService {
...
+      this.optionsPart!.state = {
+        ...this.optionsPart!.state,
+        ...config,
+      };
```
Wroud: in mobx context `Object.assign(this.optionsPart!.state, config)` is more preferable

Wroud (changes requested):

webapp/packages/plugin-connections-administration/src/ConnectionForm/ConnectionAccess/ConnectionAccess.tsx
```diff
@@ -19,62 +20,95 @@ import {
...
+    {
+      editing: observable.ref,
+      toggleEdit: action.bound,
+    },
+    false,
+  );
 
   const { selected } = useTab(tabId);
+  const accessPart = getConnectionFormAccessPart(formState);
+  const [initialFormMode] = useState<FormMode>(formState.mode);
```
Wroud, line 57: I think that you can use `formState.mode` directly, it will reflect actual state of form

webapp/packages/plugin-connections/src/ConnectionForm/ConnectionForm.tsx
```diff
@@ -1,98 +1,116 @@
...
+        const state = contexts.getContext(formStatusContext);
 
-        if (validation.valid && state.saved && data.submitType === 'submit') {
-          props.onSave(config);
+        if (validation.valid && state.saved && data.state.submitType === 'submit') {
+          props.onSave(optionsPart.state);
         }
       },
     ],
   });
```
Wroud: please check users form implementation, maybe we don't need this code with new form API

webapp/packages/core-ui/src/Form/FormState.ts
sergeyteleshev: I had problems with editing different connections. The form did not dispose because it was infinitely awaiting this. Now it's okay. Reverted this change. I suppose it was like this because I had some issues with `isOutdated` in form parts case it went into infinite loading because of the resources `sync` method

webapp/packages/plugin-connections/src/ConnectionForm/Options/ConnectionFormOptionsPart.ts
```diff
@@ -0,0 +1,492 @@
...
+    }
+  }
+
+  get connectionKey() {
+    if (!this.state.connectionId || !this.formState.state.projectId) {
+      return null;
+    }
+
+    return createConnectionParam(this.formState.state.projectId, this.state.connectionId);
+  }
```
Wroud: you can re-use it in several places especially in the components

webapp/packages/plugin-connections/src/ConnectionForm/Options/ConnectionFormOptionsPart.ts
```diff
@@ -0,0 +1,492 @@
...
+      return false;
+    }
+
+    const isProjectOutdated = this.projectInfoResource.isOutdated(this.formState.state.projectId);
+
+    if (!this.connectionKey) {
+      return isProjectOutdated;
+    }
+
+    return this.connectionInfoResource.isOutdated(this.connectionKey) || isProjectOutdated;
```
```suggestion
    if (this.projectInfoResource.isOutdated(this.formState.state.projectId)) {
      return true;
    }

    return this.connectionKey && this.connectionInfoResource.isOutdated(this.connectionKey);
```

webapp/packages/plugin-connections/src/ConnectionForm/Options/ConnectionFormOptionsPart.ts
```diff
@@ -0,0 +1,492 @@
...
+  }
+
+  protected override async loader(): Promise<void> {
+    if (this.formState.mode === 'create') {
+      const defaults = await this.getDefaults();
+
+      this.setInitialState({
+        ...defaults,
+        ...this.state,
+      });
```
Wroud: setInitialState should set the initial state of the form. In case we create new entity we should always return the initial state object, we should not mix it with `state` this will lead to the form being detected as `not changed` because after calling the loader initial state and state will be equal

webapp/packages/plugin-connections/src/ConnectionForm/Options/ConnectionFormOptionsPart.ts
```diff
@@ -0,0 +1,492 @@
...
+      return;
+    }
+
+    if (!this.state.connectionId || !this.formState.state.projectId) {
+      console.error('Connection id and project id should be defined');
+      return;
+    }
+
+    const info = await this.connectionInfoResource.load(createConnectionParam(this.formState.state.projectId, this.state.connectionId), [
```
Wroud: you can reuse `this.connectionKey`

webapp/packages/plugin-connections/src/ConnectionForm/Options/ConnectionFormOptionsPart.ts
```diff
@@ -0,0 +1,492 @@
...
+      config.providerProperties = { ...info.providerProperties };
+    }
+
+    if (info.mainPropertyValues) {
+      config.mainPropertyValues = { ...info.mainPropertyValues };
+    }
+
+    this.formState.state.availableDrivers = [info.driverId];
+
+    this.setInitialState({ ...config });
```
Wroud: probably we don't need spread here

webapp/packages/core-ui/src/Form/FormState.ts
sergeyteleshev: Okay. The issue was that `statusMessage` and `statusType` did not set when form is loaded. I need it for this case:
https://github.com/dbeaver/cloudbeaver/blob/9f7ce5c5b418b8987faf23b4443c6d3a28b254cf/webapp/packages/plugin-connections/src/ConnectionForm/OriginInfo/ConnectionFormOriginInfoFormPart.ts#L97-L113

To show the notification that user need to sign in to have an access to origin (GCP on the screen) tab 
Screenshot 2025-03-25 at 11 45 32

sergeyteleshev: So I set it directly on load as well

webapp/packages/plugin-connections/src/ConnectionForm/Options/ConnectionFormOptionsPart.ts
```diff
@@ -0,0 +1,492 @@
...
+  }
+
+  isNameAutoFill() {
+    const isAutoFill = this.state.name === this.initialState.name || this.initialState.name === null;
+
+    return isAutoFill && this.formState.mode === 'create';
```
```suggestion
    if (this.formState.mode === FormMode.Edit) {
      return false;
    }

    return this.state.name === this.initialState.name || this.state.name === null;
```

webapp/packages/plugin-connections/src/ConnectionForm/Options/ConnectionFormOptionsPart.ts
```diff
@@ -0,0 +1,492 @@
...
+  }
+
+  setAuthModel(model: DatabaseAuthModel) {
+    this.state.credentials = {};
+
+    const info =
+      this.formState.state.projectId && this.state.connectionId
+        ? this.connectionInfoResource.get(createConnectionParam(this.formState.state.projectId, this.state.connectionId!))
+        : null;
```
Wroud: you can reuse `this.connectionKey`

webapp/packages/plugin-connections/src/ConnectionForm/Options/ConnectionFormOptionsPart.ts
```diff
@@ -0,0 +1,492 @@
...
+
+    if (model.id === info?.authModel) {
+      if (info.authProperties) {
+        for (const property of info.authProperties) {
+          if (!property.features.includes('password')) {
+            this.state.credentials[property.id!] = property.value;
+          }
+        }
+      }
+    }
```
Wroud: i think that we need to:
1. compare `model.id` with `this.initialState.authModelId`
2. when different reset `state.credentials` to `{}` (empty object)
3. when equal, spread the initial state to the current state: `this.state.credentials = { ...this.initialState.credentials }`

webapp/packages/plugin-connections/src/ConnectionForm/Options/ConnectionFormOptionsPart.ts
```diff
@@ -0,0 +1,492 @@
...
+      config.name = 'New connection';
+      return config;
+    }
+
+    const driver = await this.dbDriverResource.load(this.state.driverId, ['includeProviderProperties']);
+
+    config.authModelId = driver?.defaultAuthModel;
+    config.configurationType = getDefaultConfigurationType(driver);
+    config.host = this.state.host || driver?.defaultServer || 'localhost';
+    config.port = this.state.port || driver?.defaultPort;
```
Wroud: using `state` is unsafe here because it may affect initial state

webapp/packages/plugin-connections/src/ConnectionForm/Options/ConnectionFormOptionsPart.ts
```diff
@@ -0,0 +1,492 @@
...
+    if (!this.state.driverId || !this.formState.state.projectId) {
+      return;
+    }
+
+    const driver = await this.dbDriverResource.load(this.state.driverId, ['includeProviderProperties', 'includeMainProperties']);
+
+    if (this.formState.mode === 'edit') {
+      this.state.connectionId = this.state.connectionId;
+    }
+
```
Wroud: probably you can remove it

webapp/packages/plugin-connections/src/ConnectionForm/Options/ConnectionFormOptionsPart.ts
```diff
@@ -0,0 +1,492 @@
...
+    this.state.networkHandlersConfig = observable([]);
+
+    this.state.name = this.state.name?.trim();
+
+    if (this.state.name && this.formState.mode === 'create') {
+      const connections = this.connectionInfoResource.get(ConnectionInfoProjectKey(this.formState.state.projectId)).filter(isNotNullDefined);
+      const connectionNames = connections.map(connection => connection.name);
+
+      this.state.name = getUniqueName(this.state.name, connectionNames);
+    }
```
Wroud: maybe it's better to move it to the save method (cuz it will change the name and if there is be error saving connection, it's a chance that it will modify the name twice)
1. Connection name 2
2. Connection name 2 2

webapp/packages/plugin-connections/src/ConnectionForm/Options/ConnectionFormOptionsPart.ts
```diff
@@ -0,0 +1,492 @@
...
+
+      const properties = await this.getConnectionAuthModelProperties(this.state.authModelId, info);
+      const passwordProperty = properties.find(property => property.features.includes('password'));
+      const isPasswordEmpty =
+        passwordProperty &&
+        (this.state.credentials?.[passwordProperty.id!] === passwordProperty.defaultValue || !this.state.credentials?.[passwordProperty.id!]);
+
+      if (isCredentialsChanged(properties, this.state.credentials!)) {
+        this.state.credentials = prepareDynamicProperties(properties, toJS(this.state.credentials!));
+      }
```
Wroud: probably you can use `this.initialState.credentials` instead (to compare for changes)

Wroud (changes requested):

webapp/packages/plugin-connections-administration/src/ConnectionForm/ConnectionAccess/ConnectionAccess.tsx
sergeyteleshev: no. cause after saving `OptionsPart` we set `mode` to `edit` and this will show lead to incorrect behavior

webapp/packages/plugin-connections/src/ConnectionForm/ConnectionForm.tsx
sergeyteleshev: This will change the behavior. current behavior closes form after creation
If we remove this logic, we won't close the form after creation

webapp/packages/core-ui/src/Form/FormState.ts
Wroud: I think that issue is with `formAuthState` itself and not with the base form API

webapp/packages/plugin-connections/src/ConnectionForm/Options/ConnectionFormOptionsPart.ts
sergeyteleshev: A little bit changed it with `isNotNullDefined`

webapp/packages/plugin-connections/src/ConnectionForm/Options/ConnectionFormOptionsPart.ts
sergeyteleshev: seems fine after few tests, thanks

webapp/packages/plugin-connections/src/ConnectionForm/Options/ConnectionFormOptionsPart.ts
sergeyteleshev: no it is not

- it will modify name
- if there is no the same name on the second save it will be the same
- if the modified name has also the clone by name, it will modify as you said yes

moreover the same situation can happened during saving also

so lets keep it here as it was, cause most mutations like this with the state should be in format method for convenience

webapp/packages/plugin-connections/src/ConnectionForm/Options/ConnectionFormOptionsPart.ts
sergeyteleshev: lets not. it has the logic with `file`

Wroud (dismissed):

Wroud (dismissed):

webapp/packages/plugin-connections-administration/src/ConnectionForm/ConnectionAccess/ConnectionFormAccessPart.ts
```diff
@@ -31,7 +31,7 @@ export class ConnectionFormAccessPart extends FormPart<string[], IConnectionForm
   }
 
   override isOutdated(): boolean {
-    return this.optionsPart.isOutdated();
+    return Boolean(this.optionsPart.connectionKey && this.connectionInfoResource.isOutdated(this.optionsPart.connectionKey));
```
Wroud, line 34: you must call `await this.connectionInfoResource.load(this.optionsPart.connectionKey)` in the `loader` otherwise there is a chance that no one will call it and part will be endlessly loading

webapp/packages/plugin-connections/src/ConnectionForm/Options/ConnectionFormOptionsPart.ts
```diff
@@ -89,7 +100,10 @@ export class ConnectionFormOptionsPart extends FormPart<IConnectionFormOptionsSt
   }
 
   private async askCredentials(state: IConnectionFormOptionsState): Promise<IConnectionFormOptionsState | null> {
-    if (state.saveCredentials && !this.formState.state.requiredNetworkHandlersIds.length) {
+    const driver = this.state.driverId ? this.dbDriverResource.get(this.state.driverId) : undefined;
+    const isCredentialsRequired = !state.saveCredentials || this.formState.state.requiredNetworkHandlersIds.length;
+
+    if (!isCredentialsRequired || driver?.anonymousAccess) {
```
Wroud, line 100: probably `anonymousAccess` should be joined with `state.saveCredentials` because it's a chance that connection (database) don't have authentication but have SSH or SSL

webapp/packages/plugin-connections/src/ConnectionForm/OriginInfo/ConnectionFormOriginInfoFormPart.ts
```diff
@@ -51,11 +51,11 @@ export class ConnectionFormOriginInfoFormPart extends FormPart<IConnectionFormOr
...
-    return (
-      this.dbDriverResource.isOutdated(this.optionsPart.state.driverId) ||
-      !!(this.authModelId && this.databaseAuthModelsResource.isOutdated(this.authModelId)) ||
-      this.userInfoResource.isOutdated()
-    );
+    const isDriverOutdated = Boolean(this.optionsPart.state.driverId && this.dbDriverResource.isOutdated(this.optionsPart.state.driverId));
+    const isAuthModelOutdated = Boolean(this.authModelId && this.databaseAuthModelsResource.isOutdated(this.authModelId));
+    const isUserInfoOutdated = this.userInfoResource.isOutdated();
+
+    return isDriverOutdated || isAuthModelOutdated || isUserInfoOutdated;
```
Wroud: i can see that you load `connectionInfoOriginDetailsResource` in `loader` also, probably you need to add it here too

webapp/packages/plugin-connections/src/ConnectionForm/SSL/ConnectionFormSSLPart.ts
```diff
@@ -52,7 +52,10 @@ export class ConnectionFormSSLPart extends FormPart<INetworkHandlerConfig, IConn
   }
 
   override isOutdated(): boolean {
-    return this.dbDriverResource.isOutdated(this.optionsPart.state.driverId) || this.networkHandlerResource.isOutdated();
+    const isDriverOutdated = Boolean(this.optionsPart.state.driverId && this.dbDriverResource.isOutdated(this.optionsPart.state.driverId));
+    const isNetworkHandlerOutdated = this.networkHandlerResource.isOutdated(CachedMapAllKey);
+
+    return isDriverOutdated || isNetworkHandlerOutdated;
```
Wroud: you also use `connectionInfoResource` in loader (but without .load you need to change it to load) and you should add it here too (you also need to make sure that `networkHandlersConfig` loaded (probably you need to add `include`)

Wroud (changes requested):

webapp/packages/plugin-connections-administration/src/ConnectionForm/ConnectionAccess/ConnectionFormAccessPart.ts
sergeyteleshev: sure! deleted it then since it is just perform update method in the part loader

webapp/packages/plugin-connections/src/ConnectionForm/SSL/ConnectionFormSSLPart.ts
sergeyteleshev: I added new resources for each `include`, as we agreed on the call. also added each resource to `isOutdated` respectively and in the components where it is required

webapp/packages/plugin-connections/src/ConnectionForm/OriginInfo/ConnectionFormOriginInfoFormPart.ts
sergeyteleshev: thanks. refactored it a bit according to the loader logic

webapp/packages/plugin-connections/src/ConnectionForm/Options/ConnectionFormOptionsPart.ts
sergeyteleshev: then it wont work as before. for instance the case with creating SQLite connection. you will have the ask credentials popup but you shouldn't

Wroud (dismissed):

Wroud (dismissed):

Wroud (dismissed):

Wroud (approved):

Wroud (approved):

PR #3289, opened 2025-02-25
CB-6271 add timeout of inaction in configuration mode

HocKu7:
The user screen after the time is up
image


yagudin10 (dismissed):

devnaumov:
Why do we need a wrapper for validate method in form part. Imagine this scenario: We need to save the form state, but the server is down for example. The user clicks the save button, triggering the form validation process. The validation method could be asynchronous and relies on a .load method from an resource. However, this .load method throws an error, and since there is no try-catch block within the validation method, the execution of the entire save process stops.

As a result, no error messages are displayed from the validation handler, and form.state.exception is not set because the .save handler never executed. This leaves us in a situation where the failure is unclear, making debugging difficult.

To prevent this, we should wrap the validation logic in a try-catch block at the form part base level.

Wroud (dismissed):

sergeyteleshev (dismissed):

yagudin10 (approved):

PR #3294, opened 2025-02-28
87 create checkbox component

SychevAndrey:
image
image


common-react/@dbeaver/ui-kit/src/Checkbox/Checkbox.css
```diff
@@ -0,0 +1,98 @@
+@import './_base.css';
+@import './_sizes.css';
+
+@layer base {
```
Wroud, line 4: Should we use the `components` layer instead?

common-react/@dbeaver/ui-kit/src/Checkbox/Checkbox.css
```diff
@@ -0,0 +1,98 @@
...
+    outline-offset: var(--dbv-kit-outline-offset);
+    outline-width: var(--dbv-kit-control-outline-width);
+    outline-color: var(--dbv-kit-color-outline);
+    transition-property: stroke-dashoffset;
+    transition-timing-function: cubic-bezier(0.4, 0, 0.2, 1);
+    transition-duration: 150ms;
+    stroke-dasharray: 20;
+    stroke-dashoffset: 20;
+
+    &[data-focus-visible] {
```
Wroud, line 56: Should we add `&:focus-visible,` as you did for input and button?

common-react/@dbeaver/ui-kit/src/Checkbox/Checkbox.tsx
```diff
@@ -6,15 +6,64 @@
...
+  size?: ControlSize;
+  icon?: React.ReactNode;
+  indeterminate?: boolean;
+  indeterminateIcon?: React.ReactNode;
+}
 
-export function Checkbox({ children, ...props }: CheckboxProps) {
+export function Checkbox({ children, className, icon, indeterminate, indeterminateIcon, size = 'medium', ...props }: CheckboxProps) {
+  const [checked, setChecked] = useState(props.defaultChecked ?? props.checked ?? false);
+  const [focusVisible, setFocusVisible] = useState(false);
```
Wroud: This is not correctly synchronized code. Control may be controlled and not controlled; therefore, "controlling" props must be prioritized over the inner state.

```
const [innerChecked, setInnerChecked] = ...
const checked = props.checked ?? innerChecked;
```

Wroud (changes requested):

common-react/@dbeaver/ui-kit/src/Checkbox/Checkbox.css
SychevAndrey: No, it's artificial node and can't get any state except `:hover`. Only parent can be :focus-visible

common-react/@dbeaver/ui-kit/src/Checkbox/Checkbox.css
SychevAndrey: base is good, because we want our classes be easily adjusted on other levels

common-react/@dbeaver/ui-kit/src/Checkbox/Checkbox.tsx
SychevAndrey: Thank you!

Wroud (dismissed):

Wroud (approved):

PR #3302, opened 2025-03-04
88 create select component

SychevAndrey:


SychevAndrey:
Why we need state in SelectField component:
https://github.com/ariakit/ariakit/issues/4392

common-react/@dbeaver/ui-kit/src/Select/SelectField.tsx
```diff
@@ -0,0 +1,169 @@
...
+  disabled?: boolean;
+
+  /** Is field required */
+  required?: boolean;
+
+  /** Select's width */
+  width?: string | number;
+
+  /** Custom class name */
+  className?: string;
```
Wroud: These fields don't need any documentation (same for others fields like value or onChange)

common-react/@dbeaver/ui-kit/src/Select/SelectField.tsx
```diff
@@ -0,0 +1,169 @@
...
+   * Function or property path to extract label or render content from items
+   * Examples: 'name', 'user.name', (item) => <span className="custom">{item.firstName}</span>
+   */
+  itemRender?: PropertyGetter<ItemType, React.ReactNode>;
+
+  /**
+   * Function or property path to extract disabled state
+   * Examples: 'isDisabled', 'permissions.canSelect', (item) => !item.isActive
+   */
+  disabledGetter?: PropertyGetter<ItemType, boolean>;
```
Wroud: this fields have no common naming, i suggest to use their items / item or options / option:
```
options
optionValue
optionRender
optionDisabled
```
or
```
items
itemValue
itemRender
itemDisabled
```

common-react/@dbeaver/ui-kit/src/Select/SelectField.tsx
```diff
@@ -0,0 +1,169 @@
...
+  const path = pathOrGetter.split('.');
+  let value: any = item;
+
+  for (const prop of path) {
+    if (value === null || value === undefined) return defaultGetter(item);
+    value = value[prop];
+  }
+
+  return value !== undefined ? (value as Value) : defaultGetter(item);
+}
```
Wroud: Please remove `property path` (it's hard to typesafe and hard to support complex cases; it's just not worth it)

common-react/@dbeaver/ui-kit/src/Select/SelectField.tsx
```diff
@@ -0,0 +1,169 @@
...
+    getValueByPath<ItemType, React.ReactNode>(item, itemRender, i => ('label' in i ? (i as unknown as SelectOption<T>).label : String(i)));
+
+  const isItemDisabled = (item: ItemType): boolean =>
+    getValueByPath<ItemType, boolean>(item, disabledGetter, i => ('disabled' in i ? Boolean((i as unknown as SelectOption<T>).disabled) : false));
+
+  const [selectedValue, setSelectedValue] = useState<T | undefined>(() => {
+    if (value !== undefined) return value;
+
+    const firstEnabledOption = options.find(item => !isItemDisabled(item));
+    return options.length > 0 ? getItemValue(firstEnabledOption || options[0]!) : undefined;
```
Wroud: It is a bit strange that we filter for disabled values, but if there is only one disabled value, we will apply it.

There will also be a synchronization issue. We can't create a Select field if there is no selected value — this will trigger auto selection. Still, the onChange event won't be fired, so the selector will display some randomly selected value while the selector's owner will be in a state like "nothing is selected."

common-react/@dbeaver/ui-kit/src/Select/SelectField.tsx
```diff
@@ -0,0 +1,169 @@
...
+  });
```
Wroud: There is also the same issue as with the checkbox: controlled value is not respected

common-react/@dbeaver/ui-kit/src/Select/SelectField.tsx
```diff
@@ -0,0 +1,169 @@
...
+  placeholder?: string;
+
+  /** Makes field disabled */
+  disabled?: boolean;
+
+  /** Is field required */
+  required?: boolean;
+
+  /** Select's width */
+  width?: string | number;
```
Wroud: do we really need this field?

Wroud (changes requested):

common-react/@dbeaver/ui-kit/src/Select/SelectField.tsx
SychevAndrey: Although it' very convenient to set width like that, we can definately use tailwind utils for that easily via className prop, removed

common-react/@dbeaver/ui-kit/src/Select/SelectField.tsx
SychevAndrey: Don't get the first option, if there is no any enabled one, keep undefined

common-react/@dbeaver/ui-kit/src/Select/SelectField.tsx
```diff
@@ -0,0 +1,143 @@
...
+import './SelectField.css';
+
+export interface SelectOption<T = string> {
+  value: T;
+  label: string;
+  disabled?: boolean;
+}
+
+type PropertyPath = string;
+type PropertyGetter<ItemType, ValueType> = PropertyPath | ((item: ItemType) => ValueType);
```
Wroud: we need to use typesafe PropertyPath calculated based on ValueType and ItemType:
You need to filter fields in ItemType so only items that satisfy ValueType then we need to extract keys of this fields as fields that can be used

Wroud (approved):

PR #3311, opened 2025-03-11
CB-3798 poc api for visibility of properties

yagudin10:


webapp/packages/core-blocks/src/ObjectPropertyInfo/ObjectPropertyInfoForm/RenderField.tsx
```diff
@@ -55,25 +63,42 @@ function getValue(value: any, controlType: ObjectPropertyType) {
...
   showRememberTip,
   saved,
   className,
   canShowPassword,
   onFocus,
+  ...rest
 }) {
   const translate = useTranslate();
 
+  let readonly = rest.readOnly;
```
Wroud: It is a strange move with read-only. Let's revert it as it was from props as `readOnly`, then you can change it if needed (it's not props mutation)

webapp/packages/core-blocks/src/ObjectPropertyInfo/evaluate.ts
```diff
@@ -0,0 +1,29 @@
...
+ */
+
+import subscript from 'subscript';
+import { err, token } from 'subscript/parse';
+
+import 'subscript/feature/bool.js';
+
+const UNDEFINED_MAPPER_VALUE = new Array(2).fill(undefined, 1);
+token('undefined', 20, a => (a ? err() : UNDEFINED_MAPPER_VALUE));
+token('null', 20, a => (a ? err() : UNDEFINED_MAPPER_VALUE));
```
Wroud, line 16: It is a bit strange that we treat null as an undefined

webapp/packages/core-blocks/src/ObjectPropertyInfo/evaluate.ts
```diff
@@ -0,0 +1,29 @@
...
+ * you may not use this file except in compliance with the License.
+ */
+
+import subscript from 'subscript';
+import { err, token } from 'subscript/parse';
+
+import 'subscript/feature/bool.js';
+
+const UNDEFINED_MAPPER_VALUE = new Array(2).fill(undefined, 1);
+token('undefined', 20, a => (a ? err() : UNDEFINED_MAPPER_VALUE));
```
Wroud, line 15: why not use `token('undefined', 20, a => a ? err() : [, undefined])` as in original example?
https://github.com/dy/subscript?tab=readme-ov-file#extending

webapp/packages/core-blocks/src/ObjectPropertyInfo/evaluate.ts
```diff
@@ -0,0 +1,29 @@
...
+token('null', 20, a => (a ? err() : UNDEFINED_MAPPER_VALUE));
+
+interface IContext {
+  object: Record<string, any>;
+}
+
+export function evaluate(expression: string, object: Record<string, any>) {
```
Wroud: please fix linter

Wroud (changes requested):

HocKu7 (dismissed):

webapp/packages/core-blocks/src/ObjectPropertyInfo/evaluate.ts
devnaumov: For the backend, null means no value, for us there will be never null value for the field, only undefined, we want to always fallback null to undefined

webapp/packages/core-blocks/src/ObjectPropertyInfo/evaluate.ts
devnaumov: Eslint error here [, undefined] because of the comma

Wroud (dismissed):

webapp/packages/core-sdk/src/getObjectPropertyValue.ts
```diff
@@ -0,0 +1,24 @@
+/*
+ * CloudBeaver - Cloud Database Manager
+ * Copyright (C) 2020-2025 DBeaver Corp and others
+ *
+ * Licensed under the Apache License, Version 2.0.
+ * you may not use this file except in compliance with the License.
+ */
+
+import type { ObjectPropertyInfo } from '@cloudbeaver/core-sdk';
```
Wroud: please use relative import

webapp/packages/core-sdk/src/getObjectPropertyValue.ts
```diff
@@ -0,0 +1,24 @@
...
+
+  if (!isNotNullDefined(property.value)) {
+    return isBoolean ? false : '';
+  }
+
+  if (typeof property.value === 'string') {
+    return isBoolean ? property.value.toLowerCase() === 'true' : property.value;
+  }
+
+  return property.value;
```
Wroud:
```suggestion
  const isBoolean = property.dataType === 'boolean';

  if (typeof property.value === 'string') {
    return isBoolean ? property.value.toLowerCase() === 'true' : property.value;
  }

  if (isNotNullDefined(property.value)) {
    return property.value;
  }
  
  return isBoolean ? false : '';
```

webapp/packages/plugin-connections/src/ConnectionForm/Options/ConnectionFormOptionsPart.ts
```diff
@@ -221,7 +228,7 @@ export class ConnectionFormOptionsPart extends FormPart<IConnectionFormOptionsSt
     if (authPropertiesInfo.authProperties) {
       for (const property of authPropertiesInfo.authProperties) {
         if (!property.features.includes('password')) {
-          config.credentials[property.id!] = property.value;
+          config.credentials[property.id!] = getObjectPropertyValue(property) ?? property.defaultValue;
```
Wroud: This combination will work strangely because `getObjectPropertyValue` will never return null or undefined, so `??` will never happen

webapp/packages/plugin-connections/src/ConnectionForm/Options/ConnectionFormOptionsPart.ts
```diff
@@ -305,19 +312,24 @@ export class ConnectionFormOptionsPart extends FormPart<IConnectionFormOptionsSt
...
     if (modelId === this.initialState.authModelId) {
       this.state.credentials = { ...this.initialState.credentials };
     } else if (modelId !== this.state.authModelId) {
       this.state.credentials = {};
     }
 
+    if (modelId) {
+      const properties = await this.getConnectionAuthModelProperties(modelId);
+      this.state.credentials = prepareDynamicProperties(properties, this.state.credentials ?? {});
+    }
```
Wroud: this may lead to "changed" form state (it's unexpected)

Wroud (changes requested):

webapp/packages/plugin-connections/src/ConnectionForm/Options/ConnectionFormOptionsPart.ts
```diff
@@ -322,12 +321,8 @@ export class ConnectionFormOptionsPart extends FormPart<IConnectionFormOptionsSt
...
       this.state.credentials = { ...this.initialState.credentials };
     } else if (modelId !== this.state.authModelId) {
-      this.state.credentials = {};
-    }
-
-    if (modelId) {
-      const properties = await this.getConnectionAuthModelProperties(modelId);
-      this.state.credentials = prepareDynamicProperties(properties, this.state.credentials ?? {});
+      const properties = modelId ? await this.getConnectionAuthModelProperties(modelId) : null;
+      this.state.credentials = properties ? getObjectPropertyDefaults(properties) : {};
```
```
const properties = modelId ? await this.getConnectionAuthModelProperties(modelId) : [];
this.state.credentials = getObjectPropertyDefaults(properties);
```

Wroud (dismissed):

alexander-skoblikov (approved):

PR #3313, opened 2025-03-11
103 create radio and radio group components

SychevAndrey:
https://github.com/user-attachments/assets/412aea15-5f02-465c-b8c6-d20681bc32e2


common-react/@dbeaver/ui-kit/src/Input/Input.tsx
```diff
@@ -6,18 +6,25 @@
...
   const classNameToApply = `dbv-kit-input dbv-kit-input--${size ?? 'medium'} ${className}`.trim();
   return (
     <div className="dbv-kit-input-wrapper">
-      <input className={classNameToApply} {...props} />
+      {props.label && (
+        <label htmlFor={id} className={`dbv-kit-input__label ${props.required ? 'dbv-kit-input__label--required' : ''}`}>
+          {props.label}
+        </label>
+      )}
+      <input id={id} className={classNameToApply} {...props} />
```
Wroud: I don't think embedding labels into base input control is a good idea. I will suggest using a composite pattern where you can make more complex controls from a composition of a few simple controls (like Input + Lable)

common-react/@dbeaver/ui-kit/src/Radio/Radio.tsx
```diff
@@ -0,0 +1,25 @@
...
+export interface RadioProps extends Omit<_RadioProps, 'size'> {
+  size?: ControlSize;
+}
+
+export function Radio({ children, className, size = 'medium', ...props }: RadioProps) {
+  return (
+    <label className={`dbv-kit-radio dbv-kit-radio--${size} ${className || ''}`}>
+      <_Radio className="dbv-kit-radio__input" {...props} />
+      <div className="dbv-kit-radio__control" />
+      <span className="dbv-kit-radio__text">{children}</span>
```
Wroud: this can be any react element `dbv-kit-radio__text`

common-react/@dbeaver/ui-kit/src/index.ts
```diff
@@ -21,3 +21,15 @@ export {
   type SelectItemProps,
 } from './Select/Select.js';
 export { SelectField, type SelectFieldProps, type SelectItem } from './Select/SelectField.js';
+export {
+  Radio,
+  RadioGroup,
+  _Radio,
+  _RadioGroup,
+  _RadioProvider,
```
Wroud: if we don't use it right now it's better to not export inner primitives from ariakit

common-react/@dbeaver/ui-kit/src/Input/Input.tsx
Wroud: (this is for discussion)

common-react/@dbeaver/ui-kit/src/Radio/RadioGroup.tsx
```diff
@@ -0,0 +1,72 @@
...
+ */
+
+import { useId } from 'react';
+import './RadioGroup.css';
+import { _RadioGroup, _RadioProvider, Radio, type RadioProviderProps } from './index.js';
+import type { ControlSize } from '../types/controls.js';
+
+type ConditionalRadioProps =
+  | {
+      items: Array<{ value: string; label: string }>;
```
Wroud: should we use the same pattern as in the selector to extract value/label from any object?

common-react/@dbeaver/ui-kit/src/Input/Input.tsx
Wroud: same issue with className

Wroud: maybe we have it in other elements too

Wroud (changes requested):

common-react/@dbeaver/ui-kit/src/Radio/RadioGroup.tsx
SychevAndrey: Actually I just added this option with items prop for more flexibility, I would not recommend or promote this approach in any way, so we can just remove those props, but I would not make it more complicated.

common-react/@dbeaver/ui-kit/src/Radio/Radio.tsx
SychevAndrey: I don't understand, what we want to do and why? span here is only for a styling purposes, you can pass any children there, I don't see a problem. But im most cases it will be just a text. Label and span both have phrasing content model and accept same descendants. https://html.spec.whatwg.org/multipage/forms.html#the-label-element
And even some of the phrasing content elements are not recommended to use  in `label` because it breaks a11y, for example <a />, so <span> is a good default.

Wroud: i mean, that class name is like, "This class is only for some text element", but it's more like label, name, or something else

common-react/@dbeaver/ui-kit/src/Radio/RadioGroup.tsx
Wroud: If so, we really can remove it

Wroud (dismissed):

Wroud (approved):

PR #3317, opened 2025-03-13
CB-6115 Add dispose to FormState instances

sergeyteleshev:


.vscode/extensions.json
```diff
@@ -19,6 +19,7 @@
     "yzhang.markdown-all-in-one",
     "GraphQL.vscode-graphql-syntax",
     "mrmlnc.vscode-json5",
-    "esbenp.prettier-vscode"
+    "esbenp.prettier-vscode",
+    "stelesh.git-workspace-helper"
```
sergeyteleshev: ..🥺..
👉👈

SychevAndrey (dismissed):

.vscode/extensions.json
SychevAndrey: I'm not against adding that extension, but I'm not sure if it's a good place to do it

Wroud: I will recommend moving it to the EE repo because it's not a part of the community (it's more like tool for developers who uses our codebase every day)

webapp/packages/core-ui/src/Form/FormState.ts
```diff
@@ -94,6 +97,7 @@ export class FormState<TState> implements IFormState<TState> {
       setPartsState: action,
       setState: action,
       updateFormState: action,
+      save: action,
```
Wroud: This does nothing because `save` is asynchronous. Action works only for synchronous mutations. You can use `runInAction` instead (but probably not in this case)

Wroud (changes requested):

devnaumov (dismissed):

Wroud (dismissed):

webapp/packages/plugin-authentication-administration/src/Administration/Users/Teams/TeamsForm/useTeamsAdministrationFormState.ts
```diff
@@ -24,5 +24,12 @@ export function useTeamsAdministrationFormState(id: string | null, configure?: (
     configure?.(ref.current);
   }
 
+  useEffect(
```
Wroud, line 28: you also need to dispose any ref before setting new state

webapp/packages/plugin-authentication-administration/src/Administration/Users/Teams/TeamsTable/CreateTeamService.ts
```diff
@@ -47,4 +47,9 @@ export class CreateTeamService {
   create(): void {
     this.teamsAdministrationNavService.navToCreate();
   }
+
+  dispose() {
```
Wroud, line 52: there is no guarantee that fillData will not be executed when we already have form initialized, so please call dispose before setting new form state

Wroud (changes requested):

Wroud (dismissed):

devnaumov (approved):

PR #3338, opened 2025-03-24
Cb 6265 Return front unit tests after migration

sergeyteleshev:


SychevAndrey (dismissed):
LGTM 👍🏻

webapp/packages/core-cli/configs/vitest.config.ts
```diff
@@ -0,0 +1,43 @@
+/*
+ * CloudBeaver - Cloud Database Manager
+ * Copyright (C) 2020-2025 DBeaver Corp and others
+ *
+ * Licensed under the Apache License, Version 2.0.
+ * you may not use this file except in compliance with the License.
+ */
+/// <reference types="vitest/config" />
```
Wroud: do we really need this reference?

webapp/packages/core-cli/configs/vitest.config.ts
```diff
@@ -0,0 +1,43 @@
...
+    environment: 'jsdom',
+    root: path.resolve('.'),
+    include: [
+      // unix
+      'packages/*/lib/**/*.{spec,test}.{js,jsx}',
+      'lib/**/*.{spec,test}.{js,jsx}',
+      // windows
+      'packages\\*\\lib\\**\\*.{spec,test}.{js,jsx}',
+      'lib\\**\\*.{spec,test}.{js,jsx}',
+    ],
```
Wroud: probably, we can use `include: ["**/lib/**/*.test.js"],`
(it also should be universal and work for Windows too, I think that's handled by vitest itself)

webapp/packages/core-cli/configs/vitest.config.ts
```diff
@@ -0,0 +1,43 @@
...
+    alias: [
+      {
+        find: /.*.(jpg|jpeg|png|gif|eot|otf|webp|svg|ttf|woff|woff2|mp4|webm|wav|mp3|m4a|aac|oga)%/,
+        replacement: path.resolve(__dirname, './__mocks__/fileMock.js'),
+      },
+      {
+        find: /.*.(css|scss|less)$/,
+        replacement: path.resolve(__dirname, './__mocks__/styleMock.js'),
+      },
+    ],
```
Wroud: this handled by vitest itself and can be removed

webapp/packages/core-cli/configs/vitest.config.ts
```diff
@@ -0,0 +1,43 @@
...
+    globals: true,
```
Wroud: please remove (we want to use default's global:false)

webapp/packages/core-cli/configs/vitest.setup.ts
```diff
@@ -0,0 +1,15 @@
+/*
+ * CloudBeaver - Cloud Database Manager
+ * Copyright (C) 2020-2025 DBeaver Corp and others
+ *
+ * Licensed under the Apache License, Version 2.0.
+ * you may not use this file except in compliance with the License.
+ */
+import { expect } from 'vitest';
+import * as matchers from 'jest-extended';
+import '@testing-library/jest-dom';
```
Wroud: https://github.com/testing-library/jest-dom?tab=readme-ov-file#with-vitest

webapp/packages/core-cli/src/dependency-validator/validateDependencies.ts
```diff
@@ -77,7 +77,7 @@ export function validateDependencies(currentPackageRoot: string) {
     }
 
     if (isTestFile) {
-      devDependencies.add('@types/jest');
+      devDependencies.add('vitest');
     }
```
Wroud: i think it will be added by default because we will use import from `vitest` so we probably can remove it

webapp/packages/core-resource/src/Resource/CachedDataResource.test.ts
```diff
@@ -1,11 +1,11 @@
 /*
  * CloudBeaver - Cloud Database Manager
- * Copyright (C) 2020-2024 DBeaver Corp and others
+ * Copyright (C) 2020-2025 DBeaver Corp and others
  *
  * Licensed under the Apache License, Version 2.0.
  * you may not use this file except in compliance with the License.
  */
-import { beforeEach, describe, expect, jest, test } from '@jest/globals';
+import { beforeEach, describe, expect, vitest, test } from 'vitest';
```
Wroud, line 8: why not `vi` instead of `vitest`?

webapp/packages/tsconfig/test.d.ts
```diff
@@ -0,0 +1,9 @@
+/*
+ * CloudBeaver - Cloud Database Manager
+ * Copyright (C) 2020-2025 DBeaver Corp and others
+ *
+ * Licensed under the Apache License, Version 2.0.
+ * you may not use this file except in compliance with the License.
+ */
+/// <reference types="@testing-library/jest-dom" />
+export * from '@testing-library/jest-dom';
```
Wroud: we should avoid using globals (this will mess up with all TS files)

Wroud (changes requested):

webapp/packages/core-cli/configs/vitest.config.ts
sergeyteleshev: This is the recommendation from documentation - https://vitest.dev/guide/

Screenshot 2025-03-26 at 18 00 00

webapp/packages/core-cli/configs/vitest.config.ts
sergeyteleshev: From previous experience with the Vite assets plugin this worked out. I'll check that on Windows also

webapp/packages/core-cli/configs/vitest.config.ts
sergeyteleshev: We can safely remove mapping to the files, but not for CSS files cause it leads to the error

Removed the first alias then

Screenshot 2025-03-26 at 18 10 20

webapp/packages/core-resource/src/Resource/CachedDataResource.test.ts
sergeyteleshev: idk, just did it. its the same

webapp/packages/core-cli/configs/vitest.config.ts
Wroud: that's sounds srange: https://vitest.dev/config/#css

Wroud: this is because we need `@wroud/vite-plugin-asset-resolver`

Wroud (dismissed):

webapp/packages/core-blocks/src/importLazyComponent.test.tsx
sergeyteleshev, line 1: This test is useless; I deleted it. importLazyComponent is just a syntax sugar which has no logic to test

webapp/packages/core-utils/src/timestampToDate.ts
```diff
@@ -1,11 +1,19 @@
...
+  return new Date(timestamp).toLocaleString('en-US', {
+    year: 'numeric',
+    month: '2-digit',
+    day: '2-digit',
+    hour: '2-digit',
+    minute: '2-digit',
+    second: '2-digit',
+    hour12: false,
+  });
 }
```
Wroud: Please revert. This must be handled on the test's side. You are currently changing the behavior of the application

webapp/packages/core-cli/configs/vitest.setup.ts
```diff
@@ -5,13 +5,26 @@
...
 import { cleanup } from '@testing-library/react';
 import '@testing-library/jest-dom/vitest';
 
 // different machine has its own timezones and some tests can fail because of it
 process.env.TZ = 'UTC';
 
-afterEach(() => {
+function resetMocks() {
+  vi.clearAllTimers();
+  vi.useFakeTimers();
```
Wroud: Do we want to use fake timers by default?

Wroud (changes requested):

webapp/packages/core-cli/configs/vitest.setup.ts
sergeyteleshev: Sure. Real-timers run in real-time. So it takes tests longer

If tests have timers, it is required to fake timers
If tests have no timers, there is no harm to performance or test run environment

These changes save you a few lines of code, so you don't need to mock it again and again, and they also save you from forgetting about mocking timers

webapp/packages/core-utils/src/__mocks__/mockToLocaleString.ts
```diff
@@ -0,0 +1,32 @@
...
+ * CloudBeaver - Cloud Database Manager
+ * Copyright (C) 2020-2025 DBeaver Corp and others
+ *
+ * Licensed under the Apache License, Version 2.0.
+ * you may not use this file except in compliance with the License.
+ */
+
+import { afterEach, beforeEach } from 'vitest';
+
+export function mockToLocaleString(): void {
```
Wroud: name is not descriptive

Wroud (approved):

PR #3370, opened 2025-04-08
dbeaver/pro#5234 mark children nodes as deleted

devnaumov:
closes [5234](https://github.com/dbeaver/pro/issues/5234)


webapp/packages/plugin-navigation-tree/src/TreeNew/useTreeData.ts
```diff
@@ -27,27 +27,33 @@ interface IOptions {
...
   nodeTransformers?: TreeDataTransformer<INode>[];
+  parentTransformers?: TreeDataTransformer<string | null>[];
   stateTransformers?: TreeDataTransformer<INodeState>[];
 }
 
 export function useTreeData(options: IOptions): ITreeData {
   options = useObservableRef(
     {
       ...options,
+      getParent: options.getParent ?? (() => null),
```
Wroud: you don't need to fallback to `() => null` because the type remains nullable and later, you still using ? operand `options.getParent?.(id)`

Wroud (changes requested):

Wroud (approved):

PR #3374, opened 2025-04-09
#5362 refactor: move blob formatter to shared codebase

SychevAndrey:
Screenshot 2025-04-09 at 15 18 35
Screenshot 2025-04-09 at 15 18 49


common-typescript/@dbeaver/result-set-api/src/ResultSetApi.ts
```diff
@@ -9,6 +9,14 @@ export class ResultSetApi {
     return new Date(stringValue);
   }
 
+  static getBlobValue(value: any): string | null {
+    if (value === null) {
+      return null;
+    }
+
+    return 'blob';
+  }
```
Wroud: We need to return `Blob` here instead. Also, please extend `getValueDataType` to support getting type based on the value

Wroud (changes requested):

Wroud (dismissed):

common-react/@dbeaver/react-data-grid/src/formatters/BlobFormatter.tsx
```diff
@@ -0,0 +1,14 @@
+import { clsx } from '@dbeaver/ui-kit';
+
+interface Props {
+  className?: string;
+  onClick?: React.MouseEventHandler<HTMLElement>;
+}
+
+export function BlobFormatter({ className, onClick }: Props) {
+  return (
+    <span className={clsx('tw:uppercase', className)} onClick={onClick}>
```
sergeyteleshev, line 10: is this applied here also? it was in previous implementation

`overflow: hidden;
white-space: pre;
text-overflow: ellipsis;`

sergeyteleshev: I can see on the screen that it is applied. But are those styles included in `tw:uppercase`?

sergeyteleshev (dismissed):

common-react/@dbeaver/react-data-grid/src/formatters/BlobFormatter.tsx
SychevAndrey: no, those styles are not applied and we don't need them there. tw:uppercase it's a simple text-transform:uppercase

Wroud (dismissed):

webapp/packages/product-base/src/manualChunks.ts
```diff
@@ -7,90 +7,60 @@
...
-                    const packageName = packageMatch[1]; // e.g. "plugin-data-export"
+                if (id.includes('common-')) {
+                  return 'common';
+                }
 
-                    const moduleInfo = getModuleInfo(id);
-                    if (!moduleInfo) {
-                      return null;
-                    }
+                if (id.includes('plugin-')) {
```
Wroud: `core-` packages not covered, this will lead to may files from this chunks

webapp/packages/product-base/src/manualChunks.ts
```diff
@@ -7,90 +7,60 @@
...
-                    // info.dynamicImporters is the subset of those that import it dynamically.
-                    // If there’s any importer that did a static import, we consider this module sync.
-                    return info.importers && info.importers.some(importer => !info.dynamicImporters.includes(importer));
-                  }
-
-                  const nodeModulesMatch = /[\\/]node_modules[\\/](.*?)[\\/]/.exec(id);
+const vendorChunks = new Map([
+  ['d3-', 'vendor/d3'],
+  ['lodash', 'vendor/lodash'],
+  ['codemirror', 'vendor/codemirror'],
```
Wroud: probably we want to extract react and react-dom as well as some dependencies listed here: `packages/core-cli/configs/excludedFromVendor.js`

(not all of them should be extracted to separate chunks, for example d3 and degree should be in the same chunk because usually used together)

`codemirror` + `@codemirror` + `@lezer`

`leaflet` + `react-leaflet`

Wroud (changes requested):

Wroud (approved):

PR #3384, opened 2025-04-16
CB-5465 fixes checkbox defaultChecked wrong behavior

HocKu7:
Closes https://github.com/dbeaver/pro/issues/5125


webapp/packages/core-blocks/src/FormControls/Checkboxes/useCheckboxState.ts
```diff
@@ -56,6 +56,8 @@ export function useCheckboxState<TKey extends string>(options: CheckboxStateOpti
     } else {
       checked = !!currentState;
     }
+  } else if (name !== undefined && state !== undefined) {
+    checked = false;
```
SychevAndrey: looks strange. It should be a part of the first condition imho. like  `if (name in state === false) checked = false`.

Wroud (approved):

PR #3415, opened 2025-05-06
5490 cb create spinner UI kit component

SychevAndrey:
https://github.com/user-attachments/assets/24e5caab-46cb-4b8f-973f-7c5193a90c7b


sergeyteleshev (dismissed):

common-react/@dbeaver/ui-kit/src/Button/Button.tsx
```diff
@@ -10,6 +10,7 @@ import clsx from 'clsx';
 import './Button.css';
 import type { ControlSize } from '../types/controls.js';
 import { componentProviderWrapper } from '../componentProviderWrapper.js';
+import { Spin } from '../index.js';
```
Wroud: It's better not to import from the index (may lead to wrong circular dependencies)

common-react/@dbeaver/ui-kit/src/Spin/Spin.tsx
```diff
@@ -0,0 +1,21 @@
...
+import './Spin.css';
+import type { ControlSize } from '../types/controls.js';
+
+export interface SpinProps {
+  enabled?: boolean;
+  className?: string;
+  size?: ControlSize;
+}
+
+export function Spin({ className, size, enabled = true }: SpinProps) {
```
Wroud: i think that `Spinner` is more common name for such component (I found Spin only in Ant design)

common-react/@dbeaver/ui-kit/src/Spin/Spin.tsx
```diff
@@ -0,0 +1,21 @@
+import clsx from 'clsx';
+import './Spin.css';
+import type { ControlSize } from '../types/controls.js';
+
+export interface SpinProps {
+  enabled?: boolean;
```
Wroud: active?

common-react/@dbeaver/ui-kit/src/Spin/Spin.tsx
```diff
@@ -0,0 +1,21 @@
...
+  className?: string;
+  size?: ControlSize;
+}
+
+export function Spin({ className, size, enabled = true }: SpinProps) {
+  return (
```
Wroud:
```js
if (!active) {
  return null;
}
```

common-react/@dbeaver/ui-kit/src/Spin/Spin.tsx
```diff
@@ -0,0 +1,21 @@
...
+    enabled && (
+      <span className={clsx('dbv-kit-spin', size ? `dbv-kit-spin--${size}` : undefined, className)}>
```
Wroud: `size ? 'dbv-kit-spin--${size}' : undefined` -> `size && 'dbv-kit-spin--${size}'`

common-react/@dbeaver/ui-kit/src/Spin/_base.css
```diff
@@ -0,0 +1,37 @@
...
+    stroke-dasharray: 45px, 100px;
+    stroke-dashoffset: -10px;
+  }
+  100% {
+    stroke-dasharray: 17px, 100px;
+    stroke-dashoffset: -58px;
+  }
+}
+
+@keyframes spinning {
```
Wroud: can we use animation from tailwindcss? https://tailwindcss.com/docs/animation

Wroud (changes requested):

common-react/@dbeaver/ui-kit/src/Spin/Spin.tsx
SychevAndrey: why it's better?

common-react/@dbeaver/ui-kit/src/Spin/_base.css
SychevAndrey: no, we need to control duration and timing function.

common-react/@dbeaver/ui-kit/src/Spin/Spin.tsx
Wroud: compare:
```
return (
  active && (
    ...
  )
);
```

```
return active && (
  ...
);
```


```
if (!active) {
  return null;
}

return (
  ...
);
```

I prefer the last variant, but between the first and second, the first one seems incorrect at all

Wroud (approved):

sergeyteleshev (approved):

Wroud (approved):

PR #3419, opened 2025-05-06
[CB] Add SQL Editor autosave setting #5405

sergeyteleshev:
closes https://github.com/dbeaver/pro/issues/5405


webapp/packages/plugin-sql-editor/src/locales/en.ts
```diff
@@ -44,5 +52,6 @@ export default [
   ['plugin_sql_editor_settings_import_max_size', 'Import limit (KB)'],
   ['plugin_sql_editor_settings_import_max_size_description', 'Maximum file size that can be imported in KB'],
   ['plugin_sql_editor_settings_auto_save', 'Auto save'],
-  ['plugin_sql_editor_settings_auto_save_description', 'Automatically save the editor content'],
+  ['plugin_sql_editor_settings_auto_save_description', 'Enables auto-saving for SQL editor sessions'],
```
Wroud: What is "SQL editor sessions"?

webapp/packages/plugin-sql-editor/src/locales/en.ts
```diff
@@ -44,5 +52,6 @@ export default [
...
+  ['plugin_sql_editor_settings_auto_save_description_te', 'Enables auto-saving for SQL editor sessions. Works only with disabled Git feature'],
```
Wroud: Actually, it has no effect (this setting) when the git feature is enabled

webapp/packages/plugin-sql-editor/src/SqlEditorSettingsService.ts
```diff
@@ -141,16 +142,18 @@ export class SqlEditorSettingsService extends Dependency {
...
-        // },
+        {
+          group: SQL_EDITOR_SETTINGS_GROUP,
+          key: 'plugin.sql-editor.autoSave',
+          access: {
+            scope: ['client', 'server'],
+          },
+          type: ESettingsValueType.Checkbox,
+          name: 'plugin_sql_editor_settings_auto_save',
+          description: this.serverConfigResource.distributed
```
Wroud: Maybe we can check here for the git feature, or write a generic message that always adds the information that the setting has no effect when git integration is enabled.

webapp/packages/plugin-sql-editor/src/locales/fr.ts
```diff
@@ -1,3 +1,11 @@
+/*
+ * CloudBeaver - Cloud Database Manager
+ * Copyright (C) 2020-2025 DBeaver Corp and others
+ *
+ * Licensed under the Apache License, Version 2.0.
+ * you may not use this file except in compliance with the License.
+ */
+
```
Wroud: please remove license headers from locale files

Wroud (changes requested):

webapp/packages/plugin-sql-editor/src/locales/en.ts
sergeyteleshev: Well this is the description from our website documentation:
https://dbeaver.com/docs/cloudbeaver/Product-configuration-parameters/#sql-editor

webapp/packages/plugin-sql-editor/src/locales/en.ts
sergeyteleshev: I slightly disagree here. See the attached video for TE version

https://github.com/user-attachments/assets/7d53c0dd-15f2-4085-ad50-0c8ba74206b8

webapp/packages/plugin-sql-editor/src/locales/en.ts
sergeyteleshev: But I agree that "sessions" abstraction can confuse user a bit. Probably we need to change description in both places

webapp/packages/plugin-sql-editor/src/SqlEditorSettingsService.ts
sergeyteleshev: Generic message is not the option cause CE, EE, and AWS products do not have git integration yet, and it can confuse users

Instead of checking distributed, added checking of git integration

webapp/packages/plugin-sql-editor/src/locales/en.ts
```diff
@@ -45,4 +45,8 @@ export default [
   ['plugin_sql_editor_settings_import_max_size_description', 'Maximum file size that can be imported in KB'],
   ['plugin_sql_editor_settings_auto_save', 'Auto save'],
   ['plugin_sql_editor_settings_auto_save_description', 'Automatically save the editor content'],
+  [
+    'plugin_sql_editor_settings_auto_save_description_git_integration',
+    'Automatically save the editor content. This feature is only available when Git integration is turned off',
```
Wroud: By default, SQL editor content is saved in local storage, but this feature kicks in when the editor is opened to edit an SQL script file.

This message may confuse users, as it controls only script-saving behaviour and doesn't affect saving in local storage

Wroud (changes requested):

PR #3420, opened 2025-05-06
dbeaver/pro#4958 Add default catalog and default schema

HocKu7:
Closes https://github.com/dbeaver/pro/issues/4958


sergeyteleshev (dismissed):

webapp/packages/core-sdk/src/queries/fragments/DatabaseConnection.gql
```diff
@@ -6,6 +6,8 @@ fragment DatabaseConnection on ConnectionInfo {
   driverId
   keepAliveInterval
   autocommit
+  defaultCatalogName
+  defaultSchemaName
```
Wroud: We don't want to load this data for every connection, we need this data only in connection form (you can re-use customIncludeOptions)

Wroud (changes requested):

webapp/packages/core-sdk/src/queries/fragments/DatabaseConnection.gql
devnaumov: Its the same as keepAliveInterval and autocommit

devnaumov: We dont want to rewrite how we load expert settings here

devnaumov: Especially when we are going to rewrite it in dynamic properties in the next sprint

Wroud: You're right; we don't need to load `keepAliveInterval` and `autocommit`, so let's change them, too.
We don't need to "rewrite" anything; we just need to pass the right flags in the right places, with everything already implemented.

We don't want to load information that we don't need.

Wroud (changes requested):

Wroud (approved):

yagudin10 (approved):

PR #3427, opened 2025-05-08
dbeaver/pro#5405 adds autosave for scripts

sergeyteleshev:
closes dbeaver/pro#5405


Wroud (dismissed):

Wroud (approved):

SychevAndrey (commented):
I would use different wording in disabled state for easier understanding of what is going on.  Instead _This feature is only available when Git integration is turned off_ I suggest using **Turn Off Git Integration to use this feature**. It is shorter and tells what to do.

PR #3435, opened 2025-05-12
Dbeaver/pro#5652 add security to endpoints

HocKu7:
Closes https://github.com/dbeaver/pro/issues/5652

### DIFF
--- a/server/bundles/io.cloudbeaver.server/src/io/cloudbeaver/service/core/DBWServiceCore.java
+++ b/server/bundles/io.cloudbeaver.server/src/io/cloudbeaver/service/core/DBWServiceCore.java
@@ -44,10 +44,10 @@ public interface DBWServiceCore extends DBWService {
     /**
      * Returns information of system.
      */
-    @WebAction(authRequired = false)
+    @WebAction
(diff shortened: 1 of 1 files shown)


serge-rider (approved):

server/bundles/io.cloudbeaver.server/src/io/cloudbeaver/service/core/DBWServiceCore.java
```diff
@@ -44,10 +44,10 @@ public interface DBWServiceCore extends DBWService {
     /**
      * Returns information of system.
      */
-    @WebAction(authRequired = false)
+    @WebAction
     WebPropertyInfo[] getSystemInformationProperties(@NotNull WebSession webSession);
 
-    @WebAction(authRequired = false)
+    @WebAction
```
Wroud, line 50: This can't be changed only on the backend; we need to verify that it works on the front end and make some changes.

`getProductSettings` is probably required for anonymous mode
And this is not secure information, it's just a set of possible settings that may be set for the application (it may include default values)

Wroud (changes requested):

Wroud (approved):

PR #3454, opened 2025-05-19
dbeaver/pro#5753 feat: react-dnd

Wroud:
closes dbeaver/pro#5753

### DIFF
--- a/common-react/@dbeaver/react-data-grid/src/DataGrid.tsx
+++ b/common-react/@dbeaver/react-data-grid/src/DataGrid.tsx
@@ -12,6 +12,7 @@ import { mapRenderHeaderCell } from './mapRenderHeaderCell.js';
 import { mapEditCellRenderer } from './mapEditCellRenderer.js';
 import { DataGridRowContext, type IDataGridRowContext } from './DataGridRowContext.js';
 import './DataGrid.css';
+import { HeaderDnDContext, useHeaderDnD } from './useHeaderDnD.js';
 
 export interface ICellPosition {
   rowIdx: number;
@@ -46,9 +47,11 @@ export const DataGrid = forwardRef<DataGridRef, DataGridProps>(function DataGrid
     headerElement,
     getHeaderWidth,
     headerText,
+    getHeaderOrder,
     getHeaderResizable,
     getHeaderHeight,
     getHeaderPinned,
+    getHeaderDnD,
     cell,
     cellText,
     cellElement,
@@ -59,6 +62,7 @@ export const DataGrid = forwardRef<DataGridRef, DataGridProps>(function DataGrid
     rowCount,
     getRowId,
     getRowHeight,
+    onHeaderReorder,
     onScroll,
     onScrollToBottom,
     onFocus,
@@ -87,21 +91,23 @@ export const DataGrid = forwardRef<DataGridRef, DataGridProps>(function DataGrid
         minWidth: 26,
         editable: row => getCellEditable?.(row.idx, i) ?? false,
         frozen: getHeaderPinned?.(i),
-        renderHeaderCell: mapRenderHeaderCell,
-        renderCell: mapCellContentRenderer,
-        renderEditCell: mapEditCellRenderer,
+        renderHeaderCell: mapRenderHeaderCell(i),
+        renderCell: mapCellContentRenderer(i),
+        renderEditCell: mapEditCellRenderer(i),
       };
     });
 
+  const dndHeaderContext = useHeaderDnD({ columns, onReorder: onHeaderReorder, getCanDrag: getHeaderDnD, getHeaderOrder });
+
   useImperativeHandle(ref, () => ({
     selectCell: (position: ICellPosition) => {
-      innerGridRef.current?.selectCell({ idx: position.colIdx, rowIdx: position.rowIdx });
+      innerGridRef.current?.selectCell({ idx: dndHeaderContext.getDataColIdx(position.colIdx), rowIdx: position.rowIdx });
     },
     scrollToCell: (position: Partial<ICellPosition>) => {
-      innerGridRef.current?.scrollToCell({ idx: position.colIdx, rowIdx: position.rowIdx });
+      innerGridRef.current?.scrollToCell({ idx: position.colIdx && dndHeaderContext.getDataColIdx(position.colIdx), rowIdx: position.rowIdx });
     },
     openEditor: (position: ICellPosition) => {
-      innerGridRef.current?.selectCell({ idx: position.colIdx, rowIdx: position.rowIdx }, true);
+      innerGridRef.current?.selectCell({ idx: dndHeaderContext.getDataColIdx(position.colIdx), rowIdx: position.rowIdx }, true);
     },
   }));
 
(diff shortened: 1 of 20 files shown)


SychevAndrey (commented):
Do we really need our own react-dnd implementation? 
For what purpose?
We have a lot of things to support and maintain besides that.

Wroud:
@SychevAndrey  I can't say whether we need it or not, but I can say that we need a universal approach that works in different environments and can communicate with various parts of the application.

My initial idea was to use DataTransfer for this purpose, so I was not using the "react-dnd" library. You can't access `DataTransfer` from it.

But then I faced an issue: `DataTransfer` can be used only in `drag start` and `drop` events, and can't be used in events such as `drag over`. It can be used, but with limitations. You can read what type of data is stored in it, but you can't get values.

This information is essential for realising rich DnD features, such as detecting where you can drop something and where you cannot.

To cover this case, besides access to `DataTransfer`, I've implemented a special store that can be used in addition to `DataTransfer`, but it leads us to almost the exact implementation we have with `react-dnd`.

So now our goal is to test this approach for different tasks to understand our next steps. Theoretically, `DataTransfer` still can be used, but it will require rethinking how DnD works in the application.

common-react/@dbeaver/react-dnd/src/useDrag.ts
```diff
@@ -0,0 +1,68 @@
...
+  }
+
+  function handleDragEnd(event: React.DragEvent<HTMLElement>) {
+    if (!optionsRef.current.draggable) {
+      return;
+    }
+    setDragging(false);
+    optionsRef.current.onDragEnd?.(event, globalDnDStore.getProvider(id));
+    globalDnDStore.removeData(id);
+    console.log('drag end', event);
```
devnaumov: Is console log here for debugging purpose?

common-react/@dbeaver/react-dnd/src/store.ts
```diff
@@ -0,0 +1,44 @@
...
+      },
+      removeData: () => {
+        this.removeData(id);
+      },
+    };
+  }
+}
+
+export function getStoreProvider(event: React.DragEvent<HTMLElement>): DnDStoreProvider | null {
+  const id = event.dataTransfer.types.find(type => type.startsWith('application/dbeaver-react-dnd-'))?.slice(30);
```
devnaumov: Its better not to use plain numbers like this. I think it would be better to move it to const and name accordingly

common-react/@dbeaver/react-dnd/src/useDrag.ts
Wroud: no, i've removed it

common-react/@dbeaver/react-dnd/src/store.ts
Wroud: yep, moved to const

common-react/@dbeaver/react-dnd/src/useDrop.ts
```diff
@@ -0,0 +1,61 @@
...
+
+  function handleOver(event: React.DragEvent<HTMLElement>) {
+    const provider = getStoreProvider(event);
+    const isDroppable = optionsRef.current.canDrop?.(event, provider);
+    if (isDroppable) {
+      event.preventDefault();
+      event.dataTransfer.dropEffect = 'move';
+    } else {
+      event.dataTransfer.dropEffect = 'none';
+    }
```
sergeyteleshev, line 33:
```suggestion
    if (isDroppable) {
      event.preventDefault();
    }
    event.dataTransfer.dropEffect = 'move';
```

common-react/@dbeaver/react-data-grid/src/renderers/HeaderCellContentRenderer.tsx
```diff
@@ -1,13 +1,55 @@
...
   const getHeaderText = useGridReactiveValue(headerElement ? undefined : cellHeaderContext?.headerText, colIdx);
-  return headerElement ?? getHeaderText ?? '';
-}
+
+  const draggable = dndHeaderContext?.getCanDrag?.(colIdx) ?? false;
+  const drag = useDrag({
+    draggable,
+    onDragStart(event, store) {
+      if (dndHeaderContext?.id) {
+        store.setData('dbeaver-react-data-grid-id', dndHeaderContext?.id);
```
sergeyteleshev, line 21: would be nice to get rid of these magic numbers

common-react/@dbeaver/react-data-grid/src/renderers/HeaderCellContentRenderer.tsx
```diff
@@ -1,13 +1,55 @@
...
+  function canDrop(event: React.DragEvent<HTMLElement>, store: DnDStoreProvider | null) {
+    const gridId = store?.getData('dbeaver-react-data-grid-id');
+    return gridId === dndHeaderContext?.id;
+  }
+  const drop = useDrop({
+    canDrop,
+    onDragOver(event, store) {
+      if (!canDrop(event, store)) {
+        return;
+      }
```
sergeyteleshev, line 39: may be this check should go inside useDrop?

https://github.com/dbeaver/cloudbeaver/blob/4e1fcd859f066f0239a8244cc3bcbbfd4720b6df/common-react/%40dbeaver/react-dnd/src/useDrop.ts#L25-L35

common-react/@dbeaver/react-data-grid/src/renderers/HeaderCellContentRenderer.tsx
```diff
@@ -1,13 +1,55 @@
...
+      dndHeaderContext?.onDragOver(dragColIdx, colIdx, isAfter);
+    },
+  });
+
+  return (
+    <div className="tw:h-full tw:content-center" {...drag.props} {...drop.props}>
+      {headerElement ?? getHeaderText ?? ''}
+    </div>
+  );
+});
```
sergeyteleshev, line 55: All these components remind me of the thought to create `useDragNDrop` hook, which combines logic from both `useDrag` and `useDrop` - reducing us to write the same boilerplate logic again and again

common-react/@dbeaver/react-data-grid/src/DataGrid.tsx
```diff
@@ -124,34 +130,40 @@ export const DataGrid = forwardRef<DataGridRef, DataGridProps>(function DataGrid
...
+                renderRow: rowRenderer,
+                renderCell: cellRenderer,
+                noRowsFallback: children,
+              }}
+              minimumRowsToRender={100}
+            />
+          </DataGridCellHeaderContext>
+        </DataGridCellContext>
+      </DataGridRowContext>
+    </HeaderDnDContext>
```
sergeyteleshev, line 167: Something went wrong with the data grid and contexts. This is a combo of contexts accumulated here, which can lead to performance issues.

The odd moment for me here is that we wrap the whole data grid with `ColumnContext`, `RowContext`, even `CellContext`, and `DnDContext`

I suppose in perfect world it should be something like this:
Column - wrapped with `ColumnContext`
Row - wrapped with `RowContext`
Cell - wrapped with `CellContext`
HeaderRow - wrapped with `DnDContext`

sergeyteleshev (approved):

common-react/@dbeaver/react-dnd/src/useDrop.ts
Wroud: This will lead to different behaviour; currently, a non-droppable element will have a special "none" effect

common-react/@dbeaver/react-data-grid/src/renderers/HeaderCellContentRenderer.tsx
Wroud: I've thought about it, but this check will prevent some hooks from being fired. I don't want to implement such behavior because there are cases when you need these events even when you can't drop an element (for example, for visual feedback). The same thing is implemented in the `react-dnd` library, where you can specify `canDrop` and then call it from `monitor` if needed. But I've decided not to make this API for now and keep it simple.

common-react/@dbeaver/react-data-grid/src/renderers/HeaderCellContentRenderer.tsx
Wroud: `useDragNDrop` is sugar that can be used later, for now it's simple on purpose because there is a chance that this code will be entirely rewritten

PR #3457, opened 2025-05-20
[CB] Bump NPM dependencies #5745

sergeyteleshev:
closes https://github.com/dbeaver/pro/issues/5745

### DIFF
--- a/webapp/packages/core-cli/package.json
+++ b/webapp/packages/core-cli/package.json
@@ -34,7 +34,7 @@
     "postinstall": "yarn build"
   },
   "engines": {
-    "node": ">=20.0.0"
+    "node": ">=22.0.0"
   },
   "dependencies": {
     "glob": "^11",
--- a/webapp/packages/core-cli/package.json
+++ b/webapp/packages/core-cli/package.json
@@ -44,7 +44,7 @@
     "@cloudbeaver/tsconfig": "workspace:*",
     "@swc/core": "^1",
     "@types/js-yaml": "^4",
-    "@types/node": "^20",
+    "@types/node": "^22",
     "enquirer": "^2",
     "execa": "^9",
     "glob": "^11",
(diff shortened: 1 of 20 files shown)


webapp/packages/core-cli/package.json
```diff
@@ -44,7 +44,7 @@
     "@cloudbeaver/tsconfig": "workspace:*",
     "@swc/core": "^1",
     "@types/js-yaml": "^4",
-    "@types/node": "^20",
+    "@types/node": "^22",
```
Wroud, line 47: After this update, you probably want to update "engines" to `"node": ">=22.0.0"`
Or you should keep the  ^20 version (to avoid use of not supported features)

In case we want to update to node >=22, then we need to ensure that it's updated in documentation and the devops team updates pipelines in Jenkins, also, maybe we will need to update GitHub Actions

Wroud (changes requested):

webapp/packages/core-view/src/View/parseHotkey.ts
```diff
@@ -38,21 +54,21 @@ export function mapKey(key: string): string {
...
+export function isHotkeyModifier(key: string): boolean {
   return reservedModifierKeywords.includes(key);
 }
 
 export function parseKeysHookInput(keys: string, splitKey = ','): string[] {
   return keys.split(splitKey);
 }
 
-export function parseHotkey(hotkey: string, combinationKey = '+', description?: string): Hotkey {
+export function parseHotkey({ hotkey, combinationKey = '+', description, useKey = false, isSequence = false }: IParseHotkeysArgs): IHotkeyParsed {
```
Wroud: It is accepting arguments that no longer exist, there are only two uses of this function, and each: `parseHotkey({ hotkey: key })`

Wroud (changes requested):

Wroud (approved):

PR #3472, opened 2025-05-23
dbeaver/pro#5215 track close event on layout close

devnaumov:
closes [5215](https://github.com/dbeaver/pro/issues/5215)

Previously, when the panel was closed using the overlay close button, only the OptionsPanelService.close() method was executed. As a result, the itemId was not properly reset. I have moved the item reset logic into the executor handler to ensure it always runs.

It might have worked in some parts, but the behavior was inconsistent


webapp/packages/core-ui/src/Screens/AppScreen/OptionsPanelService.ts
```diff
@@ -55,22 +57,25 @@ export class OptionsPanelService {
...
       return false;
     }
 
+    await this.closeTask.execute('after', context);
+
     this.panelComponent = null;
     this.basePanelComponent = null;
     this.active = false;
+
     return true;
```
Wroud, line 79: This change is not backward compatible, so we need to check all cases where we are adding the handler to `on close` executor (for example, `PublicConnectionFormService`) and update them; otherwise, we will get double logic execution

Wroud (changes requested):

Wroud (dismissed):

Wroud (approved):

PR #3478, opened 2025-05-27
[CB] hotkeys with "/" and "\" dont work after hotkeys library version upgrade #5781

sergeyteleshev:
closes https://github.com/dbeaver/pro/issues/5781


webapp/packages/plugin-help/src/Shortcuts/Shortcut.tsx
```diff
@@ -16,6 +16,12 @@ interface Props {
   shortcut: IShortcut;
 }
 
+const CODE_FORMAT_MAP: Record<string, string> = {
+  comma: ',',
+  slash: '/',
+  backslash: '\\',
+};
```
Wroud: This logic should be handled in `getCommonAndOSSpecificKeys` because it's library-specific (why do we need to change our application every time the library updates)

Wroud (changes requested):

webapp/packages/plugin-help/src/Shortcuts/Shortcut.tsx
```diff
@@ -24,12 +30,19 @@ export const Shortcut: React.FC<Props> = function Shortcut({ shortcut }) {
...
-          <React.Fragment key={code}>
-            {index > 0 && <span className={s(style, { span: true })}>{translate('ui_or')}</span>}
-            <div className={s(style, { shortcutCode: true })}>{code}</div>
-          </React.Fragment>
-        ))}
+        {shortcut.code.map((code, index) => {
+          const formattedCode = code
+            .split(' ')
+            .map(code => CODE_FORMAT_MAP[code.toLowerCase()] ?? code)
+            .join(' ');
```
Wroud: Probably it should be handled in the `transformKeys`
[transformKeys](https://github.com/dbeaver/cloudbeaver/blob/0edfb29851a80780a11d959eac043beb3794dc85/webapp/packages/plugin-help/src/Shortcuts/SHORTCUTS_DATA.ts#L102)

Wroud (changes requested):

webapp/packages/plugin-help/src/Shortcuts/SHORTCUTS_DATA.ts
```diff
@@ -102,17 +102,31 @@ export const NAVIGATION_TREE_SHORTCUTS: IShortcut[] = [
...
+  }
+
+  return key;
 }
 
 function transformModToDisplayKey(key: string): string {
   const OS = getOS();
   if (OS === OperatingSystem.windowsOS || OS === OperatingSystem.linuxOS) {
-    return key.replace('MOD', 'CTRL');
+    return key.replaceAll('MOD', 'CTRL');
```
sergeyteleshev: I added also `replaceAll` logic to consider sequential shortcuts if we need it in future releases so here everything works just fine

webapp/packages/plugin-help/src/Shortcuts/SHORTCUTS_DATA.ts
```diff
@@ -102,17 +102,31 @@ export const NAVIGATION_TREE_SHORTCUTS: IShortcut[] = [
 function transformKeys(keyBinding: IKeyBinding): string[] {
   const keys = getCommonAndOSSpecificKeys(keyBinding);
 
-  return keys.map(key => transformModToDisplayKey(key.toLocaleUpperCase().replace(/\+/gi, ' + ')));
+  return keys.map(formatKeysSymbols).map(key => transformModToDisplayKey(key.toLocaleUpperCase().replace(/\+/gi, ' + ')));
+}
+
+function formatKeysSymbols(key: string): string {
+  const KEY_FORMAT_MAP: Record<string, string> = {
```
devnaumov: It can be moved outside the function to avoid creating the object each time the function runs, since its content is static

webapp/packages/plugin-help/src/Shortcuts/SHORTCUTS_DATA.ts
```diff
@@ -100,19 +124,48 @@ export const NAVIGATION_TREE_SHORTCUTS: IShortcut[] = [
...
+        result.push('OPTION');
+      }
+
+      continue;
+    }
+
+    if (FORMAT_SHORTCUT_KEYS_MAP[lowerCaseCode]) {
+      result.push(FORMAT_SHORTCUT_KEYS_MAP[lowerCaseCode]);
+      continue;
+    }
```
Wroud: Instead of `if + continue`, it's better to use `switch`:
```js
switch (true) {
  case lowerCaseCode === 'mod':
  break;
  case lowerCaseCode === 'alt':
  break;
  case !!FORMAT_SHORTCUT_KEYS_MAP[lowerCaseCode]:
  break;
  default:
  result.push(lowerCaseCode);
}
```

Also, it's better to use a map (because we replace one key with another, we can't push multiple keys for some given key):

```js
return key.split(APPLIED_DIVIDER).map(code => {
  const lowerCaseCode = code.toLowerCase();
  
  switch (true) {
    case lowerCaseCode === 'mod':
    return 'code';
    case lowerCaseCode === 'alt':
    return 'code';
    case !!FORMAT_SHORTCUT_KEYS_MAP[lowerCaseCode]:
    return 'code'
    default:
    return code;
}
}).join(APPLIED_DIVIDER)
```

There is also a different result from the original implementation, where all the codes were modified from any case to `toLocaleUpperCase`

Wroud (changes requested):

Wroud (dismissed):

Wroud (approved):

PR #3487, opened 2025-05-29
dbeaver/pro#5802 feat: table api

Wroud:
closes dbeaver/pro#5802


SychevAndrey (commented):
👍🏻 
I like Mutex!
I'm not sure about nanoevents. We have built-in EventTarget, why to add something?
We definitely need to decide how we gonna write private methods/props, using `private _method` or `#method`.

PR #3489, opened 2025-05-30
[CB] Use UIKit Button component for SplitControls implementation #5799

sergeyteleshev:
closes https://github.com/dbeaver/pro/issues/5799

https://github.com/user-attachments/assets/088e1a15-e089-4bf1-8785-df04af1c0582


webapp/packages/core-blocks/src/Split/SplitControls.tsx
```diff
@@ -65,18 +66,22 @@ export const SplitControls: React.FC = function SplitControls() {
...
       )}
       {split.state.mode !== 'maximize' && (
-        <button className={s(styles, { button: true, primary: inverse, resizeButton: isResizeMode })} type="button" onClick={handlers.handleExpand}>
-          <div className={s(styles, { ripple: true })} />
-        </button>
+        <SplitControlButton
+          className={s(styles, { resizeButton: isResizeMode })}
+          mode={inverseMode}
+          split={split.state.split}
+          variant={inverse ? 'primary' : undefined}
```
SychevAndrey: this is not the right usage of variant property. It relates to inner Button behaviour/appearance, but you are using it for adding a custom class and logic for SplitControlButton. It's better to add a special property.

SychevAndrey (changes requested):
Looks good! 👍🏻 Need to fix one issue with props.

And I'm not sure about using tailwind colors directly. It can create a problem in future, maybe it's better to create additional --theme-* tokens and use it here. But it's definitely better comparing to what we used to have before in terms of visibility and contrast.

SychevAndrey:
Do we still use _button styles after your changes? If not, we should remove it. If yes, we need to create a task to eliminate them.

SychevAndrey (approved):

sergeyteleshev:
Required autotests to check if buttons are alright

webapp/packages/core-theming/src/styles/_mixins.scss
```diff
@@ -19,7 +19,6 @@
 @import './theme-ripple';
 @import './ripple';
 @import './color';
-@import './button';
```
Wroud, line 22: 🤙🤙🤙

Wroud (approved):

PR #3501, opened 2025-06-03
[CB] spread new ConnectionInfoResources across the app #5412

sergeyteleshev:
closes https://github.com/dbeaver/pro/issues/5412

### DIFF
--- a/webapp/packages/core-connections/src/ConnectionInfoResource.ts
+++ b/webapp/packages/core-connections/src/ConnectionInfoResource.ts
@@ -50,7 +50,7 @@ import { parseConnectionKey } from './parseConnectionKey.js';
 export type Connection = DatabaseConnection & {
   authProperties?: UserConnectionAuthPropertiesFragment[];
 };
-export type ConnectionInitConfig = Omit<InitConnectionMutationVariables, 'includeAuthNeeded' | 'includeCredentialsSaved' | 'includeProperties'>;
+export type ConnectionInitConfig = Omit<InitConnectionMutationVariables, 'includeProperties'>;
 export type ConnectionInfoIncludes = Omit<GetUserConnectionsQueryVariables, 'id'>;
 
 export const NEW_CONNECTION_SYMBOL = Symbol('new-connection');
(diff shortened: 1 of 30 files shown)


webapp/packages/core-sdk/src/queries/fragments/DatabaseConnection.gql
```diff
@@ -9,52 +9,22 @@ fragment DatabaseConnection on ConnectionInfo {
...
-  providerProperties @include(if: $includeProviderProperties)
-
   requiredAuth
 
   features
   supportedDataFormats
 
-  authNeeded @include(if: $includeAuthNeeded)
+  authNeeded
   authModel
```
Wroud: I think that we can move some of the authentication-related fields to a separate resource:
```
saveCredentials
credentialsSaved
sharedCredentials
requiredAuth
mainPropertyValues?
authNeeded
authModel
```

Wroud: So, if the database is not connected, we will load this info and then perform the authentication workflow if needed

Wroud: probably we can load it together with `authProperties`

Wroud (changes requested):

webapp/packages/core-sdk/src/queries/fragments/DatabaseConnection.gql
sergeyteleshev: Moved everything to `ConnectionInfoAuthPropertiesResource` except `mainPropertyValues`, which is only required for options and located in `ConnectionInfoCustomOptionsResource`

webapp/packages/core-connections/src/ConnectionInfoResource.ts
```diff
@@ -466,10 +422,7 @@ export class ConnectionInfoResource extends CachedMapResource<IConnectionInfoPar
     await this.performUpdate(key, [], async () => {
       const { connection } = await this.graphQLService.sdk.updateConnection({
         projectId: key.projectId,
-        connectionId: key.connectionId,
```
Wroud, line 469: Is this an expected change?

Wroud (approved):

PR #3511, opened 2025-06-05
dbeaver/pro#5673 Add search ability for the value panel

SychevAndrey:
This is not a final version in terms of visual representation. We are waiting for the icons for Next/Prev match and Replace buttons.

Added search panel using codemirror search extension via createPanel.
gotoLine panel customized with styles only.

image


webapp/packages/plugin-codemirror6/src/ReactCodemirrorSearchPanel.tsx
```diff
@@ -0,0 +1,108 @@
...
+}
+
+export const ReactCodemirrorSearchPanel: React.FC<Props> = observer(function ReactCodemirrorSearchPanel({ className, incomingView, top }) {
+  const dom = useMemo(() => document.createElement('div'), []);
+  const compartment = useMemo(() => new Compartment(), []);
+  const context = useContext(ReactCodemirrorContext);
+  const view = incomingView ? context?.incomingView : context?.view;
+  const searchInputRef = useRef<HTMLInputElement>(null);
+  const [searchMatchesCount, setSearchMatchesCount] = useState({ count: 0, current: 1 });
+  const [queryState, setQueryState] = useState<SearchQuery>(view ? getSearchQuery(view?.state) : new SearchQuery({ search: '' }));
```
Wroud: Is this performant enough to execute on every render? `view ? getSearchQuery(view?.state) : new SearchQuery({ search: '' })`

I will suggest at least to use callback `() => view ? getSearchQuery(view?.state) : new SearchQuery({ search: '' })`

webapp/packages/plugin-codemirror6/src/SearchPanel/SearchPanel.css
```diff
@@ -0,0 +1,76 @@
...
+    --dbv-kit-input-height: 26px;
+    --dbv-kit-icon-btn-small-size: 20px;
+
+    display: flex;
+    align-items: center;
+    gap: var(--tw-spacing);
+
+    container-type: inline-size;
+    min-width: 180px;
+    padding: var(--tw-spacing);
```
Wroud, line 16: Why not use Tailwind classes?

webapp/packages/plugin-codemirror6/src/SearchPanel/SearchPanel.tsx
```diff
@@ -0,0 +1,249 @@
...
+
+import './SearchPanel.css';
+
+export function SearchPanel({
+  view,
+  searchMatchesCount,
+  queryState,
+  inputRef,
+}: {
+  view: EditorView;
```
Wroud: please use `interface Props`

webapp/packages/plugin-codemirror6/src/SearchPanel/SearchPanel.tsx
```diff
@@ -0,0 +1,249 @@
...
+import type { EditorView } from '@codemirror/view';
+import { useState } from 'react';
+
+import { Icon, useTranslate } from '@cloudbeaver/core-blocks';
+import { clsx, IconButton, Input } from '@dbeaver/ui-kit';
+
+import './SearchPanel.css';
+
+export function SearchPanel({
+  view,
```
Wroud: let's remove `view` from props, expose:
```
onQueryCahnge
onCaseSensetiveToggle
onLiteralToggle
onWholeWordToggle
onReplaceChange
onFindNext
onFindPrevious
onReplaceAll
onReplaceNext
onClose
```

The idea is to keep logic coupled with `view` in the same component / or hook. The coupled logic is split into two components and can't be used separately.

webapp/packages/plugin-codemirror6/src/theme/_base-code-editor.scss
```diff
@@ -93,6 +94,75 @@
...
+    & input {
+      @include mdc-typography('body2');
+      max-width: 60px;
+      height: 24px;
+      padding-inline: 8px;
+
+      &:focus {
+        @include mdc-theme-prop(border-color, primary, false);
+      }
+    }
```
Wroud, line 124: What is this style for?

Wroud (changes requested):
please fix linting errors

webapp/packages/plugin-codemirror6/src/ReactCodemirrorSearchPanel.tsx
Wroud: Or better to add a `null` variant for the value that is set by default

SychevAndrey: Yep, I guess callback is enough here.

webapp/packages/plugin-codemirror6/src/SearchPanel/SearchPanel.css
SychevAndrey: If they repeat, it's better to use CSS to decrease payload. What is the problem with this approach?

webapp/packages/plugin-codemirror6/src/theme/_base-code-editor.scss
SychevAndrey: gotoLine

SychevAndrey: image

SychevAndrey:
https://github.com/user-attachments/assets/2decdd35-2b7e-4680-8ec4-ffcb8673511e

fix resizing

webapp/packages/plugin-codemirror6/src/SearchPanel/SearchPanel.css
Wroud: What do you mean by payload?
The issue is that every time, custom styles that are repeated will increase the size of the styles files, for example:
We have multiple places where we need to apply flex styles. For instance, each time it is declared per component, it will add the same styles to the CSS files (duplication). If we use Tailwind, the size of the styles will not increase.

Another case:
We have many common patterns in UI, such as a block with a given size, and we want the text in this block to wrap or overflow with ellipsis.

Tailwind provides many utilities that will apply a set of required styles for the desired behaviour; otherwise, manually applying them every time increases the chance of losing some required styles.

Summarizing:
It may be a good idea to use a custom style for UI components or some widely reused components. On the other hand, in places where we are building some UI parts, we should avoid using custom styles (use components that exist, or apply tailwind classes) to decrease overall style duplication.

webapp/packages/plugin-codemirror6/src/SearchPanel/SearchPanel.tsx
```diff
@@ -0,0 +1,223 @@
...
+
+import { Icon, useTranslate } from '@cloudbeaver/core-blocks';
+import { clsx, IconButton, Input, Icon as UIKitIcon } from '@dbeaver/ui-kit';
+
+import './SearchPanel.css';
+
+interface SearchPanelProps {
+  searchMatchesCount?: { count: number; current: number };
+  queryState: SearchQuery;
+  inputRef?: React.RefObject<HTMLInputElement | null>;
```
Wroud: Instead, it's better to use a custom ref object (useImperativeHandle) with a single `focus` method to exclude the chance of manipulating input unexpectedly outside of the component.

webapp/packages/plugin-codemirror6/src/SearchPanel/SearchPanel.tsx
```diff
@@ -0,0 +1,223 @@
...
+  onWholeWordToggle: () => void;
+  onReplaceChange: (value: string) => void;
+  onFindNext: () => void;
+  onFindPrevious: () => void;
+  onReplaceAll: () => void;
+  onReplaceNext: () => void;
+  onClose: () => void;
+}
+
+export function SearchPanel({
```
Wroud: missing `observer` hok

webapp/packages/plugin-codemirror6/src/useEditorDefaultExtensions.ts
```diff
@@ -70,6 +72,10 @@ export interface IDefaultExtensions {
   rectangularSelection?: boolean;
   keymap?: boolean;
   lineWrapping?: boolean;
+  search?: boolean;
+  searchOptions?: {
+    createPanel?: (view: EditorView) => { dom: HTMLElement };
+  };
```
Wroud: Do we want to expose such customization logic? (I suggest to remove `searchOptions`)

Wroud (changes requested):

Wroud (dismissed):

common-react/@dbeaver/ui-kit/.ladle/components.tsx
```diff
@@ -1,2 +1,14 @@
 import '../src/index.css';
 import './global.css';
+import icons from '../assets/icons/icons.svg?raw';
+
+import type { GlobalProvider } from '@ladle/react';
+
+export const Provider: GlobalProvider = ({ children }) => {
+  return (
+    <div>
+      <div style={{ display: 'none' }} dangerouslySetInnerHTML={{ __html: icons }} />
```
sergeyteleshev: svg icons allow script tags which leads as to a injection threat here. we should purify icons svg file in this case
we have a method for that `toSafeHtmlString`. try it and see if it works here. may be it needs some tuning

webapp/packages/plugin-codemirror6/src/locales/en.ts
```diff
@@ -0,0 +1,14 @@
...
+  ['plugin_codemirror_search_case_sensitive', 'Case Sensitive'],
+  ['plugin_codemirror_search_literal', 'Use Regular Expression'],
+  ['plugin_codemirror_search_whole_word', 'Match Whole Word'],
+  ['plugin_codemirror_search_replace', 'Replace'],
+  ['plugin_codemirror_search_replace_toggle', 'Toggle Replace'],
+  ['plugin_codemirror_search_replace_all', 'Replace All'],
+  ['plugin_codemirror_search_find_next', 'Find Next (Enter)'],
+  ['plugin_codemirror_search_find_previous', 'Find Previous (Shift+Enter)'],
+  ['plugin_codemirror_search_matches_of', 'of'],
+  ['plugin_codemirror_search_matches_none', 'No found'],
```
sergeyteleshev: may be `Not found`?

webapp/packages/plugin-codemirror6/src/ReactCodemirrorSearchPanel.tsx
```diff
@@ -0,0 +1,190 @@
...
+      onWholeWordToggle={handleWholeWordToggle}
+      onReplaceChange={handleReplaceChange}
+      onFindNext={handleFindNext}
+      onFindPrevious={handleFindPrevious}
+      onReplaceAll={handleReplaceAll}
+      onReplaceNext={handleReplaceNext}
+      onClose={handleClose}
+    />,
+    dom,
+  ) as any;
```
sergeyteleshev: no need for any here

sergeyteleshev:
Regular expressions are not working.

Screenshot 2025-06-10 at 11 58 20

Screenshot 2025-06-10 at 11 58 41

sergeyteleshev (changes requested):

common-react/@dbeaver/ui-kit/.ladle/components.tsx
Wroud: But these are not external icons; they are our icons built by us. There is no threat

sergeyteleshev: well, internal threats also can exist. and we should decrease the number of entry points for threats, in my opinion

webapp/packages/plugin-codemirror6/src/ReactCodemirrorSearchPanel.tsx
```diff
@@ -33,8 +34,20 @@ interface Props extends React.PropsWithChildren {
...
   const searchQuery = new SearchQuery(config ?? getSearchQuery(state));
 
-  const cursor = searchQuery.getCursor(state);
+  let cursor;
   const counter = { count: 0, current: 1 };
+  const options = { ignoreCase: !config?.caseSensitive };
+
+  if (config?.regexp) {
+    try {
+      new RegExp(config.search);
```
Wroud: `new RegExp(config.search);` I think we need to add a comment here, that we are checking if the regex is correct, but is it checked by `RegExpCursor` or how it works?

Wroud (dismissed):

sergeyteleshev (dismissed):

webapp/packages/plugin-codemirror6/src/ReactCodemirrorSearchPanel.tsx
SychevAndrey: removed extra check

sergeyteleshev (dismissed):

Wroud (approved):

PR #3515, opened 2025-06-10
dbeaver/pro#5889 fix: unit tests

Wroud:
closes dbeaver/pro#5889

### DIFF
--- a/webapp/packages/tests-runner/src/vitest.setup.ts
+++ b/webapp/packages/tests-runner/src/vitest.setup.ts
@@ -6,19 +6,11 @@
  * you may not use this file except in compliance with the License.
  */
 import { beforeEach, vi } from 'vitest';
-import { cleanup } from '@testing-library/react';
 import '@testing-library/jest-dom/vitest';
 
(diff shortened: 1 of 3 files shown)


webapp/packages/tests-runner/src/vitest.setup.ts
```diff
@@ -6,19 +6,11 @@
  * you may not use this file except in compliance with the License.
  */
 import { beforeEach, vi } from 'vitest';
-import { cleanup } from '@testing-library/react';
 import '@testing-library/jest-dom/vitest';
 
 // different machine has its own timezones and some tests can fail because of it
 process.env.TZ = 'UTC';
 
-function resetDocument() {
```
sergeyteleshev, line 15: I think it is better to leave it according to the documentation
https://testing-library.com/docs/react-testing-library/setup/#auto-cleanup-in-vitest

so it does not depend on our render function

sergeyteleshev (approved):

SychevAndrey (approved):

PR #3517, opened 2025-06-11
VQB: the comment is lost when returning to the VQB tab from the SQL Editor tab #5574

sergeyteleshev:


Wroud:
closed in favor of https://github.com/dbeaver/cloudbeaver/pull/3518

PR #3537, opened 2025-06-21
Correct broken P2 repository URL for local development setup

EastLord:
Context & Verification
This issue is identical to one recently resolved in the main DBeaver repository. The fix applied here mirrors the official fix from the DBeaver team.

The original discussion can be found here:
https://github.com/orgs/dbeaver/discussions/38420

The official fix in the DBeaver repository is this commit:
https://github.com/dbeaver/dbeaver/commit/00824540240c5104709e60948af92a89b6921ce1

### DIFF
--- a/osgi-app.properties
+++ b/osgi-app.properties
@@ -16,7 +16,7 @@ testLibraries=\
   org.hamcrest.core;\
   org.hamcrest
 repositories=\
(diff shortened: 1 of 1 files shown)


serge-rider:
Thanks, missed that too

PR #3539, opened 2025-06-23
dbeaver/pro#6038 fix: can't click on checkbox in OutputLogMenu

SychevAndrey:
we use checkbox as an icon, so we need this element to be transparent to events


webapp/packages/core-ui/src/ContextMenu/MenuItemRenderer.tsx
```diff
@@ -89,7 +89,11 @@ export const MenuItemRenderer = observer<IMenuItemRendererProps>(function MenuIt
         checked={item.checked}
         onClick={() => onClick(false)}
       >
-        <MenuItemElement label={item.label} icon={<Checkbox checked={item.checked} size="small" />} tooltip={item.tooltip} />
+        <MenuItemElement
+          label={item.label}
+          icon={<Checkbox checked={item.checked} size="small" className="tw:pointer-events-none" />}
```
sergeyteleshev: Seems like this is affected by this PR:
https://github.com/dbeaver/cloudbeaver/pull/3530/files

I think pointer events should be applied to `MenuItemElement` in CSS styles, so we have clickable element even if it is not a checkbox

Also, it seems like `theme-ripple` causes many problems with that, and after the discussion, we thought: "do we need it?". Cause it simply implements hover and active styles. Which we can handle on our own without this class. So we should consider removing those classes and replace it with our own hover & active element styles

Wroud: i think that we should revert https://github.com/dbeaver/cloudbeaver/pull/3530/files#diff-9b864cdfebff0eb6502e6c455c6b2718ff0f8b4477d8b174e8988e768a0cfc76R51 instead

Wroud (changes requested):
see comment above

SychevAndrey:
[revert changes made in this branch and #5997, #5989](https://github.com/dbeaver/cloudbeaver/pull/3539/commits/0da0b8556bde82759481bbde8d72cc7f2e5a9ebd)

Wroud (approved):

devnaumov (approved):

PR #3545, opened 2025-06-24
[CB] 'Same as device' theme setting #5445

sergeyteleshev:
closes https://github.com/dbeaver/pro/issues/5445

https://github.com/user-attachments/assets/ea5398ab-d19e-4cb1-8a4e-19005aad3244


webapp/packages/core-theming/src/themes.ts
```diff
@@ -1,31 +1,44 @@
...
 import type { ITheme } from './ThemeService.js';
+import type { ClassCollection } from './themeUtils.js';
 
 const emptyTheme = {};
 
+export enum THEME_ID {
+  LIGHT = 'light',
+  DARK = 'dark',
+  SYSTEM = 'system',
+}
```
Wroud: The theme ID should be just a string, but themes themselves should have type: 'dark' or 'light'
So, in the `ITheme`, you need to add the `type` property

Wroud (changes requested):

webapp/packages/core-blocks/src/useS.ts
```diff
@@ -46,8 +46,8 @@ export function useS<T extends ComponentStyle[]>(...componentStyles: [...T]): Ex
   const stylesRef = useRef<ComponentStyle[]>([]);
   const [patch, forceUpdate] = useState(0);
   const loadedStyles = useRef<BaseStyles[]>([]);
-  const themeService = useService(ThemeService, true);
-  const [currentThemeId, setCurrentThemeId] = useState(() => themeService?.themeId);
+  const themeService = useService(ThemeService);
+  const [currentThemeId, setCurrentThemeId] = useState(() => themeService.themeId);
```
Wroud, line 50: Please keep it optional, it's here for a purpose

Wroud (changes requested):
this PR add a lot of changes to the `ThemeService` that we want to avoid, we can't apply this changes because of hight risk of regression and bugs. Our goal was to introduce minimal changes to current behavior while adding new functionality, but here we suggesting a big rework of theming system rather adding required changes.

PR #3547, opened 2025-06-24
[CB-6221] Add a Clear button to the Output log #4410

sergeyteleshev:
closes https://github.com/dbeaver/pro/issues/4410

https://github.com/user-attachments/assets/64ab20c2-eca5-4fca-92f3-e3eb5a80488c


webapp/packages/plugin-sql-editor/src/SqlResultTabs/OutputLogs/OutputLogsPanel.tsx
```diff
@@ -25,23 +26,22 @@ export const OutputLogsPanel = observer<Props>(function SqlOutputLogsPanel({ sql
...
-
-  const state = useOutputLogsPanelState(outputLogs, sqlEditorTabState);
+  const state = useOutputLogsPanelState();
+  const selectedLogTypes = getComputed(() => sqlEditorTabState.outputLogsTab?.selectedLogTypes);
+  const resultValue = getComputed(() => getOutputLog(outputLogs, selectedLogTypes, state.searchValue ?? ''));
 
   return (
     <Container className="theme-background-secondary" overflow vertical noWrap dense parent gap>
       <OutputLogsToolbar state={state} sqlEditorTabState={sqlEditorTabState} />
-      {data && (
```
sergeyteleshev, line 34: data is always an array

webapp/packages/plugin-sql-editor/src/SqlResultTabs/OutputLogs/OutputLogsService.ts
```diff
@@ -61,11 +77,15 @@ export class OutputLogsService {
...
   }
 
-  getOutputLogs(events: IOutputLog[], editorState: ISqlEditorTabState) {
+  getOutputLogs(events: IOutputLog[], editorState: ISqlEditorTabState): IOutputLog[] {
     const dataSource = this.sqlDataSourceService.get(editorState.editorId);
 
     return events.filter(event => event.contextId === dataSource?.executionContext?.id);
   }
+
+  clearOutputLogs(): void {
```
sergeyteleshev: `this.outputLogsResource.clear();` did work so I did this instead
By that, I mean new logs did not arrive after `clear()` was called

webapp/packages/plugin-sql-editor/src/SqlResultTabs/OutputLogs/useOutputLogsPanelState.ts
```diff
@@ -1,74 +1,29 @@
...
-            }
-
-            if (log.message) {
-              result += log.message;
-            }
-
-            return result;
-          })
-          .join('\n');
-      },
```
sergeyteleshev, line 65: I deleted that and moved to the pure function helper cause I had bugs with `useObservableRef`. It computed those values incorrectly, and the value in the `Editor` component did not update correctly after the data reset/clear.

All in all, Mobx went nuts, and I moved those calculations into the component with `Editor,` and it did the trick.

webapp/packages/plugin-sql-editor/src/SqlResultTabs/OutputLogs/ACTION_CLEAR_OUTPUT_LOGS.ts
```diff
@@ -0,0 +1,14 @@
+/*
+ * CloudBeaver - Cloud Database Manager
+ * Copyright (C) 2020-2025 DBeaver Corp and others
+ *
+ * Licensed under the Apache License, Version 2.0.
+ * you may not use this file except in compliance with the License.
+ */
+import { createAction } from '@cloudbeaver/core-view';
+
+export const ACTION_CLEAR_OUTPUT_LOGS = createAction('action-clear_output_logs', {
```
Wroud: better to rename it like this:
`ACTION_{SCOPE}_NAME` -> `ACTION_OUTPUT_LOG_CLEAR`

webapp/packages/plugin-sql-editor/src/SqlResultTabs/OutputLogs/OutputLogsResource.ts
```diff
@@ -58,6 +58,10 @@ export class OutputLogsResource extends CachedDataResource<IOutputLog[]> {
     this.setData(updatedData);
   }
 
+  reset(): void {
+    this.setData([]);
+  }
```
Wroud: Resources have a `clear` method

Wroud: But this action will clear all logs for all context IDs. I don't think that this is expected behaviour (open two SQL editors for different connections with output log)

webapp/packages/plugin-sql-editor/src/SqlResultTabs/OutputLogs/OutputLogsService.ts
Wroud: yea, i think we need to describe this as a bug, probably it related to `this.useTracker.clear()`

because we have this is the resource's constructor:
```
// hack, we need to call this.use() to initialize resource at startup
    this.useTracker.use(undefined);
```

Technically, we can override the `clear` method in that resource, so it will call `this.useTracker.use(undefined);` after `super.clear()`

webapp/packages/plugin-sql-editor/src/SqlResultTabs/OutputLogs/useOutputLogsPanelState.ts
Wroud: Please revert these changes, instead you need to:
Pass:
```js
{
  outputLogs,
  sqlEditorTabState
}
```
as a third argument (instead of `false`)

add:
```js
      outputLogs: observable.ref,
      sqlEditorTabState: observable.ref,
```
to the second argument

and reference `outputLogs` and `sqlEditorTabState` from computed via `this.`

webapp/packages/plugin-sql-editor/src/SqlResultTabs/OutputLogs/OutputLogsService.ts
```diff
@@ -26,9 +27,24 @@ export class OutputLogsService {
...
 
-  async showOutputLogs(editorState: ISqlEditorTabState): Promise<void> {
+    makeObservable(this, {
+      clearOutputLogs: action.bound,
+      settings: computed,
+      getOutputLogs: action.bound,
+      toggleWrapMode: action.bound,
+      showOutputLogs: action.bound,
+      removeOutputLogsTab: action.bound,
+    });
```
Wroud: I'm not a fan of binding methods every time. I think it's better to bind only those methods that really need it

Moreover, sometimes you don't need `action`, for example: `this.settings.wrapMode = !this.settings.wrapMode`. This is an atomic action, so there is nothing to batch (the action will have no effect there)

`getOutputLogs` can't be an action, this will break observability of data, action only for methods that mutate several observable states, it's not supposed to be used with getters

Wroud (changes requested):

Wroud (changes requested):
Currently, Clear will delete all logs across all SQL editors; we need to delete logs related to the current contextId only

Wroud (approved):

webapp/packages/plugin-sql-editor/src/SqlResultTabs/OutputLogs/OutputLogsResource.ts
```diff
@@ -58,7 +58,16 @@ export class OutputLogsResource extends CachedDataResource<IOutputLog[]> {
     this.setData(updatedData);
   }
 
-  protected async loader(): Promise<IOutputLog[]> {
+  override clear(): void {
+    super.clear();
+    this.useTracker.use(undefined);
```
SychevAndrey, line 63: don't we want to add a comment here with a problem that is solves? Or with link to a ticket. We have the comment above, but clear method is not something related to initialising a resource, imho

SychevAndrey (approved):

webapp/packages/plugin-sql-editor/src/SqlResultTabs/OutputLogs/OutputLogsResource.ts
devnaumov: Agreed, also we have useTracker logic inside clear method

devnaumov (approved):

PR #3548, opened 2025-06-25
dbeaver/pro#6075 remove commit hooks

devnaumov:
closes [6075](https://github.com/dbeaver/pro/issues/6075)

Special flags were added (merge, rebase) to disable pre-commit hooks during the merge and rebase stages. Postinstall script will be executed anyway due to [issue](https://github.com/evilmartians/lefthook/issues/198) so no need to keep it in the package json


webapp/package.json
```diff
@@ -22,8 +22,7 @@
     "validate-dependencies": "core-cli-validate-dependencies './packages/*/'",
     "add-plugin": "core-cli-add-plugin",
     "check-license": "core-cli-check-license",
-    "version:set": "yarn workspaces foreach -Ap --include '@cloudbeaver/product-*' run version:set",
-    "postinstall": "[ -z $CI ] && yarn lefthook install || true"
```
Wroud, line 26: Do we really want to remove it completely, or should we just add a special command so only frontend developers can install it?

sergeyteleshev: Yeah, I suggest to:

1. Rename the command to something like `install-lefthook` so that yarn does not execute this command after installation of the deps
2. Add a VSCode task 
3. Add to the frontend guide that after all installations, we need to execute this task so the hook works for frontend devs

sergeyteleshev: But before we do that, is there any chance to fix some config so the hook works only for frontend-specific files: ts, tsx, CSS, etc., but never throws anything for the backend developers?

sergeyteleshev (approved):

sergeyteleshev:
cool! nice solution

Wroud (approved):

PR #3557, opened 2025-06-27
6056-cb-impelement-ui-kit-select-in-combobox

sergeyteleshev:
closes https://github.com/dbeaver/pro/issues/6056


common-react/@dbeaver/ui-kit/src/Select/SelectField.tsx
```diff
@@ -67,6 +69,10 @@ export interface SelectFieldProps<T, ItemType = SelectItem<T>> {
    * Custom arrow icon React Node that will be rendered instead default one
    */
   arrowIcon?: React.ReactNode;
+
+  store?: SelectProviderProps['store'];
+
+  popoverProps?: Omit<SelectPopoverProps, 'gutter' | 'unmountOnHide'>;
```
Wroud: We should avoid exposing inner 'airakit' interfaces outside of our wrappers

webapp/packages/core-blocks/src/FormControls/Combobox.tsx
```diff
@@ -270,71 +300,26 @@ export const Combobox: ComboboxType = observer(function Combobox({
...
-          onClick={handleClick}
+          disabled={disabled || readOnly}
+          required={rest.required}
+          className={s(styles, { field: true, inline }, className, styles['selectField'])}
+          noItemsPlaceholder={translate('combobox_no_results_placeholder')}
+          selectedRender={selectedRender}
+          arrowIcon={<Icon name="arrow" viewBox="0 0 16 16" className={styles['icon']} />}
+          store={menu}
+          popoverProps={{
+            autoFocusOnShow: !searchable,
```
Wroud: Probably it's better to expose only the `autoFocusOnShow` property instead of the `popoverProps`

common-react/@dbeaver/ui-kit/src/Select/SelectField.tsx
Wroud: But for `store`, probably it's okay...

Wroud (changes requested):

common-react/@dbeaver/ui-kit/src/Select/Select.css
```diff
@@ -80,6 +83,9 @@
     cursor: pointer;
     border: none;
     background: none;
+    overflow: hidden;
+    white-space: nowrap;
+    text-overflow: ellipsis;
```
SychevAndrey: why do we need this changes in the basic Select implementation? Multi-row items are pretty standard, we might want to have it somewhere at some time. We can add this styles to cloudbeaver combobox.

image

webapp/packages/core-blocks/src/FormControls/Combobox.tsx
```diff
@@ -270,71 +300,24 @@ export const Combobox: ComboboxType = observer(function Combobox({
...
-          disabled={disabled || hideMenu}
-          readOnly={readOnly || select}
-          data-focus={focus}
-          data-select={select}
-          className={s(styles, { input: true, select, focus })}
-          onChange={handleChange}
-          onKeyDown={handleKeyDown}
-          onClick={handleClick}
+          disabled={disabled || readOnly}
+          required={rest.required}
```
SychevAndrey: you don't need this because you have {...rest}

webapp/packages/core-blocks/src/FormControls/Combobox.tsx
```diff
@@ -225,41 +204,92 @@ export const Combobox: ComboboxType = observer(function Combobox({
...
+
+  function selectedRender(val: typeof value, item: (typeof items)[number] | undefined): React.ReactNode {
+    if (searchable) {
+      return (
+        <div className={s(styles, { itemSearch: true })} title={title}>
+          {renderIcon(selectedItem)}
+          <input
+            {...rest}
+            ref={setInputRef}
+            required={rest.required}
```
SychevAndrey: already came from {...rest}

common-react/@dbeaver/ui-kit/src/Select/SelectField.tsx
```diff
@@ -115,22 +124,22 @@ export function SelectField<T, ItemType extends {} = SelectItem<T>>({
...
           {displayValue}
           {arrowIcon ?? <Select.Arrow />}
         </Select>
         {description && <span className="dbv-kit-select__description">{description}</span>}
 
-        <SelectPopover gutter={4} unmountOnHide>
+        <SelectPopover autoFocusOnShow={autoFocusItemsOnShow} gutter={4} unmountOnHide>
           {items.length === 0 ? (
-            <div className="dbv-kit-select__empty">No items</div>
+            <div className="dbv-kit-select__empty">{noItemsPlaceholder}</div>
```
SychevAndrey, line 149: this prop is not required, do we need a fallback here? Or maybe make it required to define with l10n on the usage level

SychevAndrey (changes requested):

common-react/@dbeaver/ui-kit/src/Select/SelectField.tsx
sergeyteleshev: We have a fallback here

Screenshot 2025-07-08 at 11 03 54

Wroud (dismissed):

SychevAndrey (dismissed):

SychevAndrey (dismissed):

SychevAndrey (changes requested):
we should not use select from ui-kit in combobox 😢

Wroud (dismissed):

webapp/packages/core-blocks/src/FormControls/Select.tsx
```diff
@@ -0,0 +1,200 @@
...
+        state[name] = id;
+      }
+      if (onSelect && changed) {
+        onSelect(id, name, value);
+      }
+      if (context && changed) {
+        context.change(id, name);
+      }
+    },
+    [value, state, name, menu, context, onSelect],
```
SychevAndrey: since the `value` is in the dependency array, it should also be memoized

devnaumov (dismissed):

webapp/packages/core-blocks/src/FormControls/Select.tsx
```diff
@@ -0,0 +1,200 @@
...
+  );
+
+  useLayoutEffect(() => {
+    onSwitch?.(isOpened);
+  }, [onSwitch, isOpened]);
+
+  function renderIcon(item: (typeof items)[number]): React.ReactNode {
+    if (item && iconSelector && iconSelector(item)) {
```
SychevAndrey: early return will make this function a bit easier to read

webapp/packages/core-blocks/src/FormControls/Select.tsx
```diff
@@ -0,0 +1,200 @@
...
+        items={items}
+        value={value}
+        itemValue={itemValue}
+        itemRender={itemRender}
+        itemDisabled={itemDisabled}
+        name={name}
+        disabled={disabled || readOnly}
+        noItemsPlaceholder={translate('combobox_no_results_placeholder')}
+        selectedRender={selectedRender}
+        arrowIcon={<Icon name="arrow" viewBox="0 0 16 16" className="select__icon" />}
```
SychevAndrey: is it our old chevron icon? Do we really need it? Since we are using thin icons now, the standard one can be more appropriate + less code.

sergeyteleshev: well, it did not work because of this layer of styles. I apply `fill: none` to `svg` but those styles still applies even if it is stroked and mine not
Screenshot 2025-07-15 at 15 40 59

SychevAndrey (dismissed):

Wroud (approved):

PR #3561, opened 2025-07-02
Duplicate translation keys in localization file #3558

### LINKED ISSUE #3558: Duplicate translation keys in localization file
### Description

There are duplicate keys in the translation file:

```
export default [
  ['plugin_browser_update_dialog_title', 'Installing application...'],
  ['plugin_browser_update_dialog_title', 'Installing an update...'],
];

```
This causes the first value ('Installing application...') to be overwritten by the second ('Installing an update...'). As a result, the intended translation for 'Installing application...' is lost.

**Suggested solution**

Use distinct keys, for example:

```
export default [
  ['plugin_browser_install_dialog_title', 'Installing application...'],
  ['plugin_browser_update_dialog_title', 'Installing an update...'],
];
```

**File location:**
[webapp/packages/plugin-browser/src/locales/en.ts](https://github.com/dbeaver/cloudbeaver/blob/devel/webapp/packages/plugin-browser/src/locales/en.ts#L3)

### Steps to reproduce

_No response_

### Expected/Desired Behavior

The localization file should have unique keys for each string. Each key should appear only once with the correct message.

Actual Behavior

The key 'plugin_browser_update_dialog_title' appears twice with different values, causing ambiguity and potential issues in the UI rendering.

### CloudBeaver Version

devel

### Additional context

_No response_

sergeyteleshev:
closes https://github.com/dbeaver/cloudbeaver/issues/3558


sergeyteleshev:
Everything was removed, because it is not used at all + we have no questions about updating/installing the app behavior

**For QA:**
Just to let you know, no tests are required. We simply need to be sure that the app builds fine

Wroud (approved):

devnaumov (approved):

PR #3562, opened 2025-07-02
[CB] SQL Editor - container blinks when changing scheme

sergeyteleshev:
closes https://github.com/dbeaver/pro/issues/6008


webapp/packages/plugin-datasource-context-switch/src/ConnectionSchemaManager/ConnectionSchemaManagerService.ts
```diff
@@ -396,11 +396,11 @@ export class ConnectionSchemaManagerService {
         this.pendingSchemaId = schemaId;
       });
 
+      await this.activeItem.changeSchemaId(schemaId, this.activeItem.context);
```
devnaumov: What I don't like about this is that this code hasn't changed for a long time, yet the bug appeared recently. So we might not be addressing the root cause, but rather just its consequences

Wroud: Agree, it seems that we have changingConnectionContainer and pendingSchemaId specially for that case, but something changed, we need at least an explanation of what happened and how this change addresses it

sergeyteleshev: I pushed a new fix. Seems like Mobx didn't catch changes in execution context updation

sergeyteleshev: I haven't found what commit or PR caused this issue. But I suppose we had this issue for a long time, and some changes just revealed this issue that the updation was not synced with MobX

Wroud (approved):

SychevAndrey (approved):

PR #3570, opened 2025-07-08
dbeaver/pro#5455 Feat: add repeat password field in easyconfig

SychevAndrey:


webapp/packages/plugin-administration/src/ConfigurationWizard/ServerConfiguration/IServerConfigurationFormPartState.ts
```diff
@@ -11,6 +11,7 @@ const ServerConfigurationFormPartStateConfigSchema = schema.object({
   adminCredentialsSaveEnabled: schema.boolean().optional(),
   adminName: schema.string().optional(),
   adminPassword: schema.string().optional(),
+  adminPasswordRepeat: schema.string().optional(),
```
Wroud, line 14: i think we can validate it only on the component level and not add this form state / or form part checks

SychevAndrey: It will create a bug that will let the user finish the form without this field. So, if we don't want to validate it on the wizard level, it's better to close the ticket w/o additional field.

Wroud: hm, we need to check if the wizard can be finished after a page refresh (fill passwords, then go to the next step, and make a refresh, check that all steps back/forward work)

SychevAndrey: We clear password fields after refresh. I believe it's the right logic. Why should it work after refresh?

Wroud: If we clear it, then okay. The password can be remembered (between refreshes), and validation will not pass if the repeated password is not filled in.

sergeyteleshev: Adding `passwordRepeat` to the part state breaks the pattern of the form. Here should be only parameters we want to send/get to/from the server
Custom input validation via hook should do a thing and prevent the form from saving
It worked in couple of other places (user info, user profile)
So definitely something went wrong in this form and we need to figure out what happened and how we can treat it

Wroud: As I understand, there is a problem with easy-config, we have next/back buttons there, and they probably don't trigger frontend validation?

SychevAndrey: https://github.com/user-attachments/assets/fb99fe66-7413-4d4a-beb8-6b6f79910f11

Generally, yes. It works sometimes, but we always have an ability to navigate via menu on the left without additional validation.

devnaumov: It seems there is an issue with the video. The password field had validation in place and the validation was triggered every time the user clicked the 'Next' button. If it is regression, we should fix it too

devnaumov: you shouldn't be able to navigate through the left menu, as the step remains disabled until the validation on the current tab is successfully completed

devnaumov: https://github.com/user-attachments/assets/fb2e9c4d-5756-4973-aa9a-857447eafa5f

devnaumov: Maybe we can subscribe to the part submit or validate event from AdminForm component and interrupt execution if passwords dont match

Wroud (approved):

PR #3577, opened 2025-07-08
dbeaver/pro#6150 sync options part

devnaumov:
closes [6150](https://github.com/dbeaver/pro/issues/6150)

Two additional bugs were fixed as part of this ticket:

If custom properties were defined and any changes were made to the connection options without visiting the Properties tab, all custom properties were deleted.

If custom properties were created and the driver was changed, the properties remained persistent across drivers.

### DIFF
--- a/webapp/packages/plugin-connections/src/ConnectionForm/DriverProperties/DriverProperties.tsx
+++ b/webapp/packages/plugin-connections/src/ConnectionForm/DriverProperties/DriverProperties.tsx
@@ -9,7 +9,7 @@ import { computed, observable, runInAction } from 'mobx';
 import { observer } from 'mobx-react-lite';
 import { useMemo, useState } from 'react';
 
-import { ColoredContainer, Group, type IProperty, PropertiesTable, s, useAutoLoad, useResource, useS } from '@cloudbeaver/core-blocks';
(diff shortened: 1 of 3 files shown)


webapp/packages/plugin-connections/src/ConnectionForm/DriverProperties/DriverProperties.tsx
```diff
@@ -41,7 +42,20 @@ export const DriverProperties: TabContainerPanelComponent<IConnectionFormProps>
...
+    return { propertiesList, add, remove, reset };
+  });
+
+  useExecutor({
+    executor: optionsPart.onDriverIdChange,
+    handlers: [
+      function handleDriverChange() {
+        state.reset();
+      },
+    ],
```
Wroud, line 58: This may not work if the component is unmounted (when the options tab is active)

Wroud (approved):

webapp/packages/plugin-connections/src/ConnectionForm/DriverProperties/DriverProperties.tsx
sergeyteleshev: Agree, seems like it is better to do it inside the part

devnaumov: We dont have access to this state in part, it exists only on the component level, because of this we dont want to run it if the component is ummounted

PR #3578, opened 2025-07-08
dbeaver/pro#6201 feat: ai chat query insertion

Wroud:
closes dbeaver/pro#6201


webapp/packages/plugin-sql-editor/src/SqlDataSource/BaseSqlDataSource.ts
```diff
@@ -153,6 +153,10 @@ export abstract class BaseSqlDataSource implements ISqlDataSource {
   }
 
   setScript(script: string, source?: string, cursor?: ISqlEditorCursor): void {
+    if (cursor) {
+      // TODO: should we call update?
```
sergeyteleshev: after setScript executes this.onUpdate according to line 106 of this file

webapp/packages/plugin-sql-editor/src/SqlEditor/useActiveQuery.ts
```diff
@@ -0,0 +1,87 @@
...
+          // TODO: getResolvedSegment will return segment without end delimiter
+          //       we need this to avoid duplicated end delimiter
+          const dialectDelimiter = sqlEditorData.dialect?.scriptDelimiter ?? '';
+          const endDelimiterLength = dialectDelimiter.length;
+          const endDelimiter = sqlEditorData.value.slice(segment.end, segment.end + endDelimiterLength);
+
+          if (endDelimiter === dialectDelimiter && query.endsWith(endDelimiter)) {
+            query = query.slice(0, -endDelimiterLength);
+          }
+        }
```
sergeyteleshev, line 57: would be nice to move it to pure function like `getQueryWithoutDelimiter`

sergeyteleshev (dismissed):

webapp/packages/plugin-sql-editor/src/SqlEditor/useSqlEditor.ts
```diff
@@ -435,24 +435,35 @@ export function useSqlEditor(state: ISqlEditorTabState): ISQLEditorData {
         const projectId = this.dataSource?.executionContext?.projectId;
         const connectionId = this.dataSource?.executionContext?.connectionId;
 
-        await data.updateParserScripts();
+        while (true) {
```
sergeyteleshev, line 438: all together seems dangerous to me
i suppose we should find some time to refactor it so all logic with the cursor more reliable. as for me now it is too tricky

webapp/packages/plugin-sql-editor/src/SqlEditor/useActiveQuery.ts
Wroud: Yes, it will be, but I want to wait for one more case with this logic (we had in our ticket about VQB sync) to make a proper generalization. So we will know what we have in common in both places and what we can move to a shared function

webapp/packages/plugin-sql-editor/src/SqlEditor/useSqlEditor.ts
Wroud: Agree, i will create a tech-dept ticket for that

SychevAndrey (dismissed):

sergeyteleshev (dismissed):

devnaumov (dismissed):

SychevAndrey (approved):

webapp/packages/plugin-sql-editor/src/SqlDataSource/LocalStorage/LocalStorageSqlDataSource.ts
```diff
@@ -76,6 +79,11 @@ export class LocalStorageSqlDataSource extends BaseSqlDataSource {
     return true;
   }
 
+  // TODO: should we move it to the BaseSqlDataSource?
```
devnaumov, line 82: If this logic could be applied to all datasources, why not?

webapp/packages/plugin-sql-editor/src/SqlEditor/useActiveQuery.ts
```diff
@@ -0,0 +1,87 @@
...
+        if (segment) {
+          query = query.trim();
+          // TODO: getResolvedSegment will return segment without end delimiter
+          //       we need this to avoid duplicated end delimiter
+          const dialectDelimiter = sqlEditorData.dialect?.scriptDelimiter ?? '';
+          const endDelimiterLength = dialectDelimiter.length;
+          const endDelimiter = sqlEditorData.value.slice(segment.end, segment.end + endDelimiterLength);
+
+          if (endDelimiter === dialectDelimiter && query.endsWith(endDelimiter)) {
+            query = query.slice(0, -endDelimiterLength);
```
devnaumov, line 55: "-" before endDelimiterLength seems unclear, what are we trying to do here?  Could it be rewritten like this?

`query.slice(0, query.length - endDelimiterLength)`

devnaumov (approved):

webapp/packages/plugin-sql-editor/src/SqlEditor/useActiveQuery.ts
Wroud: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/slice#end
This is its behaviour described in the documentation, I think it's clear enough

webapp/packages/plugin-sql-editor/src/SqlDataSource/LocalStorage/LocalStorageSqlDataSource.ts
Wroud: This is the point. Can it be applied to SQL files? I think we need more usage examples to decide if we want to make it general or local

So, for now, we have it only in this data source, and in others, we don't need to think about it at all

sergeyteleshev (approved):

webapp/packages/plugin-sql-editor/src/SqlEditor/useSqlEditor.ts
Wroud: https://github.com/dbeaver/pro/issues/6327

PR #3583, opened 2025-07-14
dbeaver/pro#6105 fix: form validation fails on test in GCP

SychevAndrey:
Removed active state from Options component resources because it unload resources when we switch to other tabs which leads to broken connection form state and entails data inconsistency.

For example, the check driver?.useCustomPage fails because driver resource has been unloaded and the component renders standard Parameters form instead of custom taking state from the wrong destination.

### DIFF
--- a/webapp/packages/plugin-connections/src/ConnectionForm/Options/Options.tsx
+++ b/webapp/packages/plugin-connections/src/ConnectionForm/Options/Options.tsx
@@ -48,7 +48,7 @@ import { useService } from '@cloudbeaver/core-di';
 import { ProjectInfoResource } from '@cloudbeaver/core-projects';
 import { EAdminPermission, ServerConfigResource } from '@cloudbeaver/core-root';
 import { DriverConfigurationType } from '@cloudbeaver/core-sdk';
-import { type TabContainerPanelComponent, TabsContext, useTab } from '@cloudbeaver/core-ui';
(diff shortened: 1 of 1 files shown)


webapp/packages/plugin-connections/src/ConnectionForm/Options/Options.tsx
```diff
@@ -86,27 +86,18 @@ const driverConfiguration: IDriverConfiguration[] = [
 ];
 
 export const Options: TabContainerPanelComponent<IConnectionFormProps> = observer(function Options({ formState, tabId }) {
-  const { selected } = useTab(tabId);
   const isAdmin = usePermission(EAdminPermission.admin);
-  const serverConfigResource = useResource(Options, ServerConfigResource, undefined, {
```
devnaumov, line 91: This boolean ensures that if the tab is not selected, the related logic is not loaded and no resources are triggered, which is the seems as expected behavior. Could you please explain how does this affect bug?

devnaumov (approved):

Wroud (approved):

webapp/packages/plugin-connections/src/ConnectionForm/Options/Options.tsx
SychevAndrey: please read commit message

sergeyteleshev (approved):

PR #3585, opened 2025-07-15
dbeaver/pro#5241 [CB] Screen reader does not read some fields in the application

SychevAndrey:
Fixed input component, linked controls with labels in field via for attribute.
This approach doesn't work for 100% in all assistive software, that's why I also defined aria-label for inputs as well.


sergeyteleshev (dismissed):

webapp/packages/core-blocks/src/FormControls/Textarea.tsx
```diff
@@ -102,12 +103,14 @@ export const Textarea: TextareaType = observer(function Textarea({
...
     <Field {...layoutProps} className={s(styles, { field: true, embedded }, className)}>
-      <FieldLabel className={s(styles, { fieldLabel: true })} title={labelTooltip || rest.title} required={required}>
+      <FieldLabel htmlFor={inputId} className={s(styles, { fieldLabel: true })} title={labelTooltip || rest.title} required={required}>
         {children}
       </FieldLabel>
       <textarea
         {...rest}
         ref={textareaRef}
+        id={inputId}
+        aria-label={children?.toString()}
```
Wroud: Should we change the type of children to "string"?

webapp/packages/core-blocks/src/FormControls/InputField/InputFieldBase.tsx
```diff
@@ -116,13 +117,15 @@ export const InputFieldBase = observer<InputFieldBaseProps, HTMLInputElement>(
...
-        <FieldLabel title={labelTooltip || rest.title} className={s(styles, { fieldLabel: true })} required={required}>
+        <FieldLabel htmlFor={inputId} title={labelTooltip || rest.title} className={s(styles, { fieldLabel: true })} required={required}>
           {children}
         </FieldLabel>
         <div className={s(styles, { inputContainer: true })}>
           <Input
             ref={mergedRef}
             {...rest}
+            id={inputId}
+            aria-label={children?.toString() || rest['aria-label'] || rest.title}
```
Wroud: Should we change the type of children to "string"?

Wroud: or maybe it's better to remove `children?.toString()`

webapp/packages/core-blocks/src/FormControls/Textarea.tsx
Wroud: or maybe it's better to remove `children?.toString()`

webapp/packages/core-blocks/src/FormControls/InputField/InputFieldBase.tsx
SychevAndrey: actually I don't like a child here at all because in HTML inputs don't accept children. It's better to have a prop like 'label' or something. Or we can change the type to string, because I can't imagine a situation when we can use an image in input label or link or smth like that.

webapp/packages/core-blocks/src/FormControls/Textarea.tsx
SychevAndrey: same as for input, but in textarea case, the situation is even worse, because in spec, textarea's children = value.

Wroud (dismissed):

Wroud (dismissed):

Wroud (approved):

devnaumov (approved):

sergeyteleshev (approved):

PR #3586, opened 2025-07-15
Dbeaver/pro#5630 add cache for queries

HocKu7:
Closes https://github.com/dbeaver/pro/issues/5630


sergeyteleshev (dismissed):

webapp/packages/plugin-sql-editor/src/QueryDataSource.ts
```diff
@@ -185,9 +185,13 @@ export class QueryDataSource<TOptions extends IDataQueryOptions = IDataQueryOpti
     const limit = this.count;
 
     let firstResultId: string | undefined;
+    let useServerCache = false;
 
     if (this.requestInfo.query === this.options?.query) {
       firstResultId = this.getPreviousResultId(prevResults, executionContextInfo);
+      if (firstResultId !== undefined && !refresh) {
```
devnaumov, line 192: This logic is duplicated in ContainerDataSource, if it is possible, its better to move it to the separate method, so if we made some changes in one data source, we will not forget about the second one

devnaumov (dismissed):

sergeyteleshev (dismissed):

PR #3597, opened 2025-07-22
dbeaver/pro#6283 Add Combobox component

SychevAndrey:
https://github.com/user-attachments/assets/7a3ce264-cdd6-4154-b2b8-7760488a5a7c


common-react/@dbeaver/ui-kit/src/Combobox/Combobox.tsx
```diff
@@ -0,0 +1,257 @@
...
+    if (isObjectSearch) {
+      return itemsList.map(value => {
+        const data = itemsData.get(value) || {};
+        return { value, ...data };
+      });
+    }
+    return itemsList;
+  }, [itemsList, itemsData, isObjectSearch]);
+
+  const fuse = useMemo(() => new Fuse(searchData, searchOptions), [searchData, searchOptions]);
```
sergeyteleshev: I suppose we can move `useSearch` and `useFuzzy` search to the `ui-kit`, and the logic with `useObservableRef` rewrite to react `useState`, so you don't need to move it also to ui kit and do a huge amount of import changes

sergeyteleshev: We already have a search in the app that works fine. It would be easier to support, and `mini-search` covers all of our needs right now
plus Combobox would support a few types of searches depending on the situation

common-react/@dbeaver/ui-kit/src/Combobox/Combobox.tsx
```diff
@@ -0,0 +1,257 @@
...
+export function SearchContextProvider({
+  children,
+  searchOptions = {
+    includeScore: false,
+    threshold: 0.3,
+  },
+  defaultValue,
+}: SearchContextProviderProps) {
+  const [registeredItems, setRegisteredItems] = useState<Set<string>>(new Set());
+  const [itemsData, setItemsData] = useState<Map<string, Record<string, any>>>(new Map());
```
sergeyteleshev: seems like `useSearch` can cover the case with multiple fields to search in 1 item. so this would probably gone after `useSearch` implementation

webapp/packages/core-blocks/src/FormControls/Combobox.tsx
```diff
@@ -65,9 +55,6 @@ export interface ComboboxType {
   <TValue, TKey extends keyof TState, TState>(props: ObjectProps<TValue, TKey, TState>): React.JSX.Element;
 }
 
-{
-  /* TODO rewrite whole component to select attribute instead of input type text so it has an okay form validation */
-}
 export const Combobox: ComboboxType = observer(function Combobox({
```
sergeyteleshev, line 67: please double check that selects and comboboxes are fine after the merge of this. right now all selects are comboboxes
https://github.com/dbeaver/pro/issues/6056

common-react/@dbeaver/ui-kit/src/Combobox/Combobox.tsx
```diff
@@ -0,0 +1,257 @@
...
+
+/**
+ * ComboboxItem - An option in the combobox popover
+ * Automatically registers/unregisters itself for search
+ */
+export function ComboboxItem({ value, searchData, ref, ...props }: ComboboxItemProps & { ref?: React.Ref<HTMLDivElement> }) {
+  const { registerItem, unregisterItem, isItemVisible } = use(SearchContext);
+
+  useLayoutEffect(() => {
+    if (value == null) return;
```
devnaumov:
```
if(value) {
...
}
```

Also, please use strict comparison, we want to avoid unexpected type conversion

webapp/packages/plugin-connections/src/ConnectionForm/Options/Options.tsx
```diff
@@ -192,7 +192,6 @@ export const Options: TabContainerPanelComponent<IConnectionFormProps> = observe
                   valueSelector={driver => driver.name ?? ''}
                   titleSelector={driver => driver.description}
                   iconSelector={driver => driver.icon}
-                  searchable={drivers.length > 10}
```
devnaumov: Seems like these changes will break logic that if drivers > 10 we can search otherwise we dont?

SychevAndrey: all comboboxes have search now, we don't have this option anymore.

sergeyteleshev: yeah, it is fine. we decided to split comboboxes and selects

common-react/@dbeaver/ui-kit/src/Combobox/Combobox.tsx
```diff
@@ -0,0 +1,257 @@
...
+  },
+  defaultValue,
+}: SearchContextProviderProps) {
+  const [registeredItems, setRegisteredItems] = useState<Set<string>>(new Set());
+  const [itemsData, setItemsData] = useState<Map<string, Record<string, any>>>(new Map());
+  const store = useComboboxContext();
+  const searchValue = useStoreState(store, 'value') || '';
+  const deferredValue = useDeferredValue(searchValue);
+
+  const isObjectSearch = useMemo(() => Array.isArray(searchOptions.keys) && searchOptions.keys.length > 0, [searchOptions.keys]);
```
Wroud: useMemo is extra here, please omit it

common-react/@dbeaver/ui-kit/src/Combobox/Combobox.tsx
Wroud: Is it possible that `registeredItems` and `itemsData` will have different sets of elements? Can we keep only `itemsData`?

common-react/@dbeaver/ui-kit/src/Combobox/Combobox.tsx
```diff
@@ -0,0 +1,257 @@
...
+
+  const contextValue = useMemo(
+    () => ({
+      matches,
+      registerItem,
+      unregisterItem,
+      isItemVisible,
+    }),
+    [matches, registerItem, unregisterItem, isItemVisible],
+  );
```
Wroud: Memoization is not optimized here because any change will likely trigger data invalidation.
In such cases, it is better to determine "source" data that will be changed and "computed" data that needs to be recalculated.

So basically, you can have a single `useMemo` that will depend on `fuse`, `deferredValue`, `itemsList`, `defaultValue`, `isObjectSearch`, and will return the complete context. Right now, changing any incoming values will lead to a cascade re-calculation of most of the memorized values.

common-react/@dbeaver/ui-kit/src/Combobox/Combobox.tsx
```diff
@@ -0,0 +1,257 @@
...
+}
+
+/**
+ * ComboboxInput - Basic input component that handles search value and blur logic
+ */
+export const ComboboxInput = ({ onBlur, onKeyDown, defaultValue, ref, ...props }: ComboboxProps & { ref?: React.Ref<HTMLInputElement> }) => {
+  const store = useComboboxContext();
+  const { matches } = use(SearchContext);
+
+  const handleBlur = useCallback(
```
Wroud: Probably you can omit `useCallback` here

common-react/@dbeaver/ui-kit/src/Combobox/Combobox.tsx
```diff
@@ -0,0 +1,257 @@
...
+    (event: FocusEvent<HTMLInputElement>) => {
+      const selectedValue = store?.getState()?.selectedValue;
+      const value = store?.getState()?.value;
+
+      if (value !== undefined && !matches.includes(value)) {
+        if (selectedValue) {
+          store?.setValue(selectedValue as string);
+        } else if (defaultValue) {
```
Wroud: Better to check for `!== undefined` for `selectedValue` and `defaultValue`

common-react/@dbeaver/ui-kit/src/Combobox/Combobox.tsx
```diff
@@ -0,0 +1,257 @@
...
+          store?.setValue(defaultValue);
+        }
+      } else if (value && matches.includes(value)) {
+        store?.setSelectedValue?.(value);
+      }
```
Wroud:
```suggestion
      if (value !== undefined) {
        if (matches.includes(value)) {
          store?.setSelectedValue?.(value);
        } else {
          const nextValue = selectedValue ?? defaultValue;
          if (nextValue !== undefined) {
            store?.setValue(nextValue as string);
          }
        }
      }
```

common-react/@dbeaver/ui-kit/src/Combobox/Combobox.tsx
```diff
@@ -0,0 +1,257 @@
...
+  matches: string[];
+  registerItem: (value: string, searchData?: Record<string, any>) => void;
+  unregisterItem: (value: string) => void;
+  isItemVisible: (value: string) => boolean;
+}
+
+const SearchContext = createContext<SearchContextValue>({
+  matches: [],
+  registerItem: () => {},
+  unregisterItem: () => {},
```
Wroud: Seems like this can be covered by `useComboboxStore`

Wroud (changes requested):

common-react/@dbeaver/ui-kit/src/Combobox/Combobox.tsx
SychevAndrey: It's been decided to remove fuzzy search functionality from the Combobox component for now.

SychevAndrey:
The component was significantly simplified, for now it doesn't use advanced search mechanisms, we might add it later.
Also made changes in all UIKit components wrapping all micro-animations in media queries in order to respect the user's reduce motion settings.

SychevAndrey:
https://github.com/user-attachments/assets/b2eca64f-18f3-4737-92ca-1845b7edff24

common-react/@dbeaver/ui-kit/src/Combobox/Combobox.tsx
```diff
@@ -0,0 +1,230 @@
...
+export function ComboboxPopover({ children, className, ...props }: ComboboxPopoverProps) {
+  return (
+    <AriaComboboxPopover gutter={8} portal sameWidth unmountOnHide className={clsx('dbv-kit-combobox__popover', className)} {...props}>
+      {children}
+    </AriaComboboxPopover>
+  );
+}
+
+export interface ComboboxItemProps extends AriaComboboxItemProps {
+  searchData?: Record<string, any>;
```
Wroud: It is kind of strange to have `searchData` on each item and `searchFields` because `searchData` sounds like data only for search, not a generic type.

I will recommend removing this logic for now, mainly because we don't have use cases for it right now.

common-react/@dbeaver/ui-kit/src/Combobox/Combobox.tsx
```diff
@@ -0,0 +1,230 @@
...
+  defaultValue?: string;
+  valueFormatter?: (value: string) => string;
+}
+
+/**
+ * ComboboxInput - Basic input component that handles search value and blur logic
+ * Restores the selected value if the input value is not in the items list
+ * If only one item is available, it selects that item automatically
+ */
+export const ComboboxInput = ({
```
Wroud: Please don't use anonymous functions

common-react/@dbeaver/ui-kit/src/Combobox/Combobox.tsx
```diff
@@ -0,0 +1,230 @@
...
+      searchOptions,
+      defaultValue,
+    }),
+    [deferredValue, searchOptions, defaultValue],
+  );
+
+  return <SearchContext value={contextValue}>{children}</SearchContext>;
+}
+
+export interface ComboboxProps extends AriaComboboxProps {
```
Wroud: ComboboxInputProps

common-react/@dbeaver/ui-kit/src/Combobox/Combobox.css
```diff
@@ -0,0 +1,101 @@
...
+    outline-offset: var(--dbv-kit-control-outline-offset);
+    outline-color: var(--dbv-kit-control-outline-color);
+  }
+
+  &:hover {
+    border-color: var(--dbv-kit-color-control-hover);
+  }
+
+  &[aria-disabled='true'] {
+    opacity: var(--dbv-kit-control-disabled-opacity);
```
sergeyteleshev: please check that offset works
may be it should be put here. we have checked that together previously quickly, but it would be nice to double check it so it is fine
`outline-offset: var(--dbv-kit-control-outline-offset);`

common-react/@dbeaver/ui-kit/src/Combobox/Combobox.css
```diff
@@ -0,0 +1,101 @@
...
+    max-height: var(--dbv-kit-combobox-popover-height);
+    overflow-y: auto;
+    background-color: var(--dbv-kit-combobox-popover-background);
+    color: var(--dbv-kit-combobox-popover-foreground);
+    border-color: var(--dbv-kit-combobox-border-color);
+    border-radius: var(--dbv-kit-combobox-border-radius);
+    border-width: var(--dbv-kit-combobox-border-width);
+    border-style: var(--dbv-kit-combobox-border-style);
+    box-shadow: var(--dbv-kit-combobox-popover-shadow);
+    outline: 0;
```
sergeyteleshev, line 49: may be focus with outline opened popover? so user see and intuitively know he can use a keyboard to select items

webapp/packages/core-blocks/src/FormControls/Combobox.tsx
```diff
@@ -87,245 +85,119 @@ export const Combobox: ComboboxType = observer(function Combobox({
...
-  const focus = menu.visible;
-
-  if (loading && items.length === 0) {
-    inputValue = translate('ui_processing_loading');
-  }
+  const comboboxDefaultValue = selectedItem ? valueSelector(selectedItem) : defaultValue ? String(defaultValue) : undefined;
 
   return (
-    <Field {...layoutProps} className={s(styles, { field: true, inline }, className)}>
+    <Field {...layoutProps} className={className} style={{ display: inline ? 'flex' : 'block', alignItems: inline ? 'center' : undefined }}>
```
sergeyteleshev: these styles can be implemented via tailwind

sergeyteleshev (dismissed):

common-react/@dbeaver/ui-kit/src/Combobox/Combobox.tsx
```diff
@@ -0,0 +1,230 @@
...
+  defaultValue?: string;
+  valueFormatter?: (value: string) => string;
```
Wroud: rename to `getInputFromItem?: (value: T) => string;`

common-react/@dbeaver/ui-kit/src/Combobox/Combobox.tsx
Wroud: Let's change it to `item?: T`

common-react/@dbeaver/ui-kit/src/Combobox/Combobox.css
SychevAndrey: Combobox component looks exactly like Select component does. Maybe we can discuss it later on some UX/UI call

common-react/@dbeaver/ui-kit/src/Combobox/Combobox.tsx
```diff
@@ -0,0 +1,191 @@
...
+
+    if (typeof selectedValue !== 'string' || (value === selectedValue && !getInputFromItem)) {
+      return;
+    }
+
+    const nextValue = (selectedValue || defaultValue) as string;
+
+    if (!nextValue) {
+      return;
+    }
```
Wroud:
```suggestion
    const nextValue = (selectedValue || defaultValue) as string | undefined;

    if (nextValue === undefined) {
      return;
    }
```

common-react/@dbeaver/ui-kit/src/Combobox/Combobox.tsx
```diff
@@ -0,0 +1,191 @@
...
+  type ComboboxItemProps as AriaComboboxItemProps,
+  useComboboxContext,
+  useComboboxStore,
+  useStoreState,
+  type ComboboxStoreState,
+} from '@ariakit/react';
+import clsx from 'clsx';
+import { type HTMLAttributes, type Ref, useCallback } from 'react';
+import './Combobox.css';
+
```
Wroud, line 27: \+
```ts
interface ComboboxStoreItemWithItem<T> extends ComboboxStoreItem {
  item: T
}
```

common-react/@dbeaver/ui-kit/src/Combobox/Combobox.tsx
```diff
@@ -0,0 +1,191 @@
...
+    }
+
+    const nextValue = (selectedValue || defaultValue) as string;
+
+    if (!nextValue) {
+      return;
+    }
+
+    if (getInputFromItem) {
+      const nextItem = (items?.find(item => item.value === nextValue) as ComboboxStoreState['items'][number] & { item: T })?.item;
```
Wroud: `as ComboboxStoreItemWithItem<T> | undefined`

common-react/@dbeaver/ui-kit/src/Combobox/Combobox.tsx
```diff
@@ -0,0 +1,191 @@
...
+  useComboboxContext,
+  useComboboxStore,
+  useStoreState,
+  type ComboboxStoreState,
+} from '@ariakit/react';
+import clsx from 'clsx';
+import { type HTMLAttributes, type Ref, useCallback } from 'react';
+import './Combobox.css';
+
+export interface ComboboxInputProps<T> extends AriaComboboxProps {
```
Wroud:
```ts
export interface ComboboxInputProps<T> extends Omit<AriaComboboxProps, 'autoSelect'>
```

common-react/@dbeaver/ui-kit/src/Combobox/Combobox.tsx
```diff
@@ -0,0 +1,191 @@
...
+  item,
+  ref,
+  getInputFromItem,
+  ...props
+}: ComboboxItemProps<T> & { ref?: React.Ref<HTMLDivElement> }) {
+  const store = useComboboxContext();
+  const searchValue = useStoreState(store, 'value') || '';
+  const selectedValue = useStoreState(store, 'selectedValue') || '';
+
+  const targetText = [value ?? '', item && getInputFromItem ? getInputFromItem(item) : ''].filter(Boolean).join(' ');
```
Wroud: I'm not sure that we need to filter by value if we have `item and getInputFromItem` (the user will filter by the text that he sees)

common-react/@dbeaver/ui-kit/src/Combobox/Combobox.tsx
```diff
@@ -0,0 +1,191 @@
...
+  const isVisible = searchValue === selectedValue || !searchValue.trim() || targetText.toLowerCase().includes(searchValue.trim().toLowerCase());
+  const getItem = useCallback((data: any) => ({ ...data, item }), []);
```
Wroud: `item` should be in the hook deps

common-react/@dbeaver/ui-kit/src/Combobox/Combobox.tsx
```diff
@@ -0,0 +1,191 @@
...
+
+  function setValueOnClick(event: React.MouseEvent<HTMLDivElement>) {
+    if (getInputFromItem && item) {
+      const nextValue = getInputFromItem(item);
+      store!.setValue(nextValue);
+      store!.setSelectedValue(nextValue);
```
Wroud:
```suggestion
      store!.setValue(getInputFromItem(item));
      store!.setSelectedValue(value);
```

webapp/packages/core-blocks/src/FormControls/Combobox.tsx
```diff
@@ -87,245 +85,117 @@ export const Combobox: ComboboxType = observer(function Combobox({
...
-    };
-  }, [inputRef]);
-
   const icon = selectedItem && iconSelector?.(selectedItem);
-  const focus = menu.visible;
-
-  if (loading && items.length === 0) {
-    inputValue = translate('ui_processing_loading');
-  }
+  const comboboxDefaultValue = selectedItem ? valueSelector(selectedItem) : defaultValue ? String(defaultValue) : undefined;
```
Wroud:
```suggestion
  const comboboxDefaultSelectedValue = defaultValue;
  let comboboxDefaultValue = undefined;
  
  if (comboboxDefaultSelectedValue !== undefined) {
    const defaultItem = items.find((item, index) => keySelector(item, index) === comboboxDefaultSelectedValue);
    
    if (defaultItem) {
      comboboxDefaultValue = valueSelector(defaultItem);
    }
  }
```

webapp/packages/core-blocks/src/FormControls/Combobox.tsx
```diff
@@ -87,245 +85,117 @@ export const Combobox: ComboboxType = observer(function Combobox({
...
-        />
-        <MenuButton {...menu} disabled={readOnly || disabled || hideMenu} className={styles['menuButton']}>
-          <Icon name="arrow" viewBox="0 0 16 16" className={s(styles, { icon: true, focus })} />
-        </MenuButton>
-        <Menu {...menu} ref={menuRef} aria-label={propertyName} className={s(styles, { menu: true })} modal>
-          {!filteredItems.length ? (
-            <MenuItem id="placeholder" disabled {...menu} className={s(styles, { menuItem: true })}>
-              {translate('combobox_no_results_placeholder')}
-            </MenuItem>
+      <ComboboxProvider defaultValue={comboboxDefaultValue} defaultSelectedValue={comboboxDefaultValue} setSelectedValue={handleSelect}>
```
Wroud:
```suggestion
      <ComboboxProvider defaultValue={comboboxDefaultValue} defaultSelectedValue={comboboxDefaultSelectedValue} setSelectedValue={handleSelect}>
```

Wroud (changes requested):

webapp/packages/core-blocks/src/FormControls/Combobox.tsx
```diff
@@ -87,245 +83,167 @@ export const Combobox: ComboboxType = observer(function Combobox({
...
 
-  const handleChange = useCallback(
-    (event: React.ChangeEvent<HTMLInputElement>) => {
-      const value = event.target.value;
-      onChange(value, name);
-      setSearchValue(value);
-    },
-    [name, onChange],
+  const selectedItem = items.find((item, index) => keySelector(item, index) === selectedKey);
+  const [inputValue, setInputValue] = useState<string>(() => (selectedItem ? valueSelector(selectedItem) : ''));
```
Wroud: It's easier to:
make inputValue null (or undefined) by default (in useState)
before passing this value to the combobox, do: `inputValue ?? selectedValue`
where `selectedValue = selectedItem ? valueSelector(selectedItem) : ''`

So, when resetting to the currently selected value, you just set the state to null or undefined.

This will prevent synchronization problems when selectedItem changes.

This will also simplify the `isVisible` check: `inputValue === valueSelector(selectedItem)` => `inputValue === null`

webapp/packages/core-blocks/src/FormControls/Combobox.tsx
```diff
@@ -87,245 +83,167 @@ export const Combobox: ComboboxType = observer(function Combobox({
...
 
-  const handleKeyDown = useCallback(
-    (event: React.KeyboardEvent<HTMLInputElement>) => {
-      if (event.key === 'Enter') {
-        event.preventDefault();
-        matchItems(true);
-      }
+      setTimeout(() => {
+        isSelectingRef.current = false;
+      }, 100);
```
Wroud: it seems that this code will not be executed in some cases because of early returns:
```ts
      if (!item || selectedValue === selectedKey) {
        isSelectingRef.current = false;
        return;
      }
```

Wroud: The best way for such things is to use `try / finally` because this will also work with exceptions

Wroud (changes requested):

webapp/packages/core-blocks/src/FormControls/Combobox.tsx
```diff
@@ -87,245 +83,152 @@ export const Combobox: ComboboxType = observer(function Combobox({
...
-      setSearchValue(null);
-    },
-    [value, state, name, menu, context, onSelect],
+  const selectedItem = items.find((item, index) => keySelector(item, index) === selectedKey);
+  const [inputValue, setInputValue] = useState<string | null>(null);
+
+  const selectedValue = selectedItem ? valueSelector(selectedItem) : '';
+  const displayValue = inputValue ?? selectedValue;
+
+  const filteredItems = getComputed(() =>
```
Wroud: `getComputed` will not work as expected for not primitive values, for arrays you will need to specify a custom equality function (I recommend removing `getComputed`)

Wroud (dismissed):

Wroud (dismissed):

sergeyteleshev (dismissed):

Wroud (approved):

PR #3624, opened 2025-07-31
Dbeaver/pro#2951 Show message about disabling user

vkirst:


ShadelessFox (dismissed):

E1izabeth (dismissed):

server/bundles/io.cloudbeaver.server.ce/src/io/cloudbeaver/service/session/CBSessionManager.java
```diff
@@ -407,6 +408,20 @@ public void closeUserSession(@NotNull WSUserDeletedEvent userDeletedEvent) {
         }
     }
 
+    public void closeUserSession(@NotNull WSUserDisabledEvent userDisabledEvent) {
```
yagudin10: Can we use same method as for deleting user?

server/bundles/io.cloudbeaver.server.ce/src/io/cloudbeaver/server/events/WSUserEventHandler.java
```diff
@@ -42,6 +43,11 @@ public void handleEvent(@NotNull EVENT event) {
                     sessionManager.closeUserSession(userDeletedEvent);
                 }
                 break;
+            case WSUserDisabledEvent.ID:
+                if (event instanceof WSUserDisabledEvent userDisabledEvent) {
```
yagudin10, line 61: maybe we should think about using the same class as in the deleted event or the same interface

yagudin10 (changes requested):

labanovichttps (approved):

PR #3628, opened 2025-08-04
[CB] Bump NPM dependencies

sergeyteleshev:
Removed zod translation packages, cause zod 4 supports it now - https://zod.dev/error-customization#internationalization. 
Also, they added `z.stringbool()`,  so I replaced our solution with theirs - https://zod.dev/v4?id=stringbool

closes https://github.com/dbeaver/pro/issues/5764


webapp/packages/core-administration/src/AdministrationScreen/AdministrationScreenService.ts
```diff
@@ -267,7 +267,7 @@ export class AdministrationScreenService {
       return true;
     }
 
-    return this.administrationItemService.canDeActivate(screen, toScreen, this.isConfigurationMode, screen.item !== toScreen?.item);
+    return await this.administrationItemService.canDeActivate(screen, toScreen, this.isConfigurationMode, screen.item !== toScreen?.item);
```
SychevAndrey: why do we need return await here?

webapp/packages/plugin-sql-editor/src/SqlEditorSettingsService.ts
```diff
@@ -77,11 +77,15 @@ const defaultSettings = schema.object({
       switch (value) {
         case 'DEFAULT':
           return 'DEFAULT';
+        case 'COMBINE':
```
SychevAndrey: i'm not sure why we adding this in update lib ticket

webapp/packages/plugin-connections/src/ConnectionForm/IConnectionFormState.ts
```diff
@@ -14,7 +14,7 @@ export const CONNECTION_FORM_STATE_SCHEMA = schema
     projectId: schema.string(),
     availableDrivers: schema.array(schema.string()),
     requiredNetworkHandlersIds: schema.array(schema.string()),
-    connectionId: schema.string().or(schema.undefined()),
+    connectionId: schema.string().or(schema.null()),
```
SychevAndrey: what was the cause of this changes?

SychevAndrey (commented):
Some changes look unrelated to lib updates and/or change app's logic

common-react/@dbeaver/react-tests/package.json
```diff
@@ -30,7 +30,7 @@
     "@testing-library/dom": "^10",
     "@testing-library/jest-dom": "^6",
     "@testing-library/react": "^16",
-    "@types/node": "^22",
+    "@types/node": "^24",
```
Wroud: Please keep the `^22` version for node; it's our current minimal supported node version

webapp/packages/core-utils/src/schema.ts
```diff
@@ -1,47 +1,56 @@
...
-  fallbackLng: 'en',
-  resources: {},
-});
-z.setErrorMap(zodI18nMap);
 
 export const schemaExtra = {
-  stringedBoolean() {
-    return z.union([z.enum(['false', '0']).transform(() => false), z.boolean(), z.string(), z.number()]).pipe(z.coerce.boolean());
-  },
+  locales: new Map<string, any>([]),
```
Wroud: `locales` should be private, you can move it to module scope:
```ts
const locales = new Map<string, any>([]);

export const schemaExtra = {
```

webapp/packages/plugin-sql-editor/src/SqlEditorSettingsService.ts
Wroud: I agree. Why did we make these changes?

webapp/packages/plugin-connections/src/ConnectionForm/IConnectionFormState.ts
Wroud: same question

Wroud (changes requested):

webapp/packages/core-administration/src/AdministrationScreen/AdministrationScreenService.ts
sergeyteleshev: Cause linter says `async` function should have `await`
If I remove `async`, TypeScript says: "You return Promise<boolean>", but the function has no `async`
So this is the compromise where ESLint and TypeScript are fine with the code

webapp/packages/plugin-connections/src/ConnectionForm/IConnectionFormState.ts
sergeyteleshev: Previously, this code led to this interface: 
`connectionId: schema.string().or(schema.undefined()) -> {connectionId: string | undefined }`

Now it leads to this interface:
`connectionId: schema.string().or(schema.undefined()) -> {connectionId?: string }`

We need this interface not to be optional according to the code

webapp/packages/plugin-sql-editor/src/SqlEditorSettingsService.ts
sergeyteleshev: Screenshot 2025-08-05 at 14 26 13

I changed the order of the functions. It should remain the same behavior

webapp/packages/plugin-connections/src/ConnectionForm/IConnectionFormState.ts
Wroud: Screenshot 2025-08-05 at 9 49 49 PM

`connectionId: schema.string().or(schema.null()) -> {connectionId: string | null }`

Wroud: Screenshot 2025-08-05 at 9 52 10 PM

here i've reverted to `.or(schema.undefined())` and removed `.required()`

Wroud (changes requested):

Wroud (dismissed):

webapp/packages/core-administration/src/AdministrationScreen/AdministrationScreenService.ts
SychevAndrey: As I can see that function returns `boolen | Promise<boolean>`, not only `Promise<boolean>`
It's better to fix return type

SychevAndrey (dismissed):

Wroud (approved):

PR #3635, opened 2025-08-05
dbeaver/pro#6298 refactor: useResource

Wroud:
closes dbeaver/pro#6298

This change is supposed to simplify useResource and remove unexpected update skipping


webapp/packages/core-blocks/src/ResourcesHooks/useResource.ts
```diff
@@ -128,352 +108,184 @@ export function useResource<
...
-              this.exception = new LoadingError(
-                () => {
-                  this.exception = null;
-                  this.load.bind(this, true);
-                },
-                'Application is unable to load resource',
-                { cause: exception },
-              );
-              propertiesRef.errorContext.catch(this.exception);
             }
```
sergeyteleshev, line 193: I suggest to add some early returns here to avoid nested ifs

webapp/packages/core-blocks/src/ResourcesHooks/useResource.ts
```diff
@@ -128,352 +108,184 @@ export function useResource<
...
+        // TODO: this is incorrect return type (key == null)
         return undefined;
       }
-      return propertiesRef.resource.get(propertiesRef.key);
+      return resource.get(key);
     }
 
-    return propertiesRef.resource.data;
+    return resource.data;
   }
```
sergeyteleshev, line 147: this is a pure function and can be moved outside of the hook so it can be covered with unit tests

Wroud: Yes, it can, but it uses generic types, so adding a bit more code will be required. I left it like this to keep it simple.

sergeyteleshev (approved):

devnaumov (approved):

PR #3639, opened 2025-08-06
dbeaver/pro#5968 refactor: update dark theme styles

SychevAndrey:
introduces new theme variable link-color which with a color which is used for coloring anchor elements in our app adjust this color in dark theme to have better contrast with a background, use this color in secondary button for the same reasons
slightly darken the surface color to have better contrast ratio without making the primary color too bright.
Screenshot 2025-08-06 at 15 39 02
Screenshot 2025-08-06 at 15 41 20


sergeyteleshev (dismissed):

Wroud (changes requested):
Screenshot 2025-08-07 at 8 20 41 PM

It's not enough contrast between the background and the surface color now

Wroud (dismissed):

Wroud (approved):

PR #3646, opened 2025-08-07
[CB] SQL semantic autocompletion doesn't take into account current schema

sergeyteleshev:
seems like we had filtering logic cause backend had sent not all valid options

since then now it seems to be sending relevant proposals so I decided to remove this logic at all keeping only sorting by score

closes https://github.com/dbeaver/pro/issues/6411

### DIFF
--- a/webapp/packages/plugin-sql-editor-new/src/SQLEditor/useSqlDialectAutocompletion.ts
+++ b/webapp/packages/plugin-sql-editor-new/src/SQLEditor/useSqlDialectAutocompletion.ts
@@ -29,16 +29,10 @@ export function useSqlDialectAutocompletion(data: ISQLEditorData): [Compartment,
   const optionsRef = useObjectRef({ data });
 
   const [config] = useState<CompletionConfig>(() => {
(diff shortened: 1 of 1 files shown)


Wroud (dismissed):

webapp/packages/plugin-sql-editor-new/src/SQLEditor/useSqlDialectAutocompletion.ts
```diff
@@ -29,26 +29,10 @@ export function useSqlDialectAutocompletion(data: ISQLEditorData): [Compartment,
   const optionsRef = useObjectRef({ data });
 
   const [config] = useState<CompletionConfig>(() => {
-    function getOptionsFromProposals(explicit: boolean, word: string, proposals: SQLProposal[]): SqlCompletion[] {
```
devnaumov, line 32: This will break current logic

https://github.com/user-attachments/assets/c0943038-fab6-4269-9ffe-993e6b6785de

https://github.com/user-attachments/assets/46df1e6d-5da7-41a6-818c-771dfa833a78

devnaumov (changes requested):

PR #3662, opened 2025-08-14
dbeaver/pro#6598 open panel on tab select

devnaumov:
closes [6598](https://github.com/dbeaver/pro/issues/6598)

### DIFF
--- a/webapp/packages/core-app/src/AppScreen/Main.tsx
+++ b/webapp/packages/core-app/src/AppScreen/Main.tsx
@@ -7,7 +7,7 @@
  */
 import { observer } from 'mobx-react-lite';
 
-import { getComputed, Loader, Pane, ResizerControls, s, Split, useS, useSplitUserState } from '@cloudbeaver/core-blocks';
+import { getComputed, Loader, Pane, ResizerControls, s, Split, useExecutor, useS, useSplitUserState } from '@cloudbeaver/core-blocks';
 import { useService } from '@cloudbeaver/core-di';
 import { LeftBarPanelService, SideBarPanel, SideBarPanelService } from '@cloudbeaver/core-ui';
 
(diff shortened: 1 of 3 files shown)


Wroud (approved):

webapp/packages/core-app/src/AppScreen/Main.tsx
```diff
@@ -25,6 +25,17 @@ export const Main = observer(function Main() {
   const sideBarDisabled = getComputed(() => sideBarPanelService.tabsContainer.getDisplayed().length === 0);
   const leftBarDisabled = getComputed(() => leftBarPanelService.tabsContainer.getDisplayed().length === 0);
 
+  useExecutor({
+    executor: sideBarPanelService.tabsContainer.onTabSelect,
+    handlers: [
+      function showPanel() {
+        if (splitRightState.mode === 'minimize' && !sideBarDisabled) {
```
sergeyteleshev, line 32: probably you there is no need to check that `!sideBarDisabled` cause it is already checked inside `select -> getDisplayedTabInfo` if the tab exists

sergeyteleshev: it gonna stop here:
https://github.com/dbeaver/cloudbeaver/blob/188e9ed309705c9a1020b1dcbd1bd2d656f84cbb/webapp/packages/core-ui/src/Tabs/TabsContainer/TabsContainer.ts#L52-L56

sergeyteleshev (approved):

webapp/packages/core-app/src/AppScreen/Main.tsx
devnaumov: We want to be explicit here as the logic inside TabsContainer might change

PR #3672, opened 2025-08-19
[CB] Default theme is light on server settings on clean server 

sergeyteleshev:
closes https://github.com/dbeaver/pro/issues/6424

### DIFF
--- a/webapp/packages/core-theming/src/ThemeSettingsService.ts
+++ b/webapp/packages/core-theming/src/ThemeSettingsService.ts
@@ -15,10 +15,10 @@ import {
 } from '@cloudbeaver/core-settings';
 import { schema } from '@cloudbeaver/core-utils';
 
-import { FALLBACK_THEME_ID } from './themes.js';
+export const DEFAULT_THEME_ID = 'system';
 
 const settingsSchema = schema.object({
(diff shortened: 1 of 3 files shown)


webapp/packages/core-theming/src/ThemeSettingsService.ts
```diff
@@ -15,10 +15,10 @@ import {
 } from '@cloudbeaver/core-settings';
 import { schema } from '@cloudbeaver/core-utils';
 
-import { FALLBACK_THEME_ID } from './themes.js';
+export const DEFAULT_THEME_ID = 'system';
 
 const settingsSchema = schema.object({
-  'core.theming.theme': schema.string().default(FALLBACK_THEME_ID),
+  'core.theming.theme': schema.string().default(DEFAULT_THEME_ID),
```
sergeyteleshev, line 21: The initial issue was here.

When we click `Restore defaults` in the settings it seems like it gets this value from `.default()` cluase. Previously, it was `light` from `FALLBACK_THEME_ID`, so I changed it to `device` from `DEFAULT_THEME_ID`

sergeyteleshev:
We decided to close the PR because the issue seems more general. The mechanism of nesting settings is broken, and it is not supposed to get settings from Zod validation default parts

It will be fixed with several similar issues by @Wroud

PR #3675, opened 2025-08-19
dbeaver/pro#6605 fix: reset filters on logout

SychevAndrey:
> This commit adds onLogout handler in DataViewerTabService which disposes opened tabs on logout.
> 
> Previously we didn't handle that at all and model was persisted in the memory after logout. On next login, if there was any filters, we would try to send them to backend and got an error, because backend doesn't support initial requests(without resultId) with filters.
> 
> The main logic of a handler repeats the already existed onDisconnect handler logic, which was extracted to handleTabsCleanup method. The difference between handlers is only in a set of tabs(onLogout we dispose all of them, onDisconnect - only tabs of a selected connection) and ability to interrupt the process.

That was initial implementation. Changed to onUnload handler in [ObjectViewerTabService.ts](https://github.com/dbeaver/cloudbeaver/pull/3675/files/b2927dfbb8c8271f8f600663fa18f8e54e5385fd#diff-f782dfe5426bf316c02a9e8765377fbdfde691fe2f76b9770f570403d74aad71)

https://github.com/user-attachments/assets/d8518a3d-60e8-4b1f-bc38-e47e15f3a192

### DIFF
--- a/webapp/packages/plugin-object-viewer/src/ObjectViewerTabService.ts
+++ b/webapp/packages/plugin-object-viewer/src/ObjectViewerTabService.ts
@@ -67,6 +67,7 @@ export class ObjectViewerTabService {
       onSelect: this.selectObjectTab.bind(this),
       onClose: this.closeObjectTab.bind(this),
       canClose: this.canCloseObjectTab.bind(this),
+      onUnload: this.unloadObjectTab.bind(this),
 
       extensions: [
         projectProvider(this.getProject.bind(this)),
@@ -102,6 +103,10 @@ export class ObjectViewerTabService {
         context.tab.handlerState.objectId = data.newNodeId;
       }
     });
+  }
+
+  private async unloadObjectTab(tab: ITab<IObjectViewerTabState>) {
+    await this.dbObjectPageService.closePages(tab);
   }
 
   isPageActive(tab: ITab<IObjectViewerTabState>, page: ObjectPage): boolean {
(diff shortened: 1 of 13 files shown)


sergeyteleshev (dismissed):
Nice fix!

Wroud (changes requested):
In `NavigationTabsService`, we have `unloadTabs`
tab handler (`TabHandler`) has `onUnload` / `onRestore` callbacks
in `ObjectViewerTabService`, we are adding `TabHandler`
In `DataViewerTabService`, we are registering the page to the Object Viewer with `onUnload` and `onRestore` handlers. `onUnload` should dispose the table model

In general, logic should be like this:
1. Closing or unloading tabs (on logout, for example)
2. Disposing of any resources used by `TabHandlers`
3. Disposing of any Pages resources used by Object Viewer Pages
4. Disposing of any Data Models used by Data Viewer pages

Logic that you've explicitly added to the Data Viewer is extra, because it's already handled by `NavigationTabsService` (tabs unloaded on logout). The problem here is that the model doesn't dispose of for some reason

webapp/packages/plugin-object-viewer/src/ObjectViewerTabService.ts
```diff
@@ -104,6 +105,10 @@ export class ObjectViewerTabService {
     });
   }
 
+  private async unloadObjectTab(tab: ITab<IObjectViewerTabState>) {
+    await this.dbObjectPageService.closePages(tab);
```
Wroud, line 110: please add todo that we want to call `Unload` on pages not `Close`

Wroud (dismissed):

webapp/packages/plugin-object-viewer/src/ObjectViewerTabService.ts
Wroud, line 110:
```suggestion
    // TODO: we need to call unloadPages, but it's not implemented in the DBObjectPageService
    await this.dbObjectPageService.closePages(tab);
```

Wroud (approved):

PR #3681, opened 2025-08-20
[CB-4362] Add trim to all inputs in forms

sergeyteleshev:
closes https://github.com/dbeaver/pro/issues/4696

### DIFF
--- a/webapp/packages/core-ui/src/Form/FormPart.ts
+++ b/webapp/packages/core-ui/src/Form/FormPart.ts
@@ -5,7 +5,7 @@
  * Licensed under the Apache License, Version 2.0.
  * you may not use this file except in compliance with the License.
  */
-import { action, computed, makeObservable, observable, toJS } from 'mobx';
+import { action, computed, makeObservable, observable, runInAction, toJS } from 'mobx';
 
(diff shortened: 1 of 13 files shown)


webapp/packages/core-ui/src/Form/FormPart.ts
```diff
@@ -174,7 +175,11 @@ export abstract class FormPart<TPartState, TFormState = any> implements IFormPar
     this.state = state;
   }
 
-  protected format(data: IFormState<TFormState>, contexts: IExecutionContextProvider<IFormState<TFormState>>): void | Promise<void> {}
+  protected format(data: IFormState<TFormState>, contexts: IExecutionContextProvider<IFormState<TFormState>>): void | Promise<void> {
+    runInAction(() => {
+      propagateTrim(this.state);
+    });
```
Wroud, line 181: I don't like this idea. We have a schema for the state where we can manually specify which fields should be truncated:
https://zod.dev/api?id=strings

Truncating is not always required and may lead to unwanted behaviour, so it can't be applied to the whole state, especially when we have complex states

Wroud (changes requested):

sergeyteleshev:
This solution seems too "magic" and can cause unexpected behavior
We are not ready for the experiments, so we decided to go with `Zod.trim()` transformations

PR #3688, opened 2025-08-21
dbeaver/pro#6692 do not load settings until authorized

devnaumov:
closes [6693](https://github.com/dbeaver/pro/issues/6693)

### DIFF
--- a/webapp/packages/core-authentication/src/UserInfoResource.ts
+++ b/webapp/packages/core-authentication/src/UserInfoResource.ts
@@ -30,6 +30,7 @@ export interface ILoginOptions {
 export type IFederatedLoginOptions = Omit<ILoginOptions, 'credentials'>;
 
 export const ANONYMOUS_USER_ID = 'anonymous';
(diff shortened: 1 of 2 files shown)


Wroud (approved):

webapp/packages/core-authentication/src/UserInfoResource.ts
```diff
@@ -30,6 +30,7 @@ export interface ILoginOptions {
 export type IFederatedLoginOptions = Omit<ILoginOptions, 'credentials'>;
 
 export const ANONYMOUS_USER_ID = 'anonymous';
+export const UNAUTHORIZED_ID = 'unauthorized';
```
sergeyteleshev, line 33: I don't understand this change. Could you explain why?

Now this new `UNAUTHORIZED_ID` is not reused anywhere. Moreover, `ANONYMOUS_USER_ID` can now be moved inside `webapp/packages/plugin-query-manager/src/Table/Cell.tsx` cause it is not reused

PR #3704, opened 2025-08-27
dbeaver/pro#6744 refactor: migration to `@wroud/di`

Wroud:
closes dbeaver/pro#6744


webapp/packages/core-blocks/src/useS.ts
```diff
@@ -46,7 +46,10 @@ export function useS<T extends ComponentStyle[]>(...componentStyles: [...T]): Ex
   const stylesRef = useRef<ComponentStyle[]>([]);
   const [patch, forceUpdate] = useState(0);
   const loadedStyles = useRef<BaseStyles[]>([]);
-  const themeService = useService(ThemeService, true);
+  let themeService: ThemeService | null = null;
+  try {
+    themeService = useService(optional(ThemeService)).resolve();
+  } catch {}
```
SychevAndrey: Can you explain a bit how it works and what's the difference with a previous approach?

webapp/packages/core-di/src/App.ts
```diff
@@ -76,107 +79,63 @@ export class App {
...
 
-        if ('load' in serviceInstance) {
-          await serviceInstance.load();
+    if (preload) {
+      for (const service of this.serviceProvider.getServices(IPreloadService)) {
+        await service.register?.();
+      }
+    } else {
+      for (const service of this.serviceProvider.getServices(Bootstrap)) {
+        if ('register' in service) {
```
sergeyteleshev: better to check `service instanceof Bootstrap` so inside the if clause you will have the whole interface of the service
plus if you change the API inside Bootstrap like `register -> activate` and forget to change this line to `'activate' in service` it probably would skip and break the logic

webapp/packages/core-di/src/App.ts
```diff
@@ -76,107 +79,63 @@ export class App {
...
-      i++;
-
-      if (i >= concurrency) {
-        const services = (await Promise.all(queue)).flat();
-
-        for (const service of services) {
-          yield service;
+    } else {
+      for (const service of this.serviceProvider.getServices(Bootstrap)) {
+        if ('load' in service) {
```
sergeyteleshev: same here

sergeyteleshev (dismissed):

webapp/packages/core-di/src/App.ts
Wroud: actually i can just remove it

SychevAndrey (dismissed):

webapp/packages/core-blocks/src/useS.ts
devnaumov: Agree, it looks strange, especially try catch block

Wroud: changed a bit, no difference with a previous approach, this service here is optional because this hook may be used while application not initialized

devnaumov (dismissed):

webapp/packages/core-blocks/src/useS.ts
Wroud: rewrited to default approach without try/catch

SychevAndrey (dismissed):

sergeyteleshev (approved):

PR #3710, opened 2025-09-01
dbeaver/pro#6639 fix: don't render table error component when no error

SychevAndrey:
We used to render TableError component all the time no matter if we have an error or not, just hiding this component visually. This approach breaks accessibility because Table error component was still reachable from a keyboard or a screen reader. This fix completely removes the node from DOM if the error is not visible.


webapp/packages/plugin-data-viewer/src/TableViewer/TableError.tsx
```diff
@@ -87,31 +87,33 @@ export const TableError = observer<Props>(function TableError({ model, loading,
...
-      tabIndex={0}
-      className={s(style, { error: true, animated, collapsed: !errorInfo.display, errorHidden }, className)}
-    >
-      <div className={s(style, { errorBody: true })}>
-        <IconOrImage className={s(style, { iconOrImage: true })} icon={icon} title={error.message} onClick={() => errorInfo.show()} />
-        <div className={s(style, { errorMessage: true })}>{error.message}</div>
-      </div>
-      <div className={s(style, { controls: true })}>
-        <Placeholder container={dataViewerService.errorActionsContainer} model={model} />
+    error.error && (
```
Wroud: In such cases, it's better to make an early return:
```
if (error.error) {
  return null;
}
```

But in this component, this is on purpose (keeping it in the dom) for animations,
you probably need to set `tabIndex={error.error ? 0 : -1}` instead

Wroud (changes requested):

sergeyteleshev (dismissed):

Wroud (changes requested):
something wrong with indentation

Wroud (approved):

sergeyteleshev (approved):

devnaumov (approved):

PR #3727, opened 2025-09-08
dbeaver/pro#5614 api for expert settings props

yagudin10:


webapp/packages/core-connections/src/DBDriverExpertSettingsResource.ts
```diff
@@ -0,0 +1,54 @@
...
+@injectable(() => [GraphQLService, DBDriverResource])
+export class DBDriverExpertSettingsResource extends CachedMapResource<string, DriverPropertyInfo[]> {
+  constructor(
+    private readonly graphQLService: GraphQLService,
+    private readonly dbDriverResource: DBDriverResource,
+  ) {
+    super();
+
+    this.sync(this.dbDriverResource);
+    this.dbDriverResource.onItemDelete.addHandler(this.delete.bind(this));
```
sergeyteleshev, line 27: should we do something on `this.dbDriverResource.onItemUpdate`?

webapp/packages/core-connections/src/DBDriverResource.ts
```diff
@@ -27,22 +28,25 @@ export const NEW_DRIVER_SYMBOL = Symbol('new-driver');
...
+
+@injectable(() => [SessionDataResource, GraphQLService, WorkspaceConfigEventHandler, AppAuthService])
 export class DBDriverResource extends CachedMapResource<string, DBDriver, DBDriverResourceIncludes> {
   get enabledDrivers() {
     return this.values.filter(driver => driver.enabled).sort(this.compare);
   }
 
   constructor(
-    private readonly serverConfigResource: ServerConfigResource,
+    private readonly sessionDataResource: SessionDataResource,
```
Wroud: Is it true that driver objects may change depending on the session?

webapp/packages/core-connections/src/DBDriverResource.ts
```diff
@@ -27,22 +28,25 @@ export const NEW_DRIVER_SYMBOL = Symbol('new-driver');
...
     private readonly graphQLService: GraphQLService,
     private readonly workspaceConfigEventHandler: WorkspaceConfigEventHandler,
     appAuthService: AppAuthService,
   ) {
     super();
     appAuthService.requireAuthentication(this);
+
     this.sync(
-      this.serverConfigResource,
```
sergeyteleshev, line 45: This sync should remain untouched.
The server config has a list of disabled drivers. So, if we disable something in the admin settings, we should update the driver list so it has a fresh state of enabled/disabled drivers.

So we can add extra sync with `sessionDataResource` or sync `sessionDataResource` inside `serverConfigResource`. We need to think this through

webapp/packages/core-connections/src/DBDriverResource.ts
Wroud: If the only `DBDriverExpertSettingsResource` will change, it's better to target`sync` it with `SessionDataResource` together with sync with `DBDriverResource`

webapp/packages/core-connections/src/locales/zh.ts
```diff
@@ -44,12 +44,6 @@ export default [
   ['connections_connection_create_fail', '无法创建连接'],
   ['connections_connection_save_fail', '无法保存连接'],
   ['connections_connection_expert_settings', '专业设置'],
-  ['connections_connection_keep_alive', '保持连接 (秒)'],
-  ['connections_connection_autocommit', '自动提交'],
-  ['connections_connection_default_catalog', 'Default catalog'],
-  ['connections_connection_default_schema', 'Default schema'],
-  ['connections_connection_keep_alive_tooltip', '无自动断开连接'],
-  ['connections_connection_read_only', 'Read-only connection'],
```
Wroud, line 52: We need to keep localizations if they were not changed, but just moved to the server side

webapp/packages/core-sdk/src/getObjectPropertyValueType.ts
```diff
@@ -16,7 +16,7 @@ export function getObjectPropertyValueType(property: ObjectPropertyInfo): Object
     return 'password';
   }
 
-  if (dataType === 'integer') {
+  if (dataType === 'integer' || dataType === 'int') {
```
sergeyteleshev, line 19: `bigint, tinyint, float, double, etc.`
what types backend can send to us? for now only these 2?

webapp/packages/core-connections/src/module.ts
```diff
@@ -61,6 +62,7 @@ ModuleRegistry.add({
       .addSingleton(NetworkHandlerResource)
       .addSingleton(NavNodeExtensionsService)
       .addSingleton(DBDriverResource)
+      .addSingleton(DBDriverExpertSettingsResource)
```
Wroud, line 66: please also add `.addSingleton(Dependency, proxy(DBDriverExpertSettingsResource))`
This will keep the resource initialized even if it's not used

server/bundles/io.cloudbeaver.server/schema/service.core.graphqls
```diff
@@ -351,6 +351,9 @@ type DriverInfo {
     """
     providerProperties: [ObjectPropertyInfo!]!
 
+    "Expert driver settings properties. Returns properties (like keep-alive interval) that are not often used and can be hidden in UI"
+    expertSettingsProperties: [ObjectPropertyInfo!]! @since(version: "25.2.1")
```
Wroud, line 355: Should we mark old fields as deprecated?

Wroud (changes requested):

sergeyteleshev (changes requested):

webapp/packages/core-connections/src/DBDriverResource.ts
devnaumov: sessionDataResource is synced with serverConfigResource

webapp/packages/core-connections/src/DBDriverExpertSettingsResource.ts
devnaumov: seems like no

webapp/packages/core-sdk/src/getObjectPropertyValueType.ts
devnaumov: the backend itself does not know, for now we have two cases, one when properties are mapped from java objects and second when properties are mapped from xml

server/bundles/io.cloudbeaver.model/OSGI-INF/l10n/bundle_zh.properties
```diff
@@ -0,0 +1,4 @@
+meta.io.cloudbeaver.model.WebExpertSettingsProperties.readOnly.name = Только для чтения
```
Wroud: There should be an English version

webapp/packages/core-connections/src/DBDriverExpertSettingsResource.ts
Wroud: no, we don't need

Wroud (changes requested):

Wroud (dismissed):

webapp/packages/plugin-connections/src/ConnectionForm/Options/AdvancedPropertiesForm.tsx
```diff
@@ -27,70 +19,17 @@ interface Props {
...
-              readOnly={readonly || disabled}
-              title={translate('connections_connection_default_schema')}
-              state={config}
-              defaultState={DEFAULT_CONFIG}
-              fill
-            >
-              {translate('connections_connection_default_schema')}
-            </InputField>
-          </Flex>
+          <ObjectPropertyInfoForm state={config.expertSettingsValues} properties={properties.data} disabled={disabled} readOnly={readonly} />
```
sergeyteleshev, line 32: please also check if the form responses right on window/panel resize. may be props like `small/keepSize` would be fine here

sergeyteleshev (dismissed):

alexander-skoblikov (approved):

PR #3738, opened 2025-09-11
[CB] get back switch title bold styles

sergeyteleshev:
closes https://github.com/dbeaver/pro/issues/6878

Screenshot 2025-09-11 at 16 07 04

### DIFF
--- a/webapp/packages/core-blocks/src/FormControls/Checkboxes/Switch.module.css
+++ b/webapp/packages/core-blocks/src/FormControls/Checkboxes/Switch.module.css
@@ -5,26 +5,27 @@
  * Licensed under the Apache License, Version 2.0.
  * you may not use this file except in compliance with the License.
  */
+@layer components {
+  .field {
+    --dbv-kit-control-outline-offset: -8px;
+    position: relative;
 
-.field {
-  --dbv-kit-control-outline-offset: -8px;
-  position: relative;
+    &::before {
+      content: '';
+      position: absolute;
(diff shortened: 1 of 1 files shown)


webapp/packages/core-blocks/src/FormControls/Checkboxes/Switch.module.css
```diff
@@ -6,71 +6,68 @@
  * you may not use this file except in compliance with the License.
  */
 
-@layer components {
```
sergeyteleshev, line 9: i just removed this. something happend with github formatting

devnaumov: I think we should keep @layer components as we use it in all our components, so it might break something

devnaumov (dismissed):

webapp/packages/core-blocks/src/FormControls/Checkboxes/Switch.module.css
```diff
@@ -29,48 +27,48 @@
...
-      font-weight: 500;
-    }
+  & .fieldLabel {
+    cursor: pointer;
+    user-select: none;
+    display: block;
+    padding-left: 18px;
+    min-width: 50px;
+    white-space: pre-wrap;
+    font-weight: 500;
```
Wroud: probably custom styles from switchBody should be keept in the layer, but composes can be outside of it

Wroud (changes requested):

Wroud (approved):

PR #3742, opened 2025-09-12
dbeaver/pro#6327 refactor: sql editor & vqb

Wroud:
closes dbeaver/pro#6327
closes dbeaver/pro#5174
closes dbeaver/pro#4922


webapp/common-typescript/@dbeaver/js-helpers/src/debouncePromise.ts
```diff
@@ -0,0 +1,9 @@
+/*
+ * CloudBeaver - Cloud Database Manager
+ * Copyright (C) 2020-2025 DBeaver Corp and others
+ *
+ * Licensed under the Apache License, Version 2.0.
+ * you may not use this file except in compliance with the License.
+ */
+
+export { default as debouncePromise } from 'p-debounce';
```
sergeyteleshev, line 9: seems like we have async debounce implementation:
https://github.com/dbeaver/cloudbeaver/blob/efaa548ca6ab5d3bb914430f3e52a62df8450fe1/webapp/packages/core-utils/src/debounce.ts#L22-L45

webapp/packages/plugin-sql-editor-navigation-tab/src/SqlEditorTabService.ts
```diff
@@ -531,12 +535,14 @@ export class SqlEditorTabService extends Bootstrap {
...
 
   private async handleTabUnload(editorTab: ITab<ISqlEditorTabState>) {
+    await this.sqlEditorModelService.unload(editorTab.handlerState.editorId);
     await this.sqlDataSourceService.unload(editorTab.handlerState.editorId);
 
     this.sqlResultTabsService.removeResultTabs(editorTab.handlerState);
   }
 
   private async handleTabCloseSilent(editorTab: ITab<ISqlEditorTabState>) {
+    await this.sqlEditorModelService.destroySilent(editorTab.handlerState.editorId);
```
sergeyteleshev, line 545: currently destroy and destroySilent is the same implementation. do we need it yet?

webapp/packages/core-blocks/src/useSync.ts
```diff
@@ -0,0 +1,56 @@
...
+
+import { mutex } from '@dbeaver/js-helpers';
+import { useEffect, useState } from 'react';
+import { useObjectRef } from './useObjectRef.js';
+
+interface ISyncHook {
+  markOutdated(): void;
+}
+
+export function useSync(callback: () => void | Promise<void>): ISyncHook {
```
devnaumov: The name of this hook is too generic. Could we give it a more explicit name so it’s clearer why we need it?

webapp/packages/core-blocks/src/useSync.ts
```diff
@@ -0,0 +1,56 @@
...
+  const [syncMutex] = useState(() => new mutex.Mutex());
+  const [outdated, setOutdated] = useState(false);
+  const [delayedOutdated, setDelayedOutdated] = useState(false);
```
sergeyteleshev, line 20: may be we can use `ref` here instead of the `state`. because our effect depends only on `outdated` and `mutex.isLocked()`
by this change we would have for 1 less rerender cause `setState` causes it

all in all `outdated` and `delayedOutdated` state handling seems a bit complicated. may be mutex API allows to make it more simple?

Wroud: No matter how many useStates you have in a component, there will be no extra re-renders until you call change state in a synchronous manner. React will batch such updates and re-render the component only once.
Moreover, in this implementation, it's essential to have two separate `useState` because `syncMutex` is an untracked dependency, it will not change, but we need to trigger re-render to activate the effect when we have `outdated = true` and `delayedOutdated = true`.

webapp/packages/plugin-sql-editor-navigation-tab/src/SqlEditorTabService.ts
Wroud: I will keep it for consistency

webapp/common-typescript/@dbeaver/js-helpers/src/debouncePromise.ts
Wroud: We can replace our implementation later with `p-debounce`, but it can't be replaced without testing/changes because they have different behaviour

webapp/packages/core-blocks/src/useSync.ts
Wroud: I agree; it's generic (and it's really generic), but I can't imagine another name. This hook aims to provide a simple way to implement a React-based reaction to side effects. Generally, we have imperative code that will invoke the function right away. This leads to problems when an event might be called multiple times, so we will have numerous functions running in concurrent mode.

That leads to problems like this one you fixed earlier: https://github.com/dbeaver/cloudbeaver/pull/3734

`useSync` "connects" events with React lifecycle, so the interface can react to changes in a React manner by utilizing React's built-in batching functionality.

sergeyteleshev (dismissed):

devnaumov (dismissed):

sergeyteleshev (dismissed):

devnaumov (dismissed):

devnaumov (dismissed):

devnaumov (dismissed):

sergeyteleshev (dismissed):

sergeyteleshev (approved):

PR #3750, opened 2025-09-17
dbeaver/pro#5471 [CB] Add shortcuts to tooltips data editor toolbar

SychevAndrey:
Added:
1. Delete shortcut info similar to other footer actions + fix wording
2. **Mod + /** and **Mod + F** to shortcuts list
image
image

Save as script and save buttons now use the same shortcut since they never appear together and look similar so they function as a single button. (https://github.com/dbeaver/pro/issues/6891) Also fixes order.
image


webapp/packages/plugin-data-viewer/src/TableViewer/TableFooter/TableFooterMenu/TableFooterMenuService.ts
```diff
@@ -203,7 +203,7 @@ export class TableFooterMenuService {
       case ACTION_DUPLICATE:
         return { ...action.info, label: '', icon: '/icons/data_add_copy_sm.svg', tooltip: t('data_viewer_action_edit_add_copy') + ' (' + getBindingLabel(KEY_BINDING_DUPLICATE) + ')' };
       case ACTION_DELETE:
-        return { ...action.info, label: '', icon: '/icons/data_delete_sm.svg', tooltip: t('data_viewer_action_edit_delete') };
+        return { ...action.info, label: '', icon: '/icons/data_delete_sm.svg', tooltip: t('data_viewer_action_edit_delete') + ' (Delete)' };
```
Wroud: Probably we should use the bindings api instead?

webapp/packages/plugin-help/src/locales/en.ts
```diff
@@ -17,10 +17,12 @@ export default [
   ['sql_editor_shortcut_execute_script', 'Execute script'],
   ['sql_editor_shortcut_show_execution_plan', 'Show Execution plan'],
   ['sql_editor_shortcut_format', 'Format script'],
-  ['sql_editor_shortcut_save_as_script', 'Save as script'],
+  ['sql_editor_shortcut_save', 'Save'],
```
Wroud: we have `ui_processing_save`

Wroud (changes requested):

webapp/packages/plugin-data-viewer/src/TableViewer/TableFooter/TableFooterMenu/TableFooterMenuService.ts
SychevAndrey: What do you mean? Create a binding only to extract the shortcut here? Since it's not OS-specific I think it's an overkill.

webapp/packages/plugin-help/src/locales/en.ts
SychevAndrey: `ui_processing_save` is related to action elements like buttons and `sql_editor_shortcut_save` is a part of content that might be changed in any time and don't depend on actions.
I would rather remove `plugin_sql_editor_navigation_tab_resource_save_script_title` in favour of `ui_processing_save` if you think that we can't have two keys with same string value.

Wroud: No, we can. But usually, if we have `ui_*` with the same meaning, then it's better to use it, but if there is the same "No data", for example, in two different plugins, it's okay to duplicate it.
If we will decide to rename it later, we always can add translation key in the plugin.

webapp/packages/plugin-data-viewer/src/TableViewer/TableFooter/TableFooterMenu/TableFooterMenuService.ts
```diff
@@ -185,25 +204,34 @@ export class TableFooterMenuService {
...
+          icon: '/icons/data_add_sm.svg',
+          tooltip: t('data_viewer_action_edit_add') + ' (' + getBindingLabel(KEY_BINDING_ADD) + ')',
+        };
       case ACTION_DUPLICATE:
-        return { ...action.info, label: '', icon: '/icons/data_add_copy_sm.svg', tooltip: t('data_viewer_action_edit_add_copy') + ' (' + getBindingLabel(KEY_BINDING_DUPLICATE) + ')' };
+        return {
+          ...action.info,
+          label: '',
+          icon: '/icons/data_add_copy_sm.svg',
+          tooltip: t('data_viewer_action_edit_add_copy') + ' (' + getBindingLabel(KEY_BINDING_DUPLICATE) + ')',
```
Wroud: Can we specify duplicate and add bindings in the same way as 'delete'?

Wroud (dismissed):

webapp/packages/plugin-data-viewer/src/TableViewer/TableFooter/TableFooterMenu/TableFooterMenuService.ts
SychevAndrey: Yes, we can, do you want me to make it in this task or create another one?

SychevAndrey: done

Wroud (dismissed):

sergeyteleshev (dismissed):

devnaumov (dismissed):

Wroud (approved):

PR #3759, opened 2025-09-18
dbeaver/pro#6520 fix: header cell layout

SychevAndrey:
Screenshot 2025-09-18 at 11 52 18
Screenshot 2025-09-18 at 11 52 07
Screenshot 2025-09-18 at 11 51 46


Wroud (changes requested):
We need to fix the grid view when there is a description in some columns and not in others, so the column name and icon will always be at the same position (top)

Probably we can just replace `tableDataContext.hasDescription && columnDescription` with `tableDataContext.hasDescription` (`packages/plugin-data-spreadsheet-new/src/DataGrid/TableColumnHeader/TableColumnHeader.tsx)` and maybe adjust styles for description div so it always will have height

Wroud (dismissed):

SychevAndrey:
Screenshot 2025-09-22 at 08 22 24
Screenshot 2025-09-22 at 08 22 19

Wroud:
Screenshot 2025-09-22 at 4 50 35 PM

Two rows should be displayed only if there is a description in at least one column

Wroud (approved):

sergeyteleshev (approved):

PR #3760, opened 2025-09-18
[CB] Migrate context menu to aria kit #6756

sergeyteleshev:
closes https://github.com/dbeaver/pro/issues/6756


webapp/common-react/@dbeaver/ui-kit/src/Menu/MenuField.tsx
```diff
@@ -0,0 +1,101 @@
...
+  store: ReturnType<typeof useMenuStore>;
+
+  'aria-labelledby'?: string;
+
+  'aria-label'?: string;
+
+  id?: string;
+}
+
+export function MenuField<T>({
```
Wroud: i don't think that we need such component:
1. `items` used only to call `items.map(itemRender)`, this can be done on the provider level
2. Naming is strange. What is Menu Field?
3. I see `reuired` but i don't see `onChange`, `value` or any other input related props, this doesn't make sense

Moreover, what is the purpose and difference from a combobox?

Wroud (changes requested):

webapp/common-react/@dbeaver/ui-kit/src/Menu/Menu.css
```diff
@@ -0,0 +1,126 @@
...
+      opacity: var(--dbv-kit-control-disabled-opacity);
+      color: var(--dbv-kit-menu-disabled-foreground);
+      background-color: var(--dbv-kit-menu-disabled-background);
+    }
+
+    &:focus-visible {
+      outline-style: solid;
+      outline-offset: var(--dbv-kit-control-outline-offset);
+    }
+  }
```
Wroud: I think that we don't need at all these styles and special "Menu Button" component; we should use composition so that we can use any existing component (for example, Button) as a menu button

webapp/common-react/@dbeaver/ui-kit/src/Menu/Menu.css
```diff
@@ -0,0 +1,126 @@
...
+
+  .dbv-kit-menu__popover {
+    z-index: 1000;
+    min-width: max-content;
+    max-height: var(--dbv-kit-menu-popover-max-height);
```
Wroud, line 7: Probably we need to add `max-width`

Wroud (changes requested):

webapp/packages/core-blocks/src/PropertiesTable/PropertyItem.tsx
```diff
@@ -71,6 +70,7 @@ export const PropertyItem = observer<Props>(function PropertyItem({ property, va
         <Combobox
           value={value}
           title={value}
+          defaultValue={property.defaultValue}
```
Wroud, line 70: We don't need this because it will be displayed as a "manually set" value.
The idea of removing it is to display unset values as placeholders to indicate its default value and that it's not changed by the user.

Wroud (changes requested):

Wroud (dismissed):

webapp/packages/storybook/src/stories/components/Menu/menu.stories.tsx
```diff
@@ -0,0 +1,263 @@
...
+ *
+ * Licensed under the Apache License, Version 2.0.
+ * you may not use this file except in compliance with the License.
+ */
+
+import { Button, Menu, useMenuStore } from '@dbeaver/ui-kit';
+import type { Meta, StoryObj } from '@storybook/react-vite';
+import { Fragment, useMemo, useState } from 'react';
+
+const meta = {
```
SychevAndrey, line 13: let's add autodocs page from storybook

```
const meta = {
  component: Menu.Button,
  //👇 Enables auto-generated documentation for this component and includes all stories in this file
  tags: ['autodocs'],
} satisfies Meta<typeof Menu.Button>;
export default meta;
```
https://storybook.js.org/docs/writing-docs/autodocs

SychevAndrey (dismissed):

webapp/common-react/@dbeaver/ui-kit/src/Input/Input.css
```diff
@@ -32,12 +32,6 @@
     &::placeholder {
       color: var(--dbv-kit-input-placeholder-foreground);
     }
-
-    &:focus-visible,
-    &[data-focus-visible] {
-      outline: var(--dbv-kit-control-outline-width) solid var(--dbv-kit-control-outline-color);
-      outline-offset: var(--dbv-kit-control-outline-offset);
-    }
   }
```
Wroud, line 41: This change will likely remove the outline effect. Is it expected? It will affect all forms

sergeyteleshev: fixed that via border-outline-offset prop

SychevAndrey: why do we touch UI-kit elements that are not event parts of that task and change their core functionality?
We have consistent outline behaviour among all the components like inputs, buttons and selects, why we decide to change it here?

image

SychevAndrey:
We should not use 2 (or more) different version of inputs in our app.
Either we use Input from dbeaver/ui-kit or we have to have a custom Input component and use it everywhere including InputField

Wroud (dismissed):

mr-anton-t (approved):

PR #3801, opened 2025-10-06
dbeaver/pro#7001 feat: add combobox cancel feature

SychevAndrey:
image


webapp/packages/core-blocks/src/FormControls/Combobox.tsx
```diff
@@ -215,12 +234,20 @@ export const Combobox: ComboboxType = observer(function Combobox({
...
             <Spinner size="small" className="tw:absolute tw:right-2 tw:top-[50%] tw:-translate-y-1/2" />
           ) : (
-            displayPopover && (
-              <ComboboxDisclosure
-                disabled={disabled || loading || readOnly}
-                className="tw:absolute tw:right-2 tw:top-[50%] tw:-translate-y-1/2 tw:[&>svg]:fill-none! tw:text-sm! tw:cursor-pointer"
-              />
-            )
+            <>
+              {displayValue && allowClear && !disabled && !readOnly && (
```
sergeyteleshev, line 225: and `focused` may be?

SychevAndrey: why? we don't need it

webapp/packages/core-blocks/src/FormControls/Combobox.tsx
```diff
@@ -169,6 +171,19 @@ export const Combobox: ComboboxType = observer(function Combobox({
...
+    if (name !== undefined && state) {
+      state[name] = undefined;
+    }
+    if (onSelect) {
+      onSelect(undefined as any, name as undefined, selectedKey);
+    }
+    if (context) {
+      context.change('', name);
+    }
+  }
```
Wroud: Why do we need manual handling? Don't the already-provided callbacks cover it for the checkbox itself?

webapp/packages/core-blocks/src/FormControls/Combobox.tsx
```diff
@@ -215,12 +234,20 @@ export const Combobox: ComboboxType = observer(function Combobox({
...
-                disabled={disabled || loading || readOnly}
-                className="tw:absolute tw:right-2 tw:top-[50%] tw:-translate-y-1/2 tw:[&>svg]:fill-none! tw:text-sm! tw:cursor-pointer"
-              />
-            )
+            <>
+              {displayValue && allowClear && !disabled && !readOnly && (
+                <ComboboxCancel
+                  className="tw:absolute tw:right-8 tw:top-[50%] tw:-translate-y-1/2 tw:[&>svg]:fill-none! tw:text-sm! tw:cursor-pointer tw:opacity-50 hover:tw:opacity-100"
+                  onClick={handleClear}
+                />
```
Wroud, line 229: it seems like it should appear instead of Disclosure

Wroud (changes requested):

webapp/packages/core-blocks/src/FormControls/Combobox.tsx
SychevAndrey: Why? it's different elements. If we think that we don't need disclosure in case of selected value, why do we need it initially?

webapp/packages/core-blocks/src/FormControls/Combobox.tsx
SychevAndrey: WDYM? It's basically what we need on select, inside we use onSelect callback and setInput value.

Wroud (approved):

sergeyteleshev (approved):

devnaumov (approved):

PR #3819, opened 2025-10-11
[CB-5925] Charts - add export option (image) 

sergeyteleshev:
closes https://github.com/dbeaver/pro/issues/5078


webapp/packages/core-utils/package.json
```diff
@@ -24,6 +24,7 @@
   },
   "dependencies": {
     "@timohausmann/quadtree-ts": "^2",
+    "dom-to-image": "^2",
```
sergeyteleshev: A tried 3 solutions:

1. use our own one we implemented in diagrams. this did not workout - it exports graphs without a legend. moreover it requires a really hard code support and we must delete the code and use ready to go solution which already considers all cases during converting of html to image
2. `html2canvas` - in general it works, but I had problems with export. some parts of the graph (legends) exports without css styles. as result it looks really bad
3. `dom-to-image` - this lib works like a clock. exports everything easy and with all of the styles. moreover it has no dependencies so it reduces a risk of CVE. if you have an argument that it is an old one and not supported for a long time. well, to defend this solution I can say that it does what it does - exports dom to image. there is no other support required cause all functionality implemented and there are no dependencies I mentioned above. so as for me it looks fine go further. other solutions either don't work or extremely complicated to support

webapp/packages/core-blocks/src/ExportImageDialog/ExportImageDialog.module.css
```diff
@@ -0,0 +1,5 @@
+.footer {
+  align-items: center;
+  justify-content: flex-end;
+  gap: 24px;
+}
```
Wroud: can be replaced with tailwind styles

webapp/packages/core-blocks/src/ExportImageDialog/ExportImageDialogLazy.tsx
```diff
@@ -0,0 +1,11 @@
...
+ * CloudBeaver - Cloud Database Manager
+ * Copyright (C) 2020-2025 DBeaver Corp and others
+ *
+ * Licensed under the Apache License, Version 2.0.
+ * you may not use this file except in compliance with the License.
+ */
+
+import { importLazyComponent } from '../importLazyComponent.js';
+
+export const ExportImageDialogLazy = importLazyComponent(() => import('./ExportImageDialog.js').then(m => m.ExportImageDialog));
```
Wroud: for such files use `.ts` extension
also it's better to re-export with original name (`ExportImageDialog`)

webapp/packages/core-cli/configs/excludedFromVendor.js
```diff
@@ -12,7 +12,7 @@ module.exports = [
   'joint',
   'backbone',
   'lodash',
-  'canvg',
+  'domToImage',
```
Wroud: In this list should be import name of the library should be; this is not it

webapp/packages/core-utils/src/domToImage.ts
```diff
@@ -0,0 +1,32 @@
+/*
+ * CloudBeaver - Cloud Database Manager
+ * Copyright (C) 2020-2025 DBeaver Corp and others
+ *
+ * Licensed under the Apache License, Version 2.0.
+ * you may not use this file except in compliance with the License.
+ */
+// @ts-ignore
+import * as dti from 'dom-to-image';
```
Wroud: We can't use the dead library :(
How about:
1. https://www.npmjs.com/package/html-to-image
2. https://www.npmjs.com/package/modern-screenshot
3. https://www.npmjs.com/package/html2canvas-pro

and more can be found here: https://www.npmjs.com/search?q=keywords:screenshot

Wroud (changes requested):

webapp/packages/core-utils/src/domToImage.ts
sergeyteleshev: used html-to-image

webapp/packages/core-blocks/src/ExportImageDialog/ExportImageDialog.tsx
```diff
@@ -0,0 +1,101 @@
...
+  const errorDetails = useErrorDetails(state.error);
+
+  const onExportHandler = () => {
+    try {
+      props.payload.onExport(state);
+      props.resolveDialog();
+    } catch (error: any) {
+      state.error = error;
+    }
+  };
```
Wroud: please use named function

Wroud: it seems like `onExport` should be renamed to `export` and changed type from `(options: IExportImageOptions) => void;` to `(options: IExportImageOptions) => Promise<void>;`

and you need to await it here

Wroud (changes requested):

Wroud (dismissed):

devnaumov (dismissed):

SychevAndrey (dismissed):

devnaumov (dismissed):

Wroud (dismissed):

devnaumov (dismissed):

Wroud (approved):

PR #3823, opened 2025-10-13
6294 cb reordering columns UI got worse

SychevAndrey:
https://github.com/user-attachments/assets/e3fb973d-9a84-4f34-9328-200b75f083b1

Used react-grid native events. Our hooks are buggy and don't work.


Wroud (changes requested):
we can't have 3 dnd implementations (2 in the react-data-grid) and one in the `plugin-data-spreadsheet-new`
The goal of this ticket is to fix dnd styles in the `plugin-data-spreadsheet-new`

Wroud:
fixed here: https://github.com/dbeaver/cloudbeaver/pull/3824

PR #3824, opened 2025-10-14
dbeaver/pro#6294 fix: dnd styles

Wroud:
closes dbeaver/pro#6294


SychevAndrey (dismissed):
This ticket is about UI. How does this fix make it better? It looks broken

https://github.com/user-attachments/assets/d5aa766d-2ed9-4aac-995a-9462bd09705b

Comparing [to that](https://github.com/dbeaver/cloudbeaver/pull/3823) I can't drop on the header, I need to target text inside + the border is short and ugly and hard to see.

Wroud:
this is exactly how it was working before it broke

devnaumov (approved):

sergeyteleshev (approved):

SychevAndrey:
> this is exactly how it was working before it broke

Why not to make it better? When did this point work at all?

PR #3829, opened 2025-10-15
dbeaver/pro#6026 feat: merge value panel tabs in one group

SychevAndrey:
move all text representations to the first tab level specific representation is used if present (boolean, image) auto type detection is used only for text

https://github.com/user-attachments/assets/0cbac55c-efc7-4e93-ab20-a84d7f5e89bc


webapp/packages/plugin-data-viewer/src/TableViewer/ValuePanel/DataValuePanelService.ts
```diff
@@ -13,6 +13,7 @@ import type { IDatabaseDataModel } from '../../DatabaseDataModel/IDatabaseDataMo
 
 export interface IDataValuePanelOptions {
   dataFormat: ResultDataFormat[];
+  contentType?: string;
```
Wroud: This is hacky. Can you remove it? (it duplicates key)

webapp/packages/plugin-data-viewer/src/TableViewer/ValuePanel/ValuePanel.tsx
```diff
@@ -58,12 +61,27 @@ export const ValuePanel: DataPresentationComponent = observer(function ValuePane
...
+
   let currentTabId = state.currentTabId;
 
-  const hasCurrentTabCells = currentTabId && displayed.some(tab => tab.key === currentTabId);
+  const hasCurrentTab = currentTabId && displayed.some(tab => tab.key === currentTabId);
 
-  if (displayed.length > 0 && !hasCurrentTabCells) {
-    currentTabId = displayed[0]!.key;
+  if (displayed.length > 0 && !hasCurrentTab) {
+    currentTabId = hasOnlyTextRepresentations ? autoContentType : displayed[0]!.key;
```
Wroud: What is the purpose of the `hasOnlyTextRepresentations`?

I think logic should be like this:
Try to find a tab for `autoContentType`, if presented, use it; if not, fallback to the first tab

webapp/packages/plugin-data-viewer/src/TableViewer/ValuePanel/ValuePanel.tsx
```diff
@@ -58,12 +61,27 @@ export const ValuePanel: DataPresentationComponent = observer(function ValuePane
...
+    if (currentTabId !== state.currentTabId) {
+      state.setCurrentTabId(currentTabId);
+    }
```
Wroud: This is also something new, I think it will break the previous behavior:
If the user manually selects a tab, it will save it in the state; otherwise, the tab will be determined automatically.
This change will set the automatically determined tab as the user-selected one. This is wrong

webapp/packages/plugin-data-viewer/src/ValuePanelPresentation/TextValue/TextValuePresentation.tsx
```diff
@@ -8,32 +8,27 @@
...
-import { type TabContainerPanelComponent, TabList, TabsState, TabStyles, useTabLocalState } from '@cloudbeaver/core-ui';
+import { type TabContainerPanelComponent, useTabLocalState } from '@cloudbeaver/core-ui';
 
 import { ResultSetDataContentAction } from '../../DatabaseDataModel/Actions/ResultSet/ResultSetDataContentAction.js';
 import { DataViewerService } from '../../DataViewerService.js';
 import { isResultSetDataModel } from '../../ResultSet/isResultSetDataModel.js';
 import type { IDataValuePanelProps } from '../../TableViewer/ValuePanel/DataValuePanelService.js';
 import { getDefaultLineWrapping } from './getDefaultLineWrapping.js';
 import { isTextValueReadonly } from './isTextValueReadonly.js';
-import styles from './shared/TextValuePresentation.module.css';
```
Wroud, line 22: Is `TextValuePresentation.module.css` still in use, or can it be removed?

webapp/packages/plugin-data-viewer/src/ValuePanelPresentation/TextValue/TextValuePresentation.tsx
```diff
@@ -8,32 +8,27 @@
...
-import TextValuePresentationTab from './shared/TextValuePresentationTab.module.css';
```
Wroud, line 23: same for `./shared/TextValuePresentationTab.module.css`

Wroud (changes requested):

webapp/packages/plugin-data-viewer/src/TableViewer/ValuePanel/ValuePanel.tsx
SychevAndrey: It's not new. 
image

webapp/packages/plugin-data-viewer/src/TableViewer/ValuePanel/ValuePanel.tsx
SychevAndrey: Ok, we can do like that, by changing useAutoContentType hook behaviour. Now it always returns the value.

webapp/packages/plugin-data-viewer/src/TableViewer/ValuePanel/DataValuePanelService.ts
SychevAndrey: Need to check, it was my way to distinct "original type" from the "text type" representation. I can search id in the list of text representations or smth like this.

SychevAndrey: changed to a boolean flag as discussed

webapp/packages/plugin-data-viewer/src/TableViewer/ValuePanel/ValuePanel.tsx
SychevAndrey: refactored the hook

webapp/packages/plugin-data-viewer/src/TableViewer/ValuePanel/ValuePanel.tsx
SychevAndrey: removed

Wroud (dismissed):

webapp/packages/plugin-data-viewer/src/ValuePanelPresentation/TextValue/useContentType.ts
```diff
@@ -55,24 +54,27 @@ function preprocessDefaultContentType(contentType: string | null | undefined) {
...
-    currentContentType = activeTabs[0]!.key;
+  const hasCurrentTab = currentContentType && displayed.some(tab => tab.key === currentContentType);
+
+  if (!hasCurrentTab) {
+    currentContentType = isEnableAutoType && currentContentType ? currentContentType : displayed[0]!.key;
   }
 
-  return currentContentType;
+  return currentContentType ?? DEFAULT_CONTENT_TYPE;
 }
```
sergeyteleshev, line 40: previously the logic was:

1. map the tab value from pure functions
2. check if the tab exists. if yes return, if no return default value

right now the logic became too complicated to understand what is going on because we added the logic of auto calculating content type. it seems tricky and it is easy to lost some case here

may be we can incapsulate this new logic in another hook or pure function? moreover we can cover it with unit tests later

sergeyteleshev: for now it feels like one more line of code here and something is gonna be broken

SychevAndrey: what I actually added is `const isEnableAutoType = displayed.every(tab => tab.options?.isTextPresentation);`

Previously we did not need this logic at all, because auto content type worked only on the second level of tabs. Now we have a condition when we enable it.

sergeyteleshev: I didnt test it but something like this

```
function useAutoContentType({ displayed, formatAction, model, elementKey }: UseContentTypeArgs): string | null {
  const isAllTextType = displayed.every(tab => tab.options?.isTextPresentation);
  const isAutoType = isAllTextType && formatAction && isResultSetDataModel(model);

  if (!isAutoType) {
    return null;
  }

  const contentValue = elementKey ? formatAction.get(elementKey) : null;
  const contentValueType = getContentTypeFromResultSetValue(contentValue);

  return preprocessDefaultContentType(contentValueType);
}

export function useContentType({ model, formatAction, currentContentType, elementKey, displayed }: UseContentTypeArgs): string {
  const autoContentType = useAutoContentType({ displayed, formatAction, model, elementKey, currentContentType });

  if (displayed.length === 0) {
    return currentContentType ?? DEFAULT_CONTENT_TYPE;
  }

  currentContentType = currentContentType ?? autoContentType;

  const hasCurrentTab = displayed.some(tab => tab.key === currentContentType);

  if (!hasCurrentTab) {
    currentContentType = currentContentType ?? displayed[0]!.key;
  }

  return currentContentType ?? DEFAULT_CONTENT_TYPE;
}
```

sergeyteleshev (dismissed):

Wroud (approved):

PR #3831, opened 2025-10-15
dbeaver/pro#6364 add option to set default view for project

HocKu7:
Closes https://github.com/dbeaver/pro/issues/6364


webapp/packages/core-resource-manager/src/SharedProjectsResource.ts
```diff
@@ -147,6 +147,53 @@ export class SharedProjectsResource extends CachedMapResource<string, SharedProj
...
+      });
+    } else {
+      await this.graphQLService.sdk.deleteUserProjectSettings({
+        projectId,
+        settingIds,
+      });
+    }
+
+    this.markOutdated(projectId);
+  }
```
sergeyteleshev: it is not in use yet. may be delete it?

webapp/packages/core-resource-manager/src/SharedProjectsResource.ts
```diff
@@ -147,6 +147,53 @@ export class SharedProjectsResource extends CachedMapResource<string, SharedProj
...
+      return this.graphQLService.sdk.getAdminUserProjectSettings({
+        projectId,
+        subjectId,
+      });
+    } else {
+      return this.graphQLService.sdk.getUserProjectSettings({
+        projectId,
+      });
+    }
+  }
```
sergeyteleshev: does it work for users?

I can see this in constructor which means this resource only available for an admin
`sessionPermissionsResource.require(this, EAdminPermission.admin);`

sergeyteleshev: Probably better to split into two resources
But user one is not gonna be in use, so maybe we just simplty delete the user part

sergeyteleshev (changes requested):

server/bundles/io.cloudbeaver.service.rm/schema/service.rm.graphqls
```diff
@@ -69,6 +69,9 @@ extend type Query {
 
     "Returns project settings that are specified for a current user"
     rmUserProjectSettings(projectId: ID!): Object @since(version: "25.2.1")
+
+    "Returns project settings that are specified for a subject"
+    rmAdminUserProjectSettings(projectId: ID!, subjectId: String!): Object @since(version: "25.2.3")
```
Wroud, line 74: maybe `rmAdminDefaultProjectSettings`?

webapp/packages/core-resource-manager/src/SharedProjectsResource.ts
```diff
@@ -147,6 +147,53 @@ export class SharedProjectsResource extends CachedMapResource<string, SharedProj
...
+        settings,
+      });
+    } else {
+      await this.graphQLService.sdk.addUserProjectSettings({
+        projectId,
+        settings,
+      });
+    }
+
+    this.markOutdated(projectId);
```
Wroud, line 164: you don't need to mark data as outdated, you can use returned information to update changed data

Wroud (changes requested):

webapp/packages/core-resource-manager/src/SharedProjectsResource.ts
SychevAndrey: Sorry I don't see what we should update here.

Wroud: hm, this is because you don't cache project settings, so you have two options:
1. remove `markOutdated` because `RMProject` type not changed so we don't need to reset cache for it
2. create separate resource for project settings

PR #3838, opened 2025-10-17
dbeaver/pro#7081 feat: conditional formatting for data editor

Wroud:
closes dbeaver/pro#7081


webapp/packages/plugin-data-spreadsheet-new/src/DataGrid/DataGridContextMenu/DataGridContextMenuCellEditingService.ts
```diff
@@ -94,12 +97,12 @@ export class DataGridContextMenuCellEditingService {
         const editor = model.source.getAction(resultIndex, IDatabaseDataEditAction);
         const select = model.source.tryGetAction(resultIndex, IDatabaseDataSelectAction);
 
-        const cellValue = view.getCellValue(key);
+        const cellHolder = view.getCellValue(key);
```
SychevAndrey: Do we still want to call this method getCellValue even though it's not returning a value anymore?

webapp/packages/plugin-data-viewer-conditional-formatting/src/formatting/DEFAULT_FORMAT_RULES.ts
```diff
@@ -0,0 +1,227 @@
...
+    parameters: {
+      value: VALUE_PARAMETER,
+    },
+    formatting: {
+      ['background-color']: COLOR_PARAMETER,
+      ['font-weight']: FONT_WEIGHT_PARAMETER,
+      ['font-style']: FONT_STYLE_PARAMETER,
+      ['text-decoration']: TEXT_DECORATION_PARAMETER,
+    },
+    match: ({ text }, { value: pattern }) => typeof text === 'string' && typeof pattern === 'string' && text.toLowerCase() === pattern.toLowerCase(),
```
SychevAndrey, line 151: Don't we need a checkbox to set register sensitivity in case of exact match? It might be important

webapp/packages/plugin-data-viewer-conditional-formatting/src/getMiddleColor.ts
```diff
@@ -0,0 +1,42 @@
...
+  const g1 = data1[1] ?? 0;
+  const b1 = data1[2] ?? 0;
+
+  // Parse second color
+  ctx.fillStyle = maxColor;
+  ctx.fillRect(0, 0, 1, 1);
+  const data2 = ctx.getImageData(0, 0, 1, 1).data;
+  const r2 = data2[0] ?? 0;
+  const g2 = data2[1] ?? 0;
+  const b2 = data2[2] ?? 0;
```
SychevAndrey: Do we really need those 0 fallbacks?

Looks like repetition and we could wrap it in parse function with color argument. Plus use `[r ,g ,b] = ctx....data`

webapp/packages/plugin-data-viewer-conditional-formatting/src/getMiddleColor.ts
```diff
@@ -0,0 +1,42 @@
...
+
+  // Calculate middle values
+  const rMid = Math.round((r1 + r2) / 2);
+  const gMid = Math.round((g1 + g2) / 2);
+  const bMid = Math.round((b1 + b2) / 2);
+
+  // Return as hex color
+  return `#${rMid.toString(16).padStart(2, '0')}${gMid.toString(16).padStart(2, '0')}${bMid.toString(16).padStart(2, '0')}`;
```
SychevAndrey, line 35: also looks like a good candidate for toHex utility fn

SychevAndrey (commented):
Very nice feature! Great job! 🚀 
Just a couple of comments:

webapp/packages/plugin-data-viewer-conditional-formatting/src/formatting/DEFAULT_FORMAT_RULES.ts
Wroud: we might wait for request from users, in google sheets there is no such option

webapp/packages/core-blocks/src/Cell.tsx
```diff
@@ -22,27 +24,39 @@ interface Props {
...
+  ripple = true,
+  big,
+  children,
+  ...rest
+}) {
   const styles = useS(style);
 
   return (
-    <div className={s(styles, { ripple, big }, className)}>
+    <Clickable {...rest} as={asElement ?? 'div'} className={s(styles, { ripple, big }, className)}>
```
sergeyteleshev, line 41: why clickable? maybe some tailwind styles would be enough?

webapp/packages/core-blocks/package.json
```diff
@@ -45,6 +45,7 @@
     "react-hotkeys-hook": "^5",
     "react-minisearch": "^7",
     "reakit": "^1",
+    "reakit-system": "^0",
```
sergeyteleshev, line 48: why do we need it?

webapp/packages/core-theming/src/SystemThemeService.ts
```diff
@@ -42,6 +42,9 @@ export class SystemThemeService extends Bootstrap {
       get class(): string {
         return systemThemeService.dynamicTheme?.class || '';
       },
+      get type() {
+        return systemThemeService.dynamicTheme?.type || 'light';
```
sergeyteleshev, line 46:
```suggestion
        return systemThemeService.dynamicTheme?.type || themeService.themes.find(t => t.id === FALLBACK_THEME_ID)?.type || 'light';
```

webapp/common-react/@dbeaver/ui-kit/src/Select/SelectField.tsx
```diff
@@ -122,20 +132,30 @@ export function SelectField<T, ItemType extends {} = SelectItem<T>>({
...
+    setSelectedValue(newItemValue);
+    onChange?.(newItemValue);
   };
 
   const currentValue = value !== undefined ? value : selectedValue;
-
-  const selectedItem = items.find(item => getItemValue(item) === currentValue);
-
+  const currentValueSerialized = currentValue !== undefined 
+    ? (itemValueSerialized ? itemValueSerialized(currentValue) : JSON.stringify(currentValue))
```
devnaumov: Please avoid nesting ternary expressions, our linter flags them, and they’re hard to read

webapp/packages/plugin-data-viewer-conditional-formatting/src/DVConditionalFormattingPluginBootstrap.ts
```diff
@@ -0,0 +1,43 @@
...
+
+  override register(): void {
+    this.registerPresentation();
+  }
+
+  private registerPresentation(): void {
+    this.dataPresentationService.add({
+      id: 'conditional-formatting-presentation',
+      type: DataPresentationType.toolsPanel,
+      title: 'Conditional formatting',
```
devnaumov: Please translate

webapp/packages/plugin-data-viewer-conditional-formatting/src/GridConditionalFormattingAction.ts
```diff
@@ -0,0 +1,251 @@
...
+          } else {
+            const t = normalize(cellValue, midV, maxValue);
+            formatting.backgroundColor = getColorMix(midColor, maxColor, t);
+            formatted = true;
+          }
+        }
+      }
+    }
+
+    return formatted ? formatting : null;
```
devnaumov, line 180: Seems like you dont need formatted flag, you can use Object.keys(formatting).length > 0

webapp/packages/plugin-data-viewer-conditional-formatting/src/GridConditionalFormattingAction.ts
```diff
@@ -0,0 +1,251 @@
...
+      if (rule.column && !GridDataKeysUtils.isEqual(rule.column, ALL_COLUMNS.key) && !GridDataKeysUtils.isEqual(rule.column, key.column)) {
+        continue;
+      }
+
+      const ruleData = rule.type === 'scale' ? COLOR_SCALE_RULE : DEFAULT_FORMAT_RULES.find(r => r.id === rule.ruleId);
+      if (!ruleData) {
+        continue;
+      }
+
+      if (rule.type === 'single') {
```
devnaumov, line 122: Seems like there is too much logic. We can abstract this logic using fabric pattern and return formatting from there. We can create abstract class something like FormattingRule with common logic and create separate classes for each rule type.

Or we can create some additional methods to split this method's logic

```
createFormatting(rule: IFormatRule): Formatting | null {
    if (rule.type === 'single') {
      return new SingleColorFormatting(rule);
    }
    
    if (rule.type === 'scale') {
      return new ScaleColorFormatting(rule);
    }
    
    return null;
  }
```

webapp/packages/plugin-data-viewer-conditional-formatting/src/LocaleService.ts
```diff
@@ -0,0 +1,37 @@
+/*
+ * CloudBeaver - Cloud Database Manager
+ * Copyright (C) 2020-2024 DBeaver Corp and others
```
devnaumov: Please update license year

devnaumov (changes requested):

webapp/packages/plugin-data-viewer-conditional-formatting/src/formatting/getValueForRuleParameter.ts
```diff
@@ -0,0 +1,29 @@
...
+  state: Record<string, any>,
+  parameters: Record<string, IFormatRuleParameter> | undefined,
+  key: string,
+): T | undefined;
+export function getValueForRuleParameter<T extends string | boolean | number>(
+  state: Record<string, any>,
+  parameters: Record<string, IFormatRuleParameter> | undefined,
+  key: string,
+  fallback?: T,
+): T | undefined {
```
devnaumov, line 27: Do we really need overloads here? Maybe we can use fallback logic in place where we execute it

webapp/packages/plugin-data-viewer-conditional-formatting/src/getSurfaceColor.ts
```diff
@@ -0,0 +1,12 @@
+/*
+ * CloudBeaver - Cloud Database Manager
+ * Copyright (C) 2020-2025 DBeaver Corp and others
+ *
+ * Licensed under the Apache License, Version 2.0.
+ * you may not use this file except in compliance with the License.
+ */
+
+// TODO: implement universal color palette for conditional coloring and use it instead
+export function getSurfaceColor(themeType: 'dark' | 'light' | undefined): string {
```
devnaumov: Please do not hardcode themeType use interface from theme package

webapp/packages/plugin-data-viewer-conditional-formatting/src/locales/zh.ts
```diff
@@ -0,0 +1,40 @@
+export default [
+  ['plugin_data_viewer_conditional_formatting_all_columns', '所有列'],
```
devnaumov, line 4: Please do not translate locales using AI as we cant be sure the AI grabs context right

webapp/packages/plugin-data-viewer-conditional-formatting/src/presentation/ConditionalFormattingPresentation.tsx
```diff
@@ -0,0 +1,157 @@
...
+    ALL_COLUMNS,
+    ...viewAction.columnKeys.map<IColumnInfo>(col => ({
+      name: viewAction.getColumnName(col) || 'unknown',
+      key: col,
+    })),
+  ];
+
+  return (
+    <div className="theme-background-secondary theme-text-on-secondary tw:relative tw:flex-1 tw:flex tw:overflow-hidden">
+      {selectedRule ? (
```
devnaumov, line 59: I suggest to create two separate components and move logic inside  {selectedRule ? (... there. This way component will be easier to read and work with

webapp/packages/plugin-data-viewer-conditional-formatting/src/presentation/FormattingPreview.tsx
```diff
@@ -0,0 +1,64 @@
...
+import { observer } from 'mobx-react-lite';
+
+interface Props {
+  rule: IFormatRule;
+  state: IFormatRuleState;
+  text?: string;
+  className?: string;
+}
+
+export const FormattingPreview = observer(function FormattingPreview({ rule, state, text, className }: Props) {
```
devnaumov: Please pass Props type like this `observer<Props>`

webapp/packages/plugin-data-viewer-conditional-formatting/src/presentation/FormattingRule.tsx
```diff
@@ -0,0 +1,90 @@
...
+interface Props {
+  columns: IColumnInfo[];
+  state: IFormatRuleState;
+  preview?: boolean;
+  selected?: boolean;
+  onSelect?: () => void;
+  onDelete?: () => void;
+}
+
+export const FormattingRule = observer(function FormattingRule({ columns, state, preview, selected, onSelect, onDelete }: Props) {
```
devnaumov: same

webapp/packages/plugin-data-viewer-conditional-formatting/src/presentation/rule-form/ColorScalePointFields.tsx
```diff
@@ -0,0 +1,85 @@
...
+import { Container, InputField, Select, useTranslate } from '@cloudbeaver/core-blocks';
+import { observer } from 'mobx-react-lite';
+import type { IFormatRuleParameter } from '../../formatting/IFormatRule.js';
+
+interface Props {
+  typeParam?: IFormatRuleParameter;
+  valueParam?: IFormatRuleParameter;
+  colorParam?: IFormatRuleParameter;
+  defaultColorValue?: string;
+  state: any;
```
devnaumov: Please avoid using any as prop type.
Please declare prop fields starting from required fields so it would be easier to read

webapp/packages/plugin-data-spreadsheet-new/src/DataGrid/DataGridContextMenu/DataGridContextMenuCellEditingService.ts
Wroud: renamed

webapp/packages/plugin-help/src/WelcomeDocs.tsx
```diff
@@ -17,21 +17,19 @@ export const WelcomeDocs = observer(function WelcomeDocs() {
...
-    <a
+    <Cell
+      as="a"
       href={WEBSITE_LINKS.getDocumentationPage(serverConfigResource.distributed, GlobalConstants.version)}
       target="_blank"
       rel="noreferrer"
-      className="tw:text-inherit! tw:no-underline!"
+      before={<IconOrImage icon="/icons/documentation_link.svg" />}
+      description={translate('plugin_help_welcome_docs_description', undefined, {
+        product: serverConfigResource.distributed ? 'DBeaver Team Edition' : 'CloudBeaver',
```
devnaumov: Why not use product_full_name locale?

devnaumov (changes requested):

webapp/packages/plugin-data-viewer-conditional-formatting/src/GridConditionalFormattingAction.ts
Wroud: actually it was initial implementation, but i've intentionally replaced it with boolean flag because [it's much faster](https://jsbenchmark.com/#eyJjYXNlcyI6W3siaWQiOiJKWXZKNm4ydHVSb0dzVjZ6eE1WM18iLCJjb2RlIjoiY29uc3Qgb2JqID0ge31cblxub2JqLmE9MTtcbm9iai5iPTI7XG5vYmouYz0zO1xuXG5PYmplY3Qua2V5cyhvYmopLmxlbmd0aCA-IDAiLCJuYW1lIjoiT2JqZWN0LmtleXMiLCJkZXBlbmRlbmNpZXMiOltdfSx7ImlkIjoieTlaMWh1UFBwQllHM1FVVnRtV1hhIiwiY29kZSI6ImNvbnN0IG9iaiA9IHt9XG5sZXQgZmxhZyA9IGZhbHNlO1xub2JqLmE9MTtcbmZsYWcgPSB0cnVlO1xub2JqLmI9MjtcbmZsYWcgPSB0cnVlO1xub2JqLmM9MztcbmZsYWcgPSB0cnVlO1xuXG5mbGFnIiwibmFtZSI6IkJvb2xlYW4iLCJkZXBlbmRlbmNpZXMiOltdfV0sImNvbmZpZyI6eyJuYW1lIjoiQmFzaWMgZXhhbXBsZSIsInBhcmFsbGVsIjp0cnVlLCJnbG9iYWxUZXN0Q29uZmlnIjp7ImRlcGVuZGVuY2llcyI6W119LCJkYXRhQ29kZSI6IiJ9fQ)

webapp/packages/plugin-data-viewer-conditional-formatting/src/formatting/getValueForRuleParameter.ts
Wroud: if it's common operation with this function, why not implement it as a part of a function?

webapp/packages/plugin-data-viewer-conditional-formatting/src/getMiddleColor.ts
Wroud: changed a bit

webapp/packages/plugin-data-viewer-conditional-formatting/src/locales/zh.ts
Wroud: I think we might discuss it with team, what better AI translated or english fallback, because it indeed might not understand context correctly, but this will be same as if user will use built-in browser translate for english (that might break react applications)

webapp/packages/plugin-help/src/WelcomeDocs.tsx
Wroud: because it's different, this was discussed with team in Slack previously

Wroud: but probably for TE we can use token

webapp/packages/plugin-data-viewer-conditional-formatting/src/GridConditionalFormattingAction.ts
Wroud: this is interesting idea, but for now there is not so many logic this probably will be more actual with full functionality

webapp/common-react/@dbeaver/ui-kit/src/Select/SelectField.tsx
```diff
@@ -23,6 +23,12 @@ export interface SelectFieldProps<T, ItemType = SelectItem<T>> {
   /** Options array - can be SelectOption objects or arbitrary objects */
   items: ItemType[];
 
+  /**
+   * Function to extract serialized value from items
+   * Example: (key) => JSON.stringify(key)
+   */
+  itemValueSerialized?: PropertyGetter<T, string>;
```
sergeyteleshev, line 30: why do we need it? 

we use it in `core-blocks` selector
but `core-blocks` selector prop do not reused it anywhere else

seems like it left from previous implementations. I suggest to delete it since it is not in use yet

webapp/packages/plugin-data-viewer-conditional-formatting/src/presentation/rule-form/ColorScaleRuleForm.tsx
```diff
@@ -0,0 +1,88 @@
...
+import type { IColumnInfo } from '../../formatting/IColumnInfo.js';
+import { getMiddleColor } from '../../getMiddleColor.js';
+import { ColorScalePreview } from '../ColorScalePreview.js';
+import { useColorScaleParameters } from './useColorScaleParameters.js';
+import { ColorScalePointFields } from './ColorScalePointFields.js';
+
+interface Props {
+  columns: IColumnInfo[];
+  state: IFormatRuleState;
+  onAddRule?: () => void;
```
SychevAndrey: don't need it anymore

webapp/packages/plugin-data-viewer-conditional-formatting/src/presentation/rule-form/SingleColorRuleForm.tsx
```diff
@@ -0,0 +1,63 @@
...
+import { observer } from 'mobx-react-lite';
+import { DEFAULT_FORMAT_RULES } from '../../formatting/DEFAULT_FORMAT_RULES.js';
+import type { IFormatRuleState } from '../../formatting/IFormatRuleState.js';
+import { RenderParameter } from './RenderParameter.js';
+import type { IColumnInfo } from '../../formatting/IColumnInfo.js';
+
+interface Props {
+  columns: IColumnInfo[];
+  state: IFormatRuleState;
+  onAddRule?: () => void;
```
SychevAndrey: same

SychevAndrey (commented):
by the way, do we all agree to use name "holder"? it's not very common in JS, I've never heard of that pattern before.

SychevAndrey (dismissed):

webapp/packages/plugin-data-viewer-conditional-formatting/src/locales/en.ts
```diff
@@ -0,0 +1,42 @@
...
+  ['plugin_data_viewer_conditional_formatting_column', 'Column'],
+  ['plugin_data_viewer_conditional_formatting_add_another_rule', 'Add another rule'],
+  ['plugin_data_viewer_conditional_formatting_no_rules', 'No formatting rules applied'],
+  ['plugin_data_viewer_conditional_formatting_single_color', 'Single color'],
+  ['plugin_data_viewer_conditional_formatting_color_scale', 'Color scale'],
+  ['plugin_data_viewer_conditional_formatting_select_rule', 'Select rule'],
+  ['plugin_data_viewer_conditional_formatting_format_cells_if', 'Format cells if...'],
+
+  ['plugin_data_viewer_conditional_formatting_parameter_color', 'Color'],
+  ['plugin_data_viewer_conditional_formatting_parameter_value', 'Value of formula'],
```
sergeyteleshev, line 15: Value of **the** formula

webapp/packages/plugin-data-viewer-conditional-formatting/src/locales/en.ts
```diff
@@ -0,0 +1,42 @@
...
+  ['plugin_data_viewer_conditional_formatting_parameter_text_decoration_overline', 'Overline'],
+  ['plugin_data_viewer_conditional_formatting_parameter_text_decoration_line_through', 'Line through'],
+
+  ['plugin_data_viewer_conditional_formatting_rule_is_empty', 'Is empty'],
+  ['plugin_data_viewer_conditional_formatting_rule_is_not_empty', 'Is not empty'],
+  ['plugin_data_viewer_conditional_formatting_rule_text_contains', 'Text contains'],
+  ['plugin_data_viewer_conditional_formatting_rule_text_does_not_contain', 'Text does not contain'],
+  ['plugin_data_viewer_conditional_formatting_rule_text_starts_with', 'Text starts with'],
+  ['plugin_data_viewer_conditional_formatting_rule_text_ends_with', 'Text ends with'],
+  ['plugin_data_viewer_conditional_formatting_rule_text_is_exactly', 'Text is exactly'],
```
sergeyteleshev, line 30: **Text is equal**
**Equals to**

webapp/packages/plugin-data-viewer-conditional-formatting/src/locales/en.ts
```diff
@@ -0,0 +1,42 @@
...
+
+  ['plugin_data_viewer_conditional_formatting_rule_color_scale', 'Color scale'],
+  ['plugin_data_viewer_conditional_formatting_parameter_minpoint', 'Minpoint'],
+  ['plugin_data_viewer_conditional_formatting_parameter_midpoint', 'Midpoint'],
+  ['plugin_data_viewer_conditional_formatting_parameter_maxpoint', 'Maxpoint'],
```
sergeyteleshev, line 35: **Min-point
Mid-point
Max-point**

webapp/packages/plugin-data-viewer-conditional-formatting/src/locales/zh.ts
sergeyteleshev: we should talk about it, sure thing. I usually double check it with Google Translate and it generates fine translation from this point of view. But it only applies to a popular languages. For example, Serbian unfortunately translates badly to other ones so there is space to create a mistake 100%

webapp/packages/plugin-data-viewer-conditional-formatting/src/presentation/rule-form/SingleColorRuleForm.tsx
```diff
@@ -0,0 +1,62 @@
...
+        <RenderParameter key={param.key} parameter={param} state={state.parameters} />
+      ))}
+
+      {onDelete && (
+        <GroupItem>
+          <Button icon="delete" iconPlacement="start" type="button" variant="secondary" onClick={onDelete}>
+            {t('ui_delete')}
+          </Button>
+        </GroupItem>
+      )}
```
sergeyteleshev, line 59: I would add a confirmation popup on delete to exclude missclick cases
In case of single color it is not sensible yet (small amount of settings), but in `color scale` we have more settings and it is much harder loss of clicked data in case of missclick

webapp/packages/plugin-data-viewer-conditional-formatting/src/presentation/FormattingRule.tsx
```diff
@@ -0,0 +1,90 @@
...
+      />
+    );
+  }
+
+  return (
+    <Cell
+      ref={ref}
+      before={<FormattingPreview state={state} rule={rule} text="A" />}
+      description={description}
+      after={<ActionIconButton name="delete" title={t('ui_delete')} onClick={handleClick} />}
```
sergeyteleshev, line 81: lets add closing rule on mouse middle click?
like we do for the tabs

webapp/packages/plugin-data-viewer-conditional-formatting/src/presentation/useRuleTransition.ts
```diff
@@ -0,0 +1,58 @@
...
+      });
+
+      const timer = setTimeout(() => {
+        setPreviousRule(null);
+      }, ANIMATION_DURATION);
+
+      prevRuleRef.current = selectedRule ? { rule: selectedRule, index: selectedRuleIndex } : null;
+      return () => clearTimeout(timer);
+    }
+    return undefined;
```
sergeyteleshev, line 54: all in call this logic seems like overhead to calculate stuff just for an animation. it feels like mostly this case can be solved on css level + (component rendered / unmounted / changed prop)

My suggestion here is to remove this logic for now and talk about animations and best practices, where and how we want to spread it all over the app at next frontend call

webapp/packages/plugin-data-viewer-conditional-formatting/src/DVConditionalFormattingPluginBootstrap.ts
```diff
@@ -0,0 +1,43 @@
...
+      type: DataPresentationType.toolsPanel,
+      title: 'plugin_data_viewer_conditional_formatting_presentation_title',
+      icon: '/icons/conditional_formatting_sm.svg',
+      hidden: (dataFormat, model, resultIndex) => {
+        if (!model.source.hasResult(resultIndex)) {
+          return true;
+        }
+
+        const data = model.source.tryGetAction(resultIndex, IDatabaseDataResultAction, GridDataResultAction);
+        return data?.empty ?? true;
```
sergeyteleshev: new tool seems to be working in combination with data editor table. may be we should hide it if we do not see a table

webapp/packages/plugin-data-viewer-conditional-formatting/src/GridConditionalFormattingAction.ts
```diff
@@ -0,0 +1,251 @@
...
+        const minValue = this.resolveStopValue(ruleData, rule, 'min', dataMin, dataMax, sortedValues)!;
+        const maxValue = this.resolveStopValue(ruleData, rule, 'max', dataMin, dataMax, sortedValues)!;
+        const midValue = this.resolveStopValue(ruleData, rule, 'mid', dataMin, dataMax, sortedValues);
+
+        const themeType = this.themeService.currentTheme?.type;
+        const minColor = getValueForRuleParameter<string>(rule.parameters, ruleData.formatting, 'min-color', getSurfaceColor(themeType));
+        const maxColor = getValueForRuleParameter<string>(rule.parameters, ruleData.formatting, 'max-color', '#00FF00');
+
+        if (midValue === null) {
+          const t = normalize(cellValue, minValue, maxValue);
```
sergeyteleshev: ratio

webapp/packages/plugin-data-viewer-conditional-formatting/src/GridConditionalFormattingAction.ts
```diff
@@ -0,0 +1,251 @@
...
+    if (n === 1) {
+      return sorted[0]!;
+    }
+    const idx = (n - 1) * percentile;
+    const lo = Math.floor(idx);
+    const hi = Math.ceil(idx);
+    const t = idx - lo;
+    return (1 - t) * sorted[lo]! + t * sorted[hi]!;
+  }
+
```
sergeyteleshev: can be moved to a pure function and later covered with unit tests

sergeyteleshev (changes requested):

webapp/packages/core-blocks/src/Cell.tsx
Wroud: In every place where we use this component, we wrapped it with Clickable or Link because of accessibility, so now this component itself is Clickable

webapp/packages/core-blocks/package.json
Wroud: It's needed for internal types from reakit (for clickable) that are not exposed

webapp/common-react/@dbeaver/ui-kit/src/Select/SelectField.tsx
Wroud: It's needed because Ariakit's implementation of Select doesn't support objects as values. Still, our outer types of Select support passing objects as values, so this is needed to serialize objects to a string. Its default is JSON.stringify, but it might be incorrect for some cases and needs to be customized. This props for this.

webapp/packages/plugin-data-viewer-conditional-formatting/src/locales/en.ts
Wroud: Screenshot 2025-10-21 at 9 41 42 PM

it's equal to google sheets right now

webapp/packages/plugin-data-viewer-conditional-formatting/src/locales/en.ts
Wroud: same as Google sheets

webapp/packages/plugin-data-viewer-conditional-formatting/src/locales/en.ts
Wroud: same as in google sheets

webapp/packages/plugin-data-viewer-conditional-formatting/src/presentation/useRuleTransition.ts
Wroud: Unfortunately, in the current component without animation, it might be hard to understand if the "Add rule" button did anything. I don't think that we need to remove this code just because it's "overhead"; if you know how to implement it in a simpler way, please suggest it.

webapp/packages/plugin-data-viewer-conditional-formatting/src/GridConditionalFormattingAction.ts
devnaumov: Then  I guess we should leave a comment here as it could be easily removed by the next iteration of different developer in the future or mention this approach in the tech document

webapp/packages/plugin-data-viewer-conditional-formatting/src/presentation/FormattingRule.tsx
Wroud: I'm not sure that this is expected behavior

webapp/packages/plugin-data-viewer-conditional-formatting/src/DVConditionalFormattingPluginBootstrap.ts
Wroud: we don't have API for that right now

webapp/packages/core-blocks/package.json
Wroud: i've created separate ticket for component migration:
https://github.com/dbeaver/pro/issues/7238

this new dependency is not a problem because it used only for types and used by reakit internally

webapp/packages/core-theming/src/SystemThemeService.ts
Wroud: I think we don't need to create fallbacks here; this is more for type safety than for correct behavior

webapp/packages/plugin-data-viewer-conditional-formatting/src/presentation/rule-form/SingleColorRuleForm.tsx
Wroud: I think this will be noisy for the user. This is not such a critical action, and you need to be precise to click the right button on the delete button instead of anything else

sergeyteleshev (dismissed):

devnaumov (dismissed):

webapp/packages/plugin-data-viewer-conditional-formatting/src/GridConditionalFormattingAction.ts
Wroud: comment added

webapp/common-react/@dbeaver/react-data-grid/src/DataGridRowContext.ts
```diff
@@ -1,7 +1,27 @@
...
+
 import { createContext } from 'react';
 import type { IGridReactiveValue } from './IGridReactiveValue.js';
 
+export interface IDataGridRowProps extends React.DetailedHTMLProps<React.HTMLAttributes<HTMLDivElement>, HTMLDivElement> {
+
+}
+
+export interface IDataGridRowRenderer {
+  (propsOverride: Partial<IDataGridRowProps> & { [key: string]: any }): React.ReactNode;
```
SychevAndrey: { [key: string]: any } doesn't look good. Is there any way to make typings better?

webapp/common-react/@dbeaver/react-data-grid/src/BaseRow.tsx
```diff
@@ -12,5 +20,18 @@ export const BaseRow = memo(function BaseRow<TRow, TSummaryRow>(props: RenderRow
       rowContext?.onScrollToBottom?.();
     }
   }, [rowsCount]);
-  return <Row {...props} />;
+
+  const renderDefaultRow = useMemo<IDataGridRowRenderer>(
+    () => override => (
+      <Row
+        {...props}
+        {...override}
```
SychevAndrey, line 28: Do we really want to let to override any prop?

Wroud: this is purpose of this function

sergeyteleshev (approved):

SychevAndrey (approved):

PR #3872, opened 2025-11-05
Add reorder option to tabs #4583

SychevAndrey:
Panels

https://github.com/user-attachments/assets/127b1f9a-38f5-46d9-8751-4d6be9f828e3

Navigator Tabs

https://github.com/user-attachments/assets/4ea913d9-d2fa-4c00-9545-fc47c23819af

ObjectView tabs remain not reorderable. Do we want to move them also?
Screenshot 2025-11-05 at 15 47 00


SychevAndrey:
SqlResultsTabs

https://github.com/user-attachments/assets/8032f306-cb21-4b8d-b255-dbb25872a54c

SychevAndrey:
Changed reorderArray implementation for useHeaderDnD (see implementation and tests). I believe this is the right implementation and we don't use this header dnd yet so it's safe to change it now.

webapp/common-typescript/@dbeaver/js-helpers/src/reorderArray.ts
```diff
@@ -0,0 +1,61 @@
...
+export function reorderArray<T>(array: T[], sourceItemOrFrom: T | number, targetOrTo: number | ReorderItemTarget<T>): T[] {
+  if (typeof sourceItemOrFrom === 'number' && typeof targetOrTo === 'number') {
+    if (sourceItemOrFrom < 0 || sourceItemOrFrom >= array.length || targetOrTo < 0 || targetOrTo >= array.length || sourceItemOrFrom === targetOrTo) {
+      return array;
+    }
+    const newArray = [...array];
+
+    newArray.splice(sourceItemOrFrom, 1);
+    newArray.splice(targetOrTo, 0, array[sourceItemOrFrom]!);
+    return newArray;
```
Wroud: This is the base logic that is repeated. It would be better to implement a simple inner method that only reorders by the provided array, from (index) to (index). You don't need to export this method.

After you need to convert arguments into `array, from, to` and pass it to this inner method
This should simplify logic in general and make it clearer

webapp/packages/core-ui/src/Tabs/TabsContainer/TabsContainer.ts
```diff
@@ -134,4 +134,16 @@ export class TabsContainer<TProps = void, TOptions extends Record<string, any> |
...
+    Object.entries(tabsOrderMap).forEach(([tabId, index]) => {
+      const existingInfo = this.tabInfoMap.get(tabId);
+      if (existingInfo) {
+        this.tabInfoMap.set(tabId, {
+          ...existingInfo,
+          order: index,
+        });
+      }
+    });
+  }
```
Wroud: please do not mutate initial container state, instead you should reorder at the component level

webapp/packages/core-ui/src/Tabs/TabsState.tsx
```diff
@@ -81,6 +86,34 @@ export const TabsState = observer(function TabsState<T = Record<string, any>>({
     manual,
   });
 
+  const tabOrderKey = container ? `tabs-order-${container.areaLabel}` : 'tabs-no-persist';
```
Wroud: `areaLabel` can't be used as a stable storage key, please introduce new prop for TabsState

webapp/packages/core-ui/src/Tabs/TabsState.tsx
```diff
@@ -81,6 +86,34 @@ export const TabsState = observer(function TabsState<T = Record<string, any>>({
...
+    order => {
+      if (container) {
+        container.applyOrder(order);
+      }
+    },
+  );
+
+  const onReorder =
+    onReorderProp ||
+    ((draggedTabId: string, targetTabId: string, position: 'before' | 'after') => {
```
Wroud: please use named function

webapp/packages/core-ui/src/Tabs/useTabDragAndDrop.ts
```diff
@@ -0,0 +1,105 @@
...
+      event.stopPropagation();
+      store.setData('dbeaver-tab-id', tabId);
+    },
+  });
+
+  const { props: dropProps } = useDrop({
+    canDrop: (event, store) => {
+      const draggedTabId = store?.getData('dbeaver-tab-id');
+      return enabled && draggedTabId !== tabId;
+    },
```
Wroud, line 66: This will display like a tab that can be dropped into any of the draggable tab lists. You should use the `reorderStateKey` property from the `TabsState` (for example) to verify that DnD between the same tabs bar

Wroud (changes requested):

webapp/packages/core-ui/src/Tabs/TabsContext.ts
```diff
@@ -38,6 +38,8 @@ export interface ITabsContext<T = Record<string, any>> {
   closeAll: () => Promise<void>;
   closeAllToTheDirection: (tabId: string, direction: TabDirection) => Promise<void>;
   closeOthers: (tabId: string) => Promise<void>;
+  reorder?: (draggedTabId: string, targetTabId: string, position: 'before' | 'after') => void;
+  reorderStateKey?: string;
```
sergeyteleshev, line 44: it is not obvious from api that `reorderStateKey` should be defined so reordering feature can work
may be comment here?

sergeyteleshev (dismissed):

webapp/packages/core-ui/src/SideBarPanel/SideBarPanel.tsx
```diff
@@ -42,9 +43,18 @@ const sideBarPanelRegistry: StyleRegistry = [
 
 export const SideBarPanel = observer<SideBarPanelProps>(function SideBarPanel({ container }) {
   const style = useS(styles);
+  const reorderStateKey = 'side-bar-panel';
```
devnaumov: reorderStateKey can be moved outside the component body since it’s hardcoded and not computed

webapp/packages/core-ui/src/Tabs/Tab/Tab.tsx
```diff
@@ -27,6 +28,19 @@ export const Tab = observer<TabProps>(function Tab({ after, ...props }) {
...
+  const {
+    ref: dragRef,
+    isDragging,
+    dropAllowed,
+    dropPosition,
+    dragProps,
+    dropProps,
+  } = useTabDragAndDrop({
+    tabId,
+    stateKey: tab.reorderStateKey || undefined,
```
devnaumov: Seems like tab.reorderStateKey has type of reorderStateKey?: string so there is no need for fallback?

webapp/packages/core-ui/src/Tabs/useTabDragAndDrop.ts
```diff
@@ -0,0 +1,113 @@
...
+ * Licensed under the Apache License, Version 2.0.
+ * you may not use this file except in compliance with the License.
+ */
+import { useRef, useState } from 'react';
+import { useDrag, useDrop } from '@dbeaver/react-dnd';
+
+export interface ITabDragAndDropOptions {
+  tabId: string;
+  stateKey: string | undefined;
+  onReorder: ((draggedTabId: string, targetTabId: string, position: 'before' | 'after') => void) | null;
```
devnaumov: we duplicate position: 'before' | 'after', probably we want to move it to separate type like TabReorderPosition

webapp/packages/core-ui/src/Tabs/useTabDragAndDrop.ts
```diff
@@ -0,0 +1,113 @@
...
+  const [dropAllowed, setDropAllowed] = useState(false);
+
+  const onReorderRef = useRef(onReorder);
+  onReorderRef.current = onReorder;
+
+  const { isDragging, props: dragProps } = useDrag({
+    draggable: !!stateKey,
+    onDragStart: (event, store) => {
+      event.stopPropagation();
+      store.setData('dbeaver-tab-id', tabId);
```
devnaumov: We repeat this variables across the hook, should we move them to const and use like this?
`DBEAVER_TAB_ID = dbeaver-tab-id`
`store?.getData(DBEAVER_TAB_ID);`

webapp/packages/core-ui/src/SideBarPanel/SideBarPanel.tsx
SychevAndrey: I decided to use container.areaLabel as a key. 
@Wroud container constructor requires that field and I believe they are unique within the app, so why not to use them as keys for now?

webapp/packages/core-ui/src/Tabs/useTabOrderPersistence.ts
```diff
@@ -0,0 +1,56 @@
...
+export function useTabOrderPersistence(persistenceKey: string, getTabIds: () => string[]): ITabOrderPersistenceResult {
+  const tabOrderKey = `tabs-order-${persistenceKey}`;
+  const tabsPersisted = useUserData(tabOrderKey, () => ({}) as Record<string, number>);
+
+  const sortTabs = useCallback(
+    (tabs: string[]): string[] =>
+      tabs.toSorted((a, b) => {
+        const orderA = tabsPersisted[a] ?? Number.MAX_SAFE_INTEGER;
+        const orderB = tabsPersisted[b] ?? Number.MAX_SAFE_INTEGER;
+        return orderA - orderB;
```
Wroud, line 27: because tabs can be dynamic and not guarantee to persist in the order state better to store tabId in the tabsPersisted state, for example:
`['tabId1', 'tabId3', tabId2']` so you can get order position using `tabsPersisted.indexOf(a)`

webapp/packages/core-ui/src/SideBarPanel/SideBarPanel.tsx
```diff
@@ -42,9 +43,18 @@ const sideBarPanelRegistry: StyleRegistry = [
 
 export const SideBarPanel = observer<SideBarPanelProps>(function SideBarPanel({ container }) {
   const style = useS(styles);
+  const { onReorder, sortTabs, persistenceKey } = useTabOrderPersistence(container.areaLabel, () => container.getIdList());
```
Wroud: please don't use areaLable as persistence key, you can add property like `id` to the `SideBarPanel` and use it as persistence key

webapp/packages/core-ui/src/Tabs/Tab/useTab.ts
```diff
@@ -67,6 +67,9 @@ export function useTab(
       get closable() {
         return this.state.canClose(this.tabId);
       },
+      get reorderStateKey() {
+        return this.state.reorderStateKey;
+      },
```
Wroud: probably you don't need it here because it's not tab specific but same for all tabs in the single `TabsContext`

webapp/packages/core-ui/src/Tabs/TabsState.tsx
```diff
@@ -151,7 +163,7 @@ export const TabsState = observer(function TabsState<T = Record<string, any>>({
       tabId: currentSelectedId,
       props,
     });
-  }, [currentSelectedId]);
+  }, [currentSelectedId, dynamic.selectedId, openExecutor, props]);
```
Wroud: This might change behaviour unexpectedly. Probably the purpose of this effect is only to fire `openExecutor` when a new tab is selected, but now it will be fired in many other cases. I recommend not changing this until you encounter a problem with the current behavior.

webapp/packages/plugin-sql-editor/src/SqlResultTabs/SqlResultTabs.tsx
```diff
@@ -78,6 +79,25 @@ export const SqlResultTabs = observer<Props>(function SqlDataResult({ state, onT
...
+
+    const reorderedTabs = reorderArray(state.tabs, draggedTab, {
+      item: targetTab,
+      position,
+    });
+
+    state.tabs = reorderedTabs.map((tab, index) => ({
+      ...tab,
+      order: index,
+    }));
```
Wroud: This is quite inefficient, `reorderArray` will mutate `state.tabs`, which will lead to re-render
`state.tabs = ...map(() => ({...}))` will mutate again and re-create every tab state, and this together will lead to a full re-render of any related data, while it's only  required to update `order` of changed tabs

webapp/packages/plugin-tools-panel/src/ToolsPanel/ToolsPanel.tsx
```diff
@@ -26,6 +26,9 @@ export const ToolsPanel = observer(function ToolsPanel() {
   const style = useS(styles);
 
   const state = useUserData<IToolsState>('tools', () => ({ selectedTabId: undefined }));
+  const { onReorder, sortTabs, persistenceKey } = useTabOrderPersistence(toolsPanelService.tabsContainer.areaLabel, () =>
```
Wroud: please don't use areaLabel for state key

Wroud (changes requested):

webapp/packages/core-ui/src/SideBarPanel/SideBarPanel.tsx
Wroud: @SychevAndrey, because the label might be changed or translated at some point, but the key of state should remain the same

Wroud (approved):

devnaumov (approved):

webapp/packages/core-ui/src/Tabs/useTabDragAndDrop.ts
devnaumov: We fixed it in one place but the type declaration repeats across entire pr)

sergeyteleshev (approved):

PR #3875, opened 2025-11-06
dbeaver/pro#6525 add search to schema and connection lists

devnaumov:
closes [6525](https://github.com/dbeaver/pro/issues/6525)

I disabled the no-sync-component-import rule for ContextMenuSearchItem for two reasons:

1. The current context menu doesn’t open the first time, lazy loading the component somehow breaks its focus lifecycle.
2. The loader UI is broken 

We’ll migrate to Ariakit and try enabling the rule again afterward, since the new menu should handle focus and related behavior more reliably

I also tried adding virtualization using the virtua library that we use in the json view. It works, but it requires markup changes. I think it makes more sense to implement it after the migration, since the layout will definitely change anyway.


webapp/packages/core-view/src/Menu/getMenuItemLabel.ts
```diff
@@ -0,0 +1,32 @@
...
+ * Licensed under the Apache License, Version 2.0.
+ * you may not use this file except in compliance with the License.
+ */
+
+import type { MenuCreatorItem } from './IMenuItemsCreator.js';
+import type { IMenuItem } from './MenuItem/IMenuItem.js';
+import { MenuBaseItem } from './MenuItem/MenuBaseItem.js';
+import { MenuActionItem } from './MenuItem/MenuActionItem.js';
+
+export function getMenuCreatorItemLabel(item: MenuCreatorItem): string {
```
Wroud: Please remove, there can be many different menu item implementations

webapp/packages/plugin-datasource-context-switch/src/ConnectionSchemaManager/ConnectionSchemaManagerBootstrap.ts
Wroud, line 163: just filter here

Wroud (changes requested):

webapp/packages/plugin-datasource-context-switch/src/ConnectionSchemaManager/ConnectionSchemaManagerBootstrap.ts
```diff
@@ -305,24 +313,28 @@ export class ConnectionSchemaManagerBootstrap extends Bootstrap {
...
-                label: title,
-                tooltip: title,
-                icon: '/icons/plugin_datasource_context_switch_schema_sm.svg',
-              },
-              {
-                onSelect: async () => {
-                  await this.connectionSchemaManagerService.selectSchema(title);
+          const excluded = !!filter && !title.toLowerCase().includes(filter.toLowerCase());
+
+          if (!excluded) {
```
Wroud:
```
if (excluded) {
  continue;
}
```

same in places below

Wroud (dismissed):

SychevAndrey (dismissed):

Wroud (approved):

PR #3891, opened 2025-11-12
dbeaver/pro#6416 refactor: adjust height property for treeNodeNested

SychevAndrey:


webapp/packages/core-blocks/src/Tree/TreeNode/TreeNodeNested.module.css
```diff
@@ -8,6 +8,7 @@
 .treeNodeNested {
   box-sizing: border-box;
   padding-left: var(--tree-node-nested-padding, 8px);
+  height: 100%;
```
Wroud: kinda strange style

Wroud (changes requested):

webapp/packages/core-blocks/src/Tree/TreeNode/TreeNodeNested.module.css
devnaumov: is it possible to change only placeholder styles? I would prefer not to change TreeNodeNested as it might affect something different

devnaumov: Maybe something like this. So we can render different placeholders depending on whether it’s a particular node or we don’t have any clouds at all. Also it will hide filter component (we dont need it in case we have nothing to filter)

```
const rootChildren = treeData?.getChildren(root) ?? [];

if (!rootChildren.length) {
      return <TextPlaceholder>{translate('plugin_cloud_explorer_tree_placeholder')}</TextPlaceholder>;
}

return <Tree />
```

SychevAndrey: We need an empty placeholder also for cases when we use filter and the result is empty.

Wroud (approved):

devnaumov (approved):

PR #3897, opened 2025-11-14
[CB] Create new folder pop-up - change focus field

sergeyteleshev:
closes https://github.com/dbeaver/pro/issues/7316

### DIFF
--- a/webapp/packages/plugin-projects/src/FolderDialog.tsx
+++ b/webapp/packages/plugin-projects/src/FolderDialog.tsx
@@ -167,7 +167,6 @@ export const FolderDialog: DialogComponent<FolderDialogPayload, IFolderDialogRes
       <CommonDialogBody>
         <Form ref={focusedRef} onSubmit={resolveHandler}>
           <Container center gap>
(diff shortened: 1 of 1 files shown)


webapp/packages/plugin-projects/src/FolderDialog.tsx
```diff
@@ -178,6 +177,7 @@ export const FolderDialog: DialogComponent<FolderDialogPayload, IFolderDialogRes
             >
               {translate('ui_name')}
             </InputField>
+            {selectProject && <ProjectSelect value={state.projectId} filter={filterProject} onChange={projectId => state.setProjectId(projectId)} />}
```
devnaumov, line 180: As I remember @Wroud specifically put select before the value, we should sync with him

Wroud: I think that's okay, we can also try to manipulate with tabIndex to make the input focused (this is the original issue), without changing input order

Wroud (approved):

SychevAndrey (approved):
I believe that seeing the name as a first field is expected.

PR #3903, opened 2025-11-17
[CB] put new user to top after creation 

sergeyteleshev:
closes https://github.com/dbeaver/pro/issues/7506

### DIFF
--- a/webapp/packages/core-authentication/src/UsersResource.ts
+++ b/webapp/packages/core-authentication/src/UsersResource.ts
@@ -91,7 +91,7 @@ export class UsersResource extends CachedMapResource<string, AdminUser, UserReso
     this.aliases.add(UsersResourceNewUsers, () => {
       const orderedKeys = this.entries
         .filter(k => isNewUser(k[1]))
-        .sort((a, b) => compareUsers(a[1], b[1]))
+        .sort((a, b) => sortUsersById(a[1], b[1]))
         .map(([key]) => key);
       return resourceKeyList(orderedKeys);
     });
@@ -325,10 +325,29 @@ export function isLocalUser(user: AdminUser): boolean {
   return user.origins.some(origin => origin.type === AUTH_PROVIDER_LOCAL_ID);
 }
 
-export function isNewUser(user: AdminUser): user is AdminUserNew {
+export function isNewUser(user: AdminUser | AdminUserNew): user is AdminUserNew {
   return NEW_USER_SYMBOL in user && user[NEW_USER_SYMBOL] === true && 'createdAt' in user && Boolean(user.createdAt);
 }
 
-export function compareUsers<T extends Pick<AdminUser, 'userId'>>(a: T, b: T): number {
+export function sortUsersById<T extends Pick<AdminUser, 'userId'>>(a: T, b: T): number {
   return a.userId.localeCompare(b.userId);
+}
+
+export function sortByNewUsers(a: AdminUser, b: AdminUser): number {
+  const aIsNew = isNewUser(a);
+  const bIsNew = isNewUser(b);
+
+  if (aIsNew && !bIsNew) {
+    return -1;
+  }
+
+  if (!aIsNew && bIsNew) {
+    return 1;
+  }
+
+  if (aIsNew && bIsNew) {
+    return b.createdAt.localeCompare(a.createdAt);
+  }
+
+  return 0;
 }
--- a/webapp/packages/core-authentication/src/UsersResource.ts
+++ b/webapp/packages/core-authentication/src/UsersResource.ts
@@ -333,7 +333,7 @@ export function compareUsers<T extends Pick<AdminUser, 'userId'>>(a: T, b: T): n
   return a.userId.localeCompare(b.userId);
 }
 
-export function sortByNewUsers(a: AdminUser, b: AdminUser): number {
+export function compareNewUsers(a: AdminUser, b: AdminUser): number {
   const aIsNew = isNewUser(a);
   const bIsNew = isNewUser(b);
 
--- a/webapp/packages/core-authentication/src/UsersResource.ts
+++ b/webapp/packages/core-authentication/src/UsersResource.ts
@@ -36,12 +36,12 @@ import { AuthInfoService } from './AuthInfoService.js';
 import { AuthProviderService } from './AuthProviderService.js';
 import type { IAuthCredentials } from './IAuthCredentials.js';
 
-const NEW_USER_SYMBOL = Symbol('new-user');
+export const NEW_USER_SYMBOL = Symbol('new-user');
 
 export type AdminUser = AdminUserInfoFragment;
 export type AdminUserOrigin = AdminUserInfoFragment['origins'][number];
 
-type AdminUserNew = AdminUser & { [NEW_USER_SYMBOL]: boolean };
+export type AdminUserNew = AdminUser & { [NEW_USER_SYMBOL]: boolean; createdAt: string };
 export type UserResourceIncludes = Omit<GetUsersListQueryVariables, 'userId' | 'page' | 'filter'>;
 
 interface IUserResourceFilterOptions {
@@ -148,6 +148,7 @@ export class UsersResource extends CachedMapResource<string, AdminUser, UserReso
 
     const newUser = user as unknown as AdminUserNew;
     newUser[NEW_USER_SYMBOL] = true;
+    newUser.createdAt = new Date().toISOString();
     this.set(user.userId, newUser);
 
     return this.get(user.userId)!;
@@ -155,7 +156,12 @@ export class UsersResource extends CachedMapResource<string, AdminUser, UserReso
 
   cleanNewFlags(): void {
     for (const user of this.data.values()) {
-      (user as AdminUserNew)[NEW_USER_SYMBOL] = false;
+      if (!isNewUser(user)) {
+        continue;
+      }
+
+      user[NEW_USER_SYMBOL] = false;
+      user.createdAt = '';
     }
   }
 
@@ -319,8 +325,8 @@ export function isLocalUser(user: AdminUser): boolean {
   return user.origins.some(origin => origin.type === AUTH_PROVIDER_LOCAL_ID);
 }
 
-export function isNewUser(user: AdminUser): boolean {
-  return NEW_USER_SYMBOL in user && user[NEW_USER_SYMBOL] === true;
+export function isNewUser(user: AdminUser): user is AdminUserNew {
+  return NEW_USER_SYMBOL in user && user[NEW_USER_SYMBOL] === true && 'createdAt' in user && Boolean(user.createdAt);
 }
 
 export function compareUsers<T extends Pick<AdminUser, 'userId'>>(a: T, b: T): number {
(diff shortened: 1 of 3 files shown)


webapp/packages/plugin-authentication-administration/src/Administration/Users/UsersTable/useUsersTable.tsx
```diff
@@ -49,15 +49,20 @@ export function useUsersTable(filters: IUserFilters) {
...
+          .filter(isDefined)
+          .sort(compareUsers);
+        const paginatedUsers = usersResource.get(pagination.allPages).filter(isDefined).sort(compareUsers);
+        const allUsers = Array.from([...filteredUsers, ...paginatedUsers].values());
+        const allUsersUniqueMap = new Map<string, AdminUser | AdminUserNew>(allUsers.map(user => [user.userId, user]));
+        const allUsersUnique: (AdminUser | AdminUserNew)[] = Array.from(allUsersUniqueMap.values());
+        const newUsers: AdminUserNew[] = allUsersUnique.filter(isNewUser).sort((a, b) => b.createdAt.localeCompare(a.createdAt));
+        const existingUsers: AdminUser[] = allUsersUnique.filter(user => !isNewUser(user));
+
+        return filters.filterUsers([...newUsers, ...existingUsers]);
```
Wroud: `filter(isDefined)` x2
`sort(compareUsers)` x2
`filter(isNewUser)` x2

You probably can do it simpler by joining filter/sort operations together, so basically:
1. Get all users
2. filter (isDefined)
3. sort (isNewUser + compareUsers)

Wroud (changes requested):

webapp/packages/plugin-authentication-administration/src/Administration/Users/UsersTable/useUsersTable.tsx
sergeyteleshev: okay! simplified a bit

webapp/packages/core-authentication/src/UsersResource.ts
```diff
@@ -148,14 +148,20 @@ export class UsersResource extends CachedMapResource<string, AdminUser, UserReso
 
     const newUser = user as unknown as AdminUserNew;
     newUser[NEW_USER_SYMBOL] = true;
+    newUser.createdAt = new Date().toISOString();
```
Wroud: maybe store timestamp (number) so it can be easily compared?

webapp/packages/plugin-authentication-administration/src/Administration/Users/UsersTable/useUsersTable.tsx
```diff
@@ -49,15 +56,15 @@ export function useUsersTable(filters: IUserFilters) {
...
-          new Set([
-            ...this.usersLoader.resource.get(UsersResourceFilterKey(searchFilter, enabledStateFilter)),
-            ...usersResource.get(pagination.allPages).filter(isDefined).sort(compareUsers),
-          ]),
-        );
-        return filters.filterUsers(users.filter(isDefined));
+        const filteredUsers = this.usersLoader.resource.get(UsersResourceFilterKey(searchFilter, enabledStateFilter));
+        const paginatedUsers = usersResource.get(pagination.allPages);
+        const allUsers = Array.from([...filteredUsers, ...paginatedUsers].values()).filter(isDefined);
+        const allUsersUniqueMap = new Map<string, AdminUser | AdminUserNew>(allUsers.map(user => [user.userId, user]));
```
Wroud: Why not use the original approach with Set? In the current implementation, there can't be the same user stored with a different object

Wroud: because right now its: 
1. new array [...filteredUsers, ...paginatedUsers]
2. iterator `.value()` 
3. new array from iterator `Array.from`
4. filter
5. map
6. iterator
7. new array from iterator
8. sort
9. sort

```ts
const filteredUsers = this.usersLoader.resource.get(UsersResourceFilterKey(searchFilter, enabledStateFilter));
const paginatedUsers = usersResource.get(pagination.allPages);
const allUsersUnique = Array.from(
          new Set([
            ...filteredUsers,
            ...paginatedUsers,
          ].filter(isDefined)),
        ).sort(sortUsersById).sort(sortByNewUsers);
```

this will make only this operations:
1. new array `[...filteredUsers, ...paginatedUsers]`
2. filter `isDefined`
3. new set
4. new array from set
5. sort
6. sort

Wroud (changes requested):

webapp/packages/plugin-authentication-administration/src/Administration/Users/UsersTable/useUsersTable.tsx
sergeyteleshev: I did that to exclude situations when the user is the same, but the object references are different (so we have two equal users)

Wroud (dismissed):

webapp/packages/core-authentication/src/UsersResource.ts
```diff
@@ -319,10 +325,29 @@ export function isLocalUser(user: AdminUser): boolean {
...
 }
 
-export function isNewUser(user: AdminUser): boolean {
-  return NEW_USER_SYMBOL in user && user[NEW_USER_SYMBOL] === true;
+export function isNewUser(user: AdminUser | AdminUserNew): user is AdminUserNew {
+  return NEW_USER_SYMBOL in user && user[NEW_USER_SYMBOL] === true && 'createdAt' in user && Boolean(user.createdAt);
 }
 
-export function compareUsers<T extends Pick<AdminUser, 'userId'>>(a: T, b: T): number {
+export function sortUsersById<T extends Pick<AdminUser, 'userId'>>(a: T, b: T): number {
```
devnaumov: I think this naming dont suit here, if we add another property, userName for example, we need to change the fn name again?

devnaumov (changes requested):

webapp/packages/core-authentication/src/UsersResource.ts
devnaumov: This fn is very generic, we have this in many resources (compareConnections, compareProviders and etc). I think we should not give it specific name

webapp/packages/core-authentication/src/UsersResource.ts
```diff
@@ -148,14 +148,20 @@ export class UsersResource extends CachedMapResource<string, AdminUser, UserReso
...
   }
 
   cleanNewFlags(): void {
     for (const user of this.data.values()) {
-      (user as AdminUserNew)[NEW_USER_SYMBOL] = false;
+      if (!isNewUser(user)) {
+        continue;
+      }
+
+      user[NEW_USER_SYMBOL] = false;
```
devnaumov, line 163: `delete user.createdAt`?

webapp/packages/core-authentication/src/UsersResource.ts
```diff
@@ -319,10 +325,29 @@ export function isLocalUser(user: AdminUser): boolean {
...
+
+  if (!aIsNew && bIsNew) {
+    return 1;
+  }
+
+  if (aIsNew && bIsNew) {
+    return b.createdAt - a.createdAt;
+  }
+
+  return 0;
```
devnaumov, line 352: `return compareUsers(a,b)`? And then you dont need to sort twice in users table

webapp/packages/core-authentication/src/UsersResource.ts
sergeyteleshev: All cool here. The function actually sorts users by ID and is renamed to reflect its purpose

`compareUsers` is too generic and tells almost nothing. This is not the case where we should keep the concept with other abstractions due to its being too tiny and having practically no logic

webapp/packages/core-authentication/src/UsersResource.ts
sergeyteleshev: This happens cause it has strict types:
`type AdminUserNew = AdminUser & { [NEW_USER_SYMBOL]: boolean; createdAt: number };`

And we cannot make it optional cause new users must have these properties always. So all we can do here is reset the values. `cleanNewFlags` is called only when we cancel user creation via the "Cancel" button. So it's fine here and cannot break something

Screenshot 2025-11-17 at 18 26 21

webapp/packages/core-authentication/src/UsersResource.ts
sergeyteleshev: The sort function helper is called `sortByNewUsers`. Therefore, it's supposed to adhere to the principle of "single responsibility". If neither user is new, we should do nothing in the context of sorting in order to keep predictable behavior. If we want to, we can always combine these two sorts: `sortByNewUsers` and `sortUsersById`, which we actually do

webapp/packages/core-authentication/src/UsersResource.ts
devnaumov: CompareUsers name tells us what we need to know, it compare users, in any place in the app where you need to sort users, you dont need to think whether you want to sort it by id or by name, you just apply generic fn from resource that sort the users the way we want. 

What's more important, you changed the way how we name things in this resource, but did not do it in others resources So it just creates a lack of consistency and consistency is very important

webapp/packages/core-authentication/src/UsersResource.ts
devnaumov: As the concept of a new user is purely ui related and doesn’t exist in the business domain, we should consider removing this logic from the resource entirely. Instead, we could store new user ids in a service near the users table and reset it when the user closes the tab and etc. We can handle this later I guess

```
onUserCreate = (userId) => this.newUsers.push(userId);
if(newUsers.includes(user.userId)) {sort of something else}
```

devnaumov (dismissed):

webapp/packages/core-authentication/src/UsersResource.ts
SychevAndrey: I agree with @devnaumov here, current naming brakes several principles like encapsulation and information hiding. From the interface we let the user of that function know that id is required and we use it somehow, but nothing stops us from adding an additional(maybe optional) field like name there and then change the implementation adding another layer of sorting.

In this case, if we do that, we need to rename the function again.

SychevAndrey: Moreover, **sort** is a wrong verb here because this function is not sorting anything, it is a comparator fn that is being using in sort(), so it's better to keep **compare** verb for it.

webapp/packages/core-authentication/src/UsersResource.ts
SychevAndrey: I believe responsibility here is to sort users. How we do that does not really matter. And we use both functions anyway only for sorting users, so why not to combine them in a single compareUsers function?

webapp/packages/core-authentication/src/UsersResource.ts
Wroud: This is an option (store new users' information separately), but it might be too complex compared to the profit (you need to sync this state correctly with the resource). Therefore, I will recommend keeping the current approach at least for now

SychevAndrey (changes requested):
compareUsers

Wroud (dismissed):

devnaumov (dismissed):

SychevAndrey (approved):

devnaumov (approved):

PR #3911, opened 2025-11-17
dbeaver/pro#7282 show alert for admin team in connection tab

devnaumov:
closes [7282](https://github.com/dbeaver/pro/issues/7282)


webapp/packages/plugin-authentication-administration/src/Administration/Users/Teams/TeamsForm/GrantedConnections/GrantedConnectionsTable.tsx
```diff
@@ -56,6 +59,12 @@ export const GrantedConnectionsTable: TabContainerPanelComponent<TeamFormProps>
     return null;
   }
 
+  const fullAccess = teamLoader.data?.teamPermissions.includes(EAdminPermission.admin);
+
+  if (fullAccess) {
+    return <Alert className="tw:h-max">{translate('connections_connection_access_admin_info')}</Alert>;
```
sergeyteleshev, line 65: Perhaps we can revise the translation from `Administrators see all connections except private connections of other users` to `Administrators have access to all connections except private connections of other users`?

And move this if clause to line 123 so returns packed together

sergeyteleshev (dismissed):

webapp/packages/plugin-authentication-administration/src/Administration/Users/Teams/TeamsForm/GrantedConnections/GrantedConnectionsTable.tsx
```diff
@@ -46,6 +48,7 @@ export const GrantedConnectionsTable: TabContainerPanelComponent<TeamFormProps>
 
   const connectionLoader = useResource(GrantedConnectionsTable, ConnectionInfoCustomOptionsResource, key, { active: selected });
   const connectionsOriginLoader = useResource(GrantedConnectionsTable, ConnectionInfoOriginResource, key, { active: selected });
+  const teamLoader = useResource(GrantedConnectionsTable, TeamsResource, formState.state.teamId, { active: selected });
```
Wroud: you should use `TeamOptionsFormPart` instead for accessing `teamPermissions` so state will be in sync

Wroud (changes requested):

webapp/packages/plugin-authentication-administration/src/Administration/Users/Teams/TeamsForm/GrantedConnections/GrantedConnectionsTable.tsx
devnaumov: I think we dont need to compute all variables and functions down the file if we know that we return just the alert so its better to return early. About translation, I think your variant is definetly better but its better to sync such thing with team as we use this message in many places

sergeyteleshev (approved):

Wroud (approved):

PR #3913, opened 2025-11-18
[CB] Fix high CVE-2025-64756 on glob

sergeyteleshev:
closes https://github.com/dbeaver/pro/issues/7520


SychevAndrey (dismissed):

webapp/packages/core-cli/package.json
```diff
@@ -39,17 +39,19 @@
...
+    "glob": ">=11.1.0 <12.0.0"
+  },
   "devDependencies": {
     "@cloudbeaver/tsconfig": "workspace:*",
     "@types/js-yaml": "^4",
     "@types/node": "^22",
     "enquirer": "^2",
     "execa": "^9",
-    "glob": "^11",
+    "glob": ">=11.1.0 <12.0.0",
```
Wroud: why `<12.0.0`?

sergeyteleshev: usually we update major version manually to avoid cases when the API changed and everything is broken after that

https://docs.npmjs.com/cli/v11/configuring-npm/package-json#dependencies - from these examples I made a conclusion that version with ">=" does not limit the major version from right border

Wroud: But you can also specify `^11.1` to make it upgrade to the version where vulnerability fixes

sergeyteleshev: agreed. done

webapp/yarn.lock
```diff
@@ -12344,7 +12344,7 @@ __metadata:
   languageName: node
   linkType: hard
 
-"glob@npm:^11, glob@npm:^11.0.0":
+"glob@npm:^11.0.0":
   version: 11.0.3
```
Wroud: now it has multiple versions of the `glob` you can try `yarn dedupe`

Wroud (changes requested):

webapp/yarn.lock
Wroud: and you probably don't need to use `resolutions`

sergeyteleshev: It has multiple versions because other packages use it as well. That's why I used `resolutions`. However, it didn't really help, even with `yarn dedupe'. We covered it only from our side. All we have left to do is wait until the other packages update it as well.

In the tech debt sprint, we will update dependencies for everything, and hope that  by this time, other packages will also handle this issue.

webapp/yarn.lock
Wroud: so this PR will not fix CVE because we still has this glob version

Wroud (changes requested):

webapp/yarn.lock
Wroud: you need to add `"glob": "^11.1",` resolution to the root package.json this will force to use `"glob": "^11.1"` version

webapp/packages/core-cli/package.json
```diff
@@ -49,7 +49,6 @@
     "@types/node": "^22",
     "enquirer": "^2",
     "execa": "^9",
-    "glob": "^11",
```
Wroud, line 52: please keep both

Wroud (changes requested):

Wroud (approved):

SychevAndrey (approved):

PR #3914, opened 2025-11-18
dbeaver/pro#5083 feat: script status message

Wroud:
closes dbeaver/pro#5083

### DIFF
--- a/webapp/packages/plugin-sql-editor-navigation-tab-script/src/ResourceSqlDataSource.ts
+++ b/webapp/packages/plugin-sql-editor-navigation-tab-script/src/ResourceSqlDataSource.ts
@@ -38,7 +38,6 @@ interface IResourceActions {
 }
 
 const VALUE_SYNC_DELAY = 1 * 1000;
-const MESSAGE_DISPLAY_DELAY = 4 * 1000;
 
 export class ResourceSqlDataSource extends BaseSqlDataSource {
   static override key = 'resource';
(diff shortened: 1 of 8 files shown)


sergeyteleshev (dismissed):

webapp/packages/plugin-sql-editor-navigation-tab-script/src/ResourceSqlDataSource.ts
```diff
@@ -85,6 +84,37 @@ export class ResourceSqlDataSource extends BaseSqlDataSource {
...
+    }
+
+    const projectId = this.executionContext?.projectId;
+    const resourceProjectId = getRmResourceKey(this.resourceKey).projectId;
+    const userProjectId = this.projectsService.userProject?.id;
+
+    return isNotNullDefined(projectId) && resourceProjectId !== projectId && resourceProjectId !== userProjectId;
+  }
+
+  override get message(): string | undefined {
```
devnaumov: TLocalizationToken instead of string?

webapp/packages/plugin-sql-editor/src/SqlEditorStatusBar.tsx
```diff
@@ -18,5 +18,20 @@ interface Props {
...
+    <Loader
+      className={s(styles, { statusPosition: true, loader: true })}
+      state={dataSource}
+      message={dataSource?.loadingMessage}
+      inline
+      inlineException
+    >
+      {dataSource?.message && (
+        <div className={s(styles, { statusPosition: true }, 'tw:p-2')}>
+          <Alert title={t(dataSource?.message)} />
```
devnaumov: no need for ?. check here

webapp/packages/plugin-sql-editor/src/SqlDataSource/BaseSqlDataSource.ts
```diff
@@ -25,7 +25,10 @@ export abstract class BaseSqlDataSource<TDataSource extends QueryDataSource = Qu
   static key = 'base';
 
   abstract get name(): string | null;
-  message?: string;
+  get message(): string | undefined {
```
devnaumov: Looks strange that we dont have any logic inside the getter, should it be abstract?

Wroud: In case it will be abstract, every source that extends `BaseSqlDataSource` will be forced to implement it, but this is optional behaviour, so that the base implementation can return `undefined`

webapp/packages/plugin-sql-editor/src/SqlEditorStatusBar.tsx
Wroud: thanks, fixed

sergeyteleshev (approved):

devnaumov (approved):

PR #3926, opened 2025-11-20
[CB] Improve description for ordering icon

sergeyteleshev:
closes https://github.com/dbeaver/pro/issues/7554


SychevAndrey (commented):
Maybe we don't need to pass the title as a prop here and could just use
`import { useTranslate } from '@dbeaver/react-translate';`
inside OrderButton

Please, check.

sergeyteleshev:
> Maybe we don't need to pass the title as a prop here and could just use `import { useTranslate } from '@dbeaver/react-translate';` inside OrderButton
> 
> Please, check.

This feature is present in 2 places. In one place, we want it. But in another we don't cause we don't have this functionality there

Wroud:
I also don't like the idea of adding props for translation. Let's move this ticket to the next tech sprint and discuss a possible solution

PR #3928, opened 2025-11-20
[CB] write unit tests

sergeyteleshev:
closes https://github.com/dbeaver/pro/issues/7563


webapp/packages/core-blocks/src/IconOrImage.test.tsx
```diff
@@ -0,0 +1,125 @@
...
+vi.mock('./StaticImage.js', () => ({
+  StaticImage: vi.fn(({ ...props }) => <img {...props} data-testid="mock-static-image" />),
+}));
+
+const STATIC_IMAGE_TEST_ID = 'mock-static-image';
+const ICON_TEST_ID = 'mock-icon';
+
+describe('IconOrImage', () => {
+  test('should render StaticImage for platform: prefixed icon', () => {
+    const { queryByTestId } = renderInApp(<IconOrImage icon="platform:/image.png" />);
```
SychevAndrey: Should we use getByRole instead https://testing-library.com/docs/queries/byrole ?

webapp/packages/core-blocks/src/StaticImage.test.tsx
```diff
@@ -0,0 +1,123 @@
...
+
+describe('StaticImage', () => {
+  test('should return null when icon is not provided', () => {
+    const { container } = renderInApp(<StaticImage />);
+    expect(container.firstChild).toBeNull();
+  });
+
+  test('should render img element when icon is provided', () => {
+    const { container } = renderInApp(<StaticImage icon="/image.png" />);
+    const img = container.querySelector('img');
```
SychevAndrey: it's also can be replaced by getByRole

webapp/packages/core-blocks/src/UploadArea.test.tsx
```diff
@@ -0,0 +1,172 @@
...
+    expect(label.tagName).toBe('LABEL');
+  });
+
+  it('should render hidden file input', () => {
+    const { container } = renderInApp(<UploadArea>Upload File</UploadArea>);
+    const input = container.querySelector('input[type="file"]');
+
+    expect(input).toBeInTheDocument();
+    expect(input).toHaveAttribute('hidden');
+  });
```
SychevAndrey: Do we need tests like that? I think we don't care much about having an input there. What is more important that our component works properly. Maybe we could add something like
const dataTransfer = new DataTransfer();
and then use fireEvent to drop this data into our input and check the result.

webapp/packages/core-blocks/src/useFuzzySearch.test.ts
```diff
@@ -0,0 +1,223 @@
...
+        sourceProposals: testData,
+        fields: ['name'],
+        threshold: 0.1,
+      }),
+    );
+
+    result.current.search('test');
+
+    expect(mockSearch).toHaveBeenCalledWith('test');
+  });
```
SychevAndrey: we don't test here that the threshold param has any effect.

SychevAndrey (changes requested):
💪🏻 a nice bunch of tests!

Let's try to use getByRole as much as possible according to [priority guide](https://testing-library.com/docs/queries/about#priority)

webapp/packages/core-blocks/src/UploadArea.test.tsx
SychevAndrey: BTW, there is [user event lib](https://testing-library.com/docs/user-event/intro/) that we can utilize for cases like that

webapp/packages/core-blocks/src/IconOrImage.test.tsx
sergeyteleshev: The role can be provided inside layout so we have a case when we have for example multiple roles and collision of what to pick via testing library

I think testid is more strict and determined way to get a element you want

webapp/packages/core-blocks/src/UploadArea.test.tsx
sergeyteleshev: okay! I fixed the case to check if we have input `type="file"`

webapp/packages/core-blocks/src/useFuzzySearch.test.ts
sergeyteleshev: removed it

webapp/packages/core-blocks/src/IconOrImage.test.tsx
SychevAndrey: But the role of the component should be determined at any point of time depending on the content. So, we should test it as well.

webapp/packages/core-blocks/src/UploadArea.test.tsx
SychevAndrey: I don't understand it anyway. What we test here? Okay, I have an input. Can I use it? How do we check that?
The previous implementation with hidden was even better, there we at least tested that input is not visible.

webapp/packages/core-blocks/src/UploadArea.test.tsx
```diff
@@ -0,0 +1,172 @@
...
+    const { fileList } = createTestFileList();
+    Object.defineProperty(input, 'files', {
+      value: fileList,
+      writable: true,
+    });
+    fireEvent.change(input);
+
+    expect(handleChange).toHaveBeenCalled();
+    expect(input.value).toBe('');
+  });
```
SychevAndrey: What we really do here is just somehow by magic call the change method on input. And we don't test that user can actually interact with the input. Maybe it's blocked on UI, it could be in disabled state, etc.
So either we need to add those checks manually, or, which I believe is better, just install user events lib, which checks all those things under the hood.

SychevAndrey (changes requested):

webapp/packages/core-blocks/src/UploadArea.test.tsx
sergeyteleshev: You can only upload files using the `input type=file` element. It tests that input actually is input type file and not anything else

SychevAndrey (dismissed):

devnaumov (approved):

devnaumov (dismissed):

Wroud (approved):

PR #3942, opened 2025-11-27
dbeaver/pro#4503 migrate admin tables to data grid

devnaumov:
closes [4503](https://github.com/dbeaver/pro/issues/4503)


webapp/packages/plugin-authentication-administration/src/Administration/Users/Teams/TeamsTable/TeamsPage.tsx
```diff
@@ -23,11 +28,16 @@ export const TeamsPage = observer<Props>(function TeamsPage({ param }) {
   const translate = useTranslate();
   const service = useService(CreateTeamService);
 
-  const table = useTeamsTable();
+  const teamsLoader = useResource(TeamsPage, TeamsResource, CachedMapAllKey);
```
sergeyteleshev, line 31: do we have clients who has huge amount of teams? will they have an interface freeze if a lot of the teams loaded at 1 request?
may be we should add here pagination and load teams by page?

sergeyteleshev: I can see for users we already do that. I suppose not bad idea to paginate teams also. May be a separate ticket if you think it is out of the scope for current PR

webapp/packages/plugin-authentication-administration/src/Administration/Users/UsersTable/UsersPage.tsx
```diff
@@ -35,9 +37,15 @@ export const UsersPage = observer<Props>(function UsersPage({ param }) {
...
   const create = param === 'create';
   const displayAuthRole = authRolesResource.data.length > 0;
-  const loading = authRolesResource.isLoading() || table.loadableState.isLoading();
+  const loading = table.loadableState.isLoading();
   const isManageable = !administrationUsersManagementService.externalUserProviderEnabled;
 
+  function loadMore() {
+    if (table.hasMore) {
+      table.loadMore();
+    }
```
sergeyteleshev: no need for this check it is already checked here:
https://github.com/dbeaver/cloudbeaver/blob/69abdc039d2b58ce825b83d19218540fb5d28299/webapp/packages/core-blocks/src/ResourcesHooks/useOffsetPagination.ts#L88-L90

webapp/packages/plugin-authentication-administration/src/Administration/Users/UsersTable/useUsersTable.tsx
```diff
@@ -49,6 +42,8 @@ export function useUsersTable(filters: IUserFilters) {
         return pagination.hasNextPage;
       },
       get users() {
+        usersLoader.data; // triggers suspense, behaves differently than forceSuspense action cause used on the getter level
```
sergeyteleshev, line 45: why do we need suspense here if we show loader manually here? we should stick to one approach here and probably to suspense

https://github.com/dbeaver/cloudbeaver/blob/69abdc039d2b58ce825b83d19218540fb5d28299/webapp/packages/plugin-authentication-administration/src/Administration/Users/UsersTable/UsersPage.tsx#L68

sergeyteleshev: I suggest to replace `...usersResource.get(pagination.allPages)` with `...usersLoader.data` and it should be fine

webapp/packages/plugin-authentication-administration/src/Administration/Users/Teams/TeamsTable/TeamsPage.tsx
```diff
@@ -52,23 +62,23 @@ export const TeamsPage = observer<Props>(function TeamsPage({ param }) {
...
+      <Container overflow gap maximum>
         {create && (
           <Group box>
             <CreateTeam />
           </Group>
         )}
-        <Group boxNoOverflow>
-          <TeamsTable teams={table.teams} state={table.state} selectedItems={table.tableState.selected} expandedItems={table.tableState.expanded} />
-        </Group>
+        <TableSelectionContext value={selection}>
```
sergeyteleshev, line 79: Nice request! I also suggest syncing with the team on the topic of UI, which has changed from a stretched table to a compact one. I like that we now have more space for something else, but from an aesthetic point of view, I feel something a bit odd compared to the previous UI

Image

Image

sergeyteleshev: Probably we should discuss it at UI/UX meeting and then with the next iteration fix issues if we find some

webapp/packages/plugin-authentication-administration/src/Administration/Users/UsersTable/UsersTable.tsx
```diff
@@ -5,88 +5,156 @@
...
+  const usersAdministrationService = useService(UsersAdministrationService);
+  const usersResource = useService(UsersResource);
+
+  if (displayAuthRole) {
+    const isRole = COLUMNS.find(c => c.key === ROLE_COLUMN.key);
+
+    if (!isRole) {
+      COLUMNS.push(ROLE_COLUMN);
+    }
+  }
```
Wroud:
```suggestion
  let columns = COLUMNS;

  if (displayAuthRole) {
    columns = [...COLUMNS, ROLE_COLUMN];
  }
```

Wroud: please don't mutate constants

Wroud (changes requested):

webapp/packages/plugin-authentication-administration/src/Administration/Users/UsersTable/UsersTable.tsx
devnaumov: Good point, better to create separate variable inside component and memo it

webapp/packages/plugin-authentication-administration/src/Administration/Users/UsersTable/useUsersTable.tsx
devnaumov: We cant, we have a workaround inside useResource for such cases named forceSuspense, but we use it manually inside getter because we want to trigger it only when we access users

CleanShot 2025-12-01 at 16 38 42@2x

devnaumov: Spinner is used here for a reason (similar to qm) as it triggers on "fetch more" action, its should not block all page as suspense does

webapp/packages/plugin-authentication-administration/src/Administration/Users/Teams/TeamsTable/TeamsPage.tsx
devnaumov: Pagination adds a lot of overhead on the resource level, also requires support from the backend side. I think we can create a ticket to support optional pagination on all get requests

webapp/packages/plugin-authentication-administration/src/Administration/Users/UsersTable/UsersTable.tsx
```diff
@@ -5,88 +5,162 @@
...
+    if (displayAuthRole) {
+      const isRole = result.find(c => c.key === ROLE_COLUMN.key);
+
+      if (!isRole) {
+        result.push(ROLE_COLUMN);
+      }
+    }
+
+    return result;
+  }, [displayAuthRole]);
```
SychevAndrey, line 57: why so complicated:

```suggestion
const columns = displayAuthRole ? [...COLUMNS, ROLE_COLUMN] : COLUMNS
```

SychevAndrey (dismissed):

webapp/packages/plugin-authentication-administration/src/Administration/Users/UsersTable/UsersTable.tsx
devnaumov: Fixed, now we dont need to check

sergeyteleshev (dismissed):

Wroud (dismissed):

Wroud (approved):

sergeyteleshev (approved):

PR #3951, opened 2025-12-02
dbeaver/pro#7527 persist right panel tabs

devnaumov:
closes [7527](https://github.com/dbeaver/pro/issues/7527)

https://github.com/user-attachments/assets/93e45914-b101-4d18-84a4-74a7daa41439


webapp/packages/core-ui/src/Screens/AppScreen/BaseOptionsPanelService.ts
```diff
@@ -40,24 +40,30 @@ export abstract class BaseOptionsPanelService<T> {
   }
 
   async open(itemId: T): Promise<boolean> {
-    if (this.optionsPanelService.isOpen(this.panelGetter)) {
+    if (this.isOpen()) {
       return true;
     }
 
+    this.itemId = itemId;
```
Wroud: I think this is incorrect, because panel opening can be canceled; in this case, the itemId will be reset anyway

webapp/packages/core-ui/src/Tabs/useTabPersistence.ts
```diff
@@ -0,0 +1,71 @@
...
+          if (!prevTabs.current.includes(id)) {
+            state.selectedTabId = id;
+            break;
+          }
+        }
+
+        prevTabs.current = tabs;
+      }
+
+      if (state.selectedTabId) {
```
Wroud: please use strict check like `isDefined` or `isNotNullDefined`

webapp/packages/core-ui/src/Tabs/useTabPersistence.ts
```diff
@@ -0,0 +1,71 @@
...
+          } else {
+            state.selectedTabId = undefined;
+          }
+        }
+      }
+    }),
+  );
+
+  const result = useMemo(() => ({ selectedTabId: state.selectedTabId, selectTab }), [state.selectedTabId, selectTab]);
+  return result;
```
Wroud: just `return useMemo`?

Wroud: and probably there is no need for memo, or you can use `useObjectRef`

webapp/packages/core-ui/src/Tabs/useTabPersistence.ts
```diff
@@ -0,0 +1,71 @@
...
+    executor: container.onTabSelect,
+    handlers: [
+      function selectTabHandler(tabId) {
+        selectTab(tabId);
+      },
+    ],
+  });
+
+  untracked(
+    action(() => {
```
Wroud: in mobx `action` is untracked by default, so you can use `runInAction` without `untracked`

webapp/packages/plugin-user-profile/src/UserInfo.tsx
```diff
@@ -28,7 +28,7 @@ export const UserInfo = observer<Props>(function UserInfo({ info }) {
       as="div"
       className={s(style, { user: true })}
       title={translate('plugin_user_profile_menu')}
-      onClick={() => userProfileOptionsPanelService.open()}
+      onClick={() => userProfileOptionsPanelService.open(null)}
```
Wroud: this looks strange

webapp/packages/plugin-tools-panel/src/ToolsPanel/ToolsPanel.tsx
```diff
@@ -5,69 +5,33 @@
...
-import { isArraysEqual } from '@cloudbeaver/core-utils';
+import { TabList, TabPanelList, TabsState, useTabOrderPersistence, useTabPersistence } from '@cloudbeaver/core-ui';
 
 import styles from './ToolsPanel.module.css';
 import { ToolsPanelService } from './ToolsPanelService.js';
 
-interface IToolsState {
-  selectedTabId: string | undefined;
-}
+const PANEL_ID = 'dbeaver-tools-panel';
```
Wroud: let's rename it without `dbeaver-` prefix

Wroud (changes requested):

webapp/packages/core-ui/src/Screens/AppScreen/BaseOptionsPanelService.ts
devnaumov: If you try to open the panel without setting itemId, it starts to open with incorrect value, in some places its fine but in general we want to have valid data right after the panel start opening. You start to render form with a=null, then in the middle of the process you set a=3, it seems as wrong approach

webapp/packages/core-ui/src/Tabs/useTabPersistence.ts
devnaumov: Personally i prefer not to use memoization at least there is a really good reason for it. However, here I did the same as in order persistence hook cause it was reviewed before. But i can remove it

webapp/packages/core-ui/src/Tabs/useTabPersistence.ts
devnaumov: I took the existing logic and reuse it. I will try to adjust it and test

webapp/packages/core-ui/src/Screens/AppScreen/BaseOptionsPanelService.ts
devnaumov: Moved it to a setter callback right before setting the panel

Wroud (approved):

webapp/packages/core-ui/src/Tabs/useTabPersistence.ts
```diff
@@ -0,0 +1,72 @@
...
+
+  const tabs = container.getIdList();
+  const prevTabs = useRef<string[]>(tabs);
+  const equal = isArraysEqual(prevTabs.current, tabs);
+
+  const selectTab = useCallback(
+    (tabId: string) => {
+      state.selectedTabId = tabId;
+    },
+    [state],
```
sergeyteleshev, line 33: `state` is a new object each render of this hook according to `useUserData` implementation. so this `useCallback` is not working properly

simply can use `state.selectedTabId` in the deps to make it work right

SychevAndrey (approved):

webapp/packages/core-ui/src/Tabs/useTabPersistence.ts
```diff
@@ -0,0 +1,72 @@
...
+interface ITabPersistenceState {
+  selectedTabId: string | undefined;
+}
+
+export function useTabPersistence(panelId: string, container: TabsContainer) {
+  const state = useUserData<ITabPersistenceState>(panelId, () => ({ selectedTabId: undefined }));
+
+  const tabs = container.getIdList();
+  const prevTabs = useRef<string[]>(tabs);
+  const equal = isArraysEqual(prevTabs.current, tabs);
```
sergeyteleshev, line 27: I think we can remove our implementation of `isArraysEqual`

cause we already have this:
https://www.npmjs.com/package/fast-deep-equal

it works with typed arrays, sets and maps. moreover it has functionality to work with react objects:
https://www.npmjs.com/package/fast-deep-equal#usage

webapp/packages/core-ui/src/Tabs/useTabPersistence.ts
```diff
@@ -0,0 +1,72 @@
...
+      function selectTabHandler(tabId) {
+        selectTab(tabId);
+      },
+    ],
+  });
+
+  runInAction(() => {
+    if (!equal) {
+      for (const id of tabs) {
+        if (!prevTabs.current.includes(id)) {
```
sergeyteleshev, line 48: for in for cycle (N^2 time complexity)

can be easily resolved if we use `Sets` instead of string `array` for `tabs & prevTabs`

webapp/packages/core-ui/src/Tabs/useTabPersistence.ts
```diff
@@ -0,0 +1,72 @@
...
+          break;
+        }
+      }
+
+      prevTabs.current = tabs;
+    }
+
+    if (isDefined(state.selectedTabId)) {
+      if (!tabs.includes(state.selectedTabId)) {
+        if (tabs.length > 0) {
```
sergeyteleshev, line 59: same will be resolved here with `Set`

webapp/packages/core-ui/src/Tabs/useTabPersistence.ts
```diff
@@ -0,0 +1,72 @@
...
+          state.selectedTabId = tabs[0];
+        } else {
+          state.selectedTabId = undefined;
+        }
```
sergeyteleshev, line 63: state.selectedTabId = tabs[0]; // returns first tab or undefined. no need for `if else` clause

webapp/packages/core-ui/src/Tabs/useTabPersistence.ts
Wroud: I don't think that we need deep equality here; if objects' links are different, we don't need to try to compare them to ensure they're actually equal deeply

PR #3952, opened 2025-12-03
dbeaver/pro#5232 [CB] Trap Focus In the Right Panel

SychevAndrey:
https://github.com/user-attachments/assets/c377491e-186d-4ed3-89de-5e560e3abe06

https://github.com/user-attachments/assets/7dfe3066-62e2-440f-9d7e-6b67d0089513

While working with this task I encountered several issues:

I couldn't use FocusTrapRegion on the Panel itself, because components should be open to extension to use it, which we could not guarantee with our design.
I couldn't apply FocusTrapRegion on the SlideElement, because it is used for the static part of the app as well.
Then I tried to use FocusTrapRegion on the SlideBox. It entailed two problems:

1. Focus will be trapped only after user interaction. We somehow steal focus from the other part of the app.
2. We have not only the Right Panel in the SlideBox, so we had to make other element inert manually which is weird in case of using library.

Having said that, I decided to use Dialog component from the AriaKit library. This component is based on the [WAI-ARIA Dialog Pattern](https://www.w3.org/WAI/ARIA/apg/patterns/dialogmodal/) and do everything what we need.

To achieve that I did the following:

1. Added new Dialog component to ui-kit
2. Used it to implement Right Panel.

Here I came across to another problem: nested dialogs were not showed properly. That was kinda expected, so I just make another bunch of changes:

3. Added some Dialog layout wrapper
4. Used new ui-kit Dialog component to implement CB's CommonDialog.

As a result, CommonDialogWrapper became much more simple. We could delete state and focus handling. We also simplified DialogsPortal component, because AriaKit Dialog can handle all the inner dialog focus movements for us.

I generated a bunch of Stories for storybook. I also added storybook source add-on. It comes in handy when you want to see examples of implementation. No need to search the workspace, you can see the code and how it works right in the storybook.

P.S.
For Right Panel I tried to implement special variant of Dialog, but due to growing complexity of that task and many changes, I stopped myself from changing RightArea logic with SlideBox and SlideElement. If we decide, we want to simplify it and remove carousel, we will be able to use Slide Dialog later.


webapp/packages/core-app/src/AppScreen/RightArea.tsx
```diff
@@ -60,12 +62,26 @@ export const RightArea = observer<Props>(function RightArea({ className }) {
...
-        </Loader>
+        <Dialog
+          portal={false}
+          unmountOnHide={false}
+          animated={false}
+          data-size="free"
+          open={optionsPanelService.active}
+          className="tw:w-full tw:h-full tw:overflow-visible! tw:bg-transparent!"
+          onClose={close}
+        >
```
Wroud: Seems like `Dialog` should be part of the `SlideElement`, or we can create a separate `SlideElement` that will act as a dialog

You also probably want to use Ariakit `render` (composition) ability and some property (in `Dialog`) to be able to use `Dialog` from the `ui-kit` without styles

Wroud (changes requested):

webapp/common-react/@dbeaver/ui-kit/src/Dialog/Dialog.tsx
```diff
@@ -0,0 +1,92 @@
+/*
+ * CloudBeaver - Cloud Database Manager
+ * Copyright (C) 2020-2025 DBeaver Corp and others
+ *
+ * Licensed under the Apache License, Version 2.0.
+ * you may not use this file except in compliance with the License.
+ */
+
+import {
```
devnaumov, line 9: Seems like new autofocus system breaks our behaviour when we select specific tab when opening user profile dialog. It shows on your second video

webapp/packages/core-blocks/src/CommonDialog/CommonDialog/CommonDialogWrapper.tsx
```diff
@@ -24,50 +23,35 @@ export interface CommonDialogWrapperProps {
...
         aria-label={ariaLabel}
-        className={s(computedStyles, { container: true })}
-        visible={context.visible}
-        hideOnClickOutside={false}
-        modal={false}
+        open={context.visible}
+        data-size={size}
+        className={s(computedStyles, { dialog: true, fixedSize, fixedWidth, freeHeight }, className)}
+        backdrop
+        unmountOnHide
```
Wroud: `unmountOnHide` might break local state in complex dialogs with subdialogs, for example, an authentication dialog. If you encounter an error and click to show details, when you return, the local state of the authentication dialog will be lost.

Wroud (changes requested):

SychevAndrey:
https://github.com/user-attachments/assets/f991f197-ce94-4aa3-8ecd-53cd322e2f4e

Dialog was removed from SlidePanel implementation, we returned back to manual ESC hotkey handling.
The differences now are the following:

1. Hidden panel becomes inert, so focus never goes there and always visible on the screen
2. Closing button are reachable from the keyboard now and get the focus when the panel is open, since it's always there and don't need to be loaded, it gives us very reliable mechanism of autofocusing on the panel.

Wroud (dismissed):

sergeyteleshev (dismissed):

webapp/packages/core-blocks/src/CommonDialog/DialogsPortal.tsx
```diff
@@ -53,11 +52,13 @@ export const DialogsPortal = observer(function DialogsPortal() {
   }
 
   return (
-    <Loader className={s(styles, { loader: true })} suspense>
-      {commonDialogService.dialogs.map((dialog, i, arr) => (
-        <NestedDialog key={dialog.id} visible={i === arr.length - 1} dialog={dialog} resolveDialog={state.resolve} rejectDialog={state.reject} />
-      ))}
-    </Loader>
+    <div className={s(styles, { backdrop: true })}>
```
devnaumov, line 56: Suspense was used to show a loading state when a dialog was processing something or triggering suspense. How does it work now?

webapp/packages/core-blocks/src/CommonDialog/CommonDialog/CommonDialogWrapper.tsx
```diff
@@ -24,50 +23,40 @@ export interface CommonDialogWrapperProps {
   fixedSize?: boolean;
   fixedWidth?: boolean;
   freeHeight?: boolean;
-  autofocus?: boolean;
   className?: string;
   children?: React.ReactNode;
+  autoFocusOnHide?: boolean | ((element: HTMLElement | null) => boolean) | undefined;
+  initialFocus?: HTMLElement | React.RefObject<HTMLElement | null> | null | undefined;
```
sergeyteleshev, line 30: maybe like this `export interface CommonDialogWrapperProps extends ExtendedDialogProps`?

webapp/packages/plugin-connection-custom/src/DriverSelector/DriverSelectorDialog.tsx
```diff
@@ -32,7 +32,16 @@ export const DriverSelectorDialog: DialogComponent<Payload> = observer(function
   });
 
   return (
-    <CommonDialogWrapper size="large" autofocus={false} fixedSize>
+    <CommonDialogWrapper
+      size="large"
+      autoFocusOnHide={() => {
+        const finalFocus = document.getElementById('slide-panel-close-button');
```
sergeyteleshev: better to move to const this id

webapp/packages/plugin-connection-custom/src/DriverSelector/DriverSelectorDialog.tsx
```diff
@@ -32,7 +32,16 @@ export const DriverSelectorDialog: DialogComponent<Payload> = observer(function
...
+        finalFocus?.focus();
+
+        return false;
+      }}
```
sergeyteleshev, line 51: may be implement this feature: via `context with ref + useContext()` here?

webapp/packages/core-blocks/src/CommonDialog/DialogsPortal.tsx
SychevAndrey: actually CommonDialogWrapper has a Loader suspense inside. Why do we need two Loaders?

webapp/packages/core-blocks/src/CommonDialog/CommonDialog/CommonDialogWrapper.tsx
SychevAndrey: we keep interfaces without dependencies

webapp/packages/plugin-connection-custom/src/DriverSelector/DriverSelectorDialog.tsx
SychevAndrey: didn't want to couple with core-app plugin

sergeyteleshev (dismissed):

devnaumov (dismissed):

Wroud (approved):

Issue #2290, opened 2024-01-10
[Security] Run Cloudbeaver as non-root by default (to be less vulnerable)


luarx:
**Is your feature request related to a problem? Please describe.**
It would be a good point to **run Cloudbeaver with a non-root user to follow best security practises**
Reference of why this is important: https://docs.bitnami.com/tutorials/why-non-root-containers-are-important-for-security

**Describe the solution you'd like**
To do that, it should define a USER in the [Dockerfile](https://github.com/dbeaver/cloudbeaver/blob/devel/deploy/docker/Dockerfile)

I see that s[omeone mentioned already this](https://github.com/dbeaver/cloudbeaver/issues/349) and suggested a solution, but it was not added to the repo as default

EvgeniaBzzz:
Hi @luarx!
Thank you for your request!
You are right, for now we do not have default option for that.
How to start a server with a non-root user - look [here](https://github.com/dbeaver/cloudbeaver/wiki/Run-Docker-Container#run-cloudbeaver-server-with-non-root-user).

luarx:
Thanks for your suggestion! Happy to know that it is possible to run as non-root and that it is documented 🙌 

On the other hand, I think that it should be the default option because of the security reasons that I shared unless there are some reasons to not do that...

luarx:
Wdyt @EvgeniaBzzz ? 😄

EvgeniaBzzz:
@luarx we will implement it in one of the future releases

brunnels:
I found the ubuntu user and group exist in the container already so no need to create a new user in the container.  I was able to work around this in my deployment by copying the `/opt/cloudbeaver` directory to an ephemeral volume and `chown ubuntu:ubuntu /opt/cloudbeaver` via a pod initContainer.    

The following could be added to the cloudbeaver Dockerfile and my workaround wouldn't be required.

```shell
RUN chown -R ubuntu:ubuntu /opt/cloudbeaver
```

PR #3526, opened 2025-06-16
dbeaver/dbeaver-devops#1604 Removed launch-product.sh and rewrited Do…

ggxed:
…ckerfile

PR #3964, opened 2025-12-08
Simple adjustment to run as non-root

hkolvenbach:
This adjustment changes the user to `dbeaver` (which is already created earlier as part of the `Dockerfile`) to run the container in environments, where root access is restricted, for example Kubernetes with `securityContext.runAsNonRoot: true`.

Addresses #2290.

**Note**: I do not have full visibility in all dependencies of dbeaver/cloudbeaver so I am unsure how to test if root access is required by any component. A local test running `cloudbeaver` with these adjustments was successful.

### DIFF
--- a/deploy/docker/cloudbeaver-ce/Dockerfile
+++ b/deploy/docker/cloudbeaver-ce/Dockerfile
@@ -18,4 +18,8 @@ WORKDIR /opt/cloudbeaver/
 
 RUN chmod +x "run-cloudbeaver-server.sh" "/opt/cloudbeaver/launch-product.sh"
 
+RUN chown -R dbeaver:dbeaver "workspace"
+
+USER dbeaver
+
 ENTRYPOINT ["./launch-product.sh"]


LonwoLonwo:
Hello @hkolvenbach 

Thanks for your contribution! It was a helpful solution to the issue. 

However, our team is also developing its own solution here: https://github.com/dbeaver/cloudbeaver/pull/3526.
Because we have a wide range of products and deployment options, numerous hidden issues cannot be taken care of solely for your PR purposes.
Anyway, thanks again for the attempt.

hkolvenbach:
Hi @LonwoLonwo, thanks for reviewing my suggestion and your feedback. Is running as non-root a planned feature for cloudbeaver-ce? The PR #3526 seems to set the correct permissions for the executable, but is not setting the Docker user to `dbeaver` (i.e. the `USER dbeaver` line that I added in my PR is missing from #3526)(https://github.com/dbeaver/cloudbeaver/pull/3964/commits/d3f78fe4ecb6164bc471d7526278883fdfd2e6da#diff-6ca13838ce6f722e84758eaa6113e1d883e0d97c81b9316989ce18c124fba85cR23)

PR #3972, opened 2025-12-09
dbeaver/pro#7090 fill fields

devnaumov:
closes [7090](https://github.com/dbeaver/pro/issues/7090)

I tried different approaches to realigning the layout, but they either have downsides or require a major redesign. So I think it’s better to just make the change we wanted and discuss how to improve the form during the ui discussion

image


Wroud (dismissed):

Wroud:
But maybe better to use `medium` or `large` to avoid too long fields in the edge cases

sergeyteleshev (dismissed):

sergeyteleshev (dismissed):

PR #3987, opened 2025-12-16
dbeaver/pro#7804 Support cb_session_id cookie for HTTP long-polling sessions

vkirst:


server/bundles/io.cloudbeaver.server/src/io/cloudbeaver/server/websockets/CBEventsLongPollingServlet.java
```diff
@@ -258,4 +251,78 @@ public static void sendError(
...
+        @Override
+        public BaseWebSession resolve(@NotNull HttpServletRequest req) {
+
+            String cookieSid = ServletAppUtils.getRequestCookie(req, CBConstants.CB_SESSION_COOKIE_NAME);
+
+            if (cookieSid == null) {
+                return null;
+            }
+
+            WebHttpRequestInfo info = new WebHttpRequestInfo(
```
serge-rider: Move it to createRequestInfo. We have pretty same code 3 times here.
Also we cannot use getRemoteAddr directly, see how we do it in https://github.com/dbeaver/pro/issues/6123

serge-rider (changes requested):

vkirst:
Resolved in https://github.com/dbeaver/pro/issues/7777

PR #3989, opened 2025-12-16
dbeaver/pro#7810 FIX: Generate SQL Item in Nav Tree Node Menu

SychevAndrey:
…ap.ts

### DIFF
--- a/webapp/packages/plugin-sql-generator/src/SqlGenerators/SqlGeneratorsBootstrap.ts
+++ b/webapp/packages/plugin-sql-generator/src/SqlGenerators/SqlGeneratorsBootstrap.ts
@@ -1,6 +1,6 @@
 /*
  * CloudBeaver - Cloud Database Manager
- * Copyright (C) 2020-2024 DBeaver Corp and others
+ * Copyright (C) 2020-2025 DBeaver Corp and others
  *
  * Licensed under the Apache License, Version 2.0.
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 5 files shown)


devnaumov (dismissed):

devnaumov (approved):

webapp/packages/plugin-sql-generator/src/SqlGenerators/SqlGeneratorsBootstrap.ts
```diff
@@ -30,17 +31,22 @@ export class SqlGeneratorsBootstrap extends Bootstrap {
   override register(): void {
     this.menuService.setHandler({
       id: 'node-sql-generators',
-      menus: [MENU_SQL_GENERATORS],
+      menus: [MENU_NAV_TREE],
```
sergeyteleshev, line 34: Unfortunately, this PR fixes the symptom, but not the source issue

It happened during the menu refactor. And the issue is sub-menus do not preload resources, so that we can definitely tell whether we should show the menu item (`isHidden`, `isApplicable`, etc.)

I added a loader here to see if it works, and it did (the screenshot)

https://github.com/dbeaver/cloudbeaver/blob/1d07d2728a9782ec82e220f9573a7b20124ee028/webapp/packages/core-ui/src/ContextMenu/SubMenuElement.tsx#L104

Image

sergeyteleshev (changes requested):

PR #4002, opened 2025-12-22
dbeaver/pro#7603 add long polling transport

devnaumov:
closes [7603](https://github.com/dbeaver/pro/issues/7603)

WebSocket ─┐
           ├─> Transport ──> SessionEventSource
Polling  ──┘

When a web socket throws an error, we switch to polling. If an error occurs during polling, we fall back to the old behavior (retries, etc.). 

Tested cases:

The socket stops working during the app’s lifetime
The socket does not work initially
The socket is blocked at the proxy level


SychevAndrey (dismissed):

sergeyteleshev (dismissed):

webapp/packages/core-root/src/LongPollingSubject.ts
```diff
@@ -0,0 +1,147 @@
...
+    this.output = new Subject();
+  }
+
+  override next(value: T): void {
+    this.send(value).catch(exception => {
+      this.output.error(exception);
+    });
+  }
+
+  async send(data: T): Promise<void> {
```
Wroud: i think it should be private

webapp/packages/core-root/src/LongPollingSubject.ts
```diff
@@ -0,0 +1,147 @@
...
+    const subscription = this.output.subscribe(subscriber);
+
+    subscriber.add(() => {
+      if (this.output.observers.length === 0) {
+        this.stopPolling();
+      }
+    });
+
+    return subscription;
+  }
```
Wroud, line 85: I'm not sure about this implementation. In RxJS, observables are lazy; they can be subscribed to multiple times, but if there are no listeners, all of them will be "cold"

webapp/packages/core-root/src/TransportSubject.ts
```diff
@@ -0,0 +1,106 @@
...
+          }
+
+          return throwError(() => err);
+        }),
+      )
+      .subscribe({
+        next: value => this.output.next(value),
+        error: err => this.output.error(err),
+        complete: () => this.output.complete(),
+      });
```
Wroud, line 97: i think we should use similar approach as socket.io to establish connection:
1. open long polling
2. try to open websocket same time
3. if websocket opened, close long pooling and use websocket

Wroud (dismissed):

SychevAndrey (approved):

PR #4020, opened 2025-12-25
7855 cb add setting   use the users os formatting for numbersdates

SychevAndrey:
Screenshot 2025-12-25 at 14 52 14
Screenshot 2025-12-25 at 14 51 58


webapp/packages/plugin-data-spreadsheet-new/src/DataGrid/Formatters/CellFormatters/DateTimeFormatter.tsx
```diff
@@ -0,0 +1,59 @@
...
+    if (!isNaN(date.getTime())) {
+      let formatter = tableDataContext.useUserFormatting.dateTime;
+
+      if (isDateOnly) {
+        formatter = tableDataContext.useUserFormatting.dateOnly;
+      }
+
+      value = formatter.format(date);
+    }
+  }
```
sergeyteleshev: can be moved to helper function which may be unit tested later

webapp/packages/plugin-data-spreadsheet-new/src/DataGrid/Formatters/CellFormatters/NumberFormatter.tsx
```diff
@@ -0,0 +1,53 @@
...
+      value = tableDataContext.useUserFormatting.number.format(numberValue);
+    }
+  }
+
+  return (
+    <div className={s(style, { numberFormatter: true })}>
+      <div className={s(style, { numberFormatterValue: true })}>{value}</div>
+    </div>
+  );
+});
```
sergeyteleshev, line 52: same for moving calculation to helper

webapp/packages/plugin-data-spreadsheet-new/src/DataGridSettingsService.ts
```diff
@@ -55,6 +64,33 @@ export class DataGridSettingsService {
...
+      for (let j = 0; j < letters.length; j++) {
+        const code = letters[i]! + letters[j]!;
+        if (isLanguageCodeSupported(code)) {
+          codes.push(code);
+        }
+      }
+    }
+
+    return codes;
+  }
```
sergeyteleshev: may be just create a const? we can research which standard used right now. when it was lastly updated. I believe it is not this frequently updated so its gonna be fine

webapp/packages/plugin-data-spreadsheet-new/src/DataGrid/Formatters/CellFormatters/NumberFormatter.tsx
SychevAndrey: I didn't understand. What calculations are you talking about?

webapp/packages/plugin-data-spreadsheet-new/src/DataGrid/Formatters/CellFormatters/DateTimeFormatter.tsx
SychevAndrey: This is the really straightforward and non-reusable piece of code, where we just create a boolean and use different formatter based on that value. What we can test here? Even if we extract it to a helper, what will be there? Formatters should come from outside, so here we can check only that isDateOnly is defined correctly which is strange?

webapp/packages/plugin-data-spreadsheet-new/src/DataGrid/Formatters/CellFormatters/DateTimeFormatter.tsx
```diff
@@ -0,0 +1,59 @@
...
+  const displayValue = getComputed(() => formatter.getDisplayString(valueHolder));
+
+  if (nullValue) {
+    return <GridNullFormatter />;
+  }
+
+  let value = displayValue;
+
+  if (tableDataContext.useUserFormatting) {
+    const isDateOnly = /^\d{4}-\d{2}-\d{2}$/.test(displayValue);
```
Wroud: `const isDateOnly = /^\d{4}-\d{2}-\d{2}$/.test(displayValue);`
Is it really required? This will hit performance in the grid noticeably, at least create a regex outside of the render function.
`new Date()` should be able to parse a date correctly without time. Maybe you can simplify the check by checking the resulting Date object (probably it will have some constant time)

(or just use unified formatter)

Wroud: Also, it's probably enough to check this format only once for one of the cells because they all will have the same format, or maybe you can check it by column type

Wroud (dismissed):

webapp/packages/plugin-data-spreadsheet-new/src/DataGrid/Formatters/CellFormatters/DateTimeFormatter.tsx
SychevAndrey: Yes, we can put it on the data context level and set something like "extended datakind" for every column with TIME/DATE/DATETIME values. In the cell we will just get the datakind of the column and use the according formatter. Do you want me refactor like that?

Image

Unfortunately, we can't rely on new Date to check the shape, so testing by regex or splitting is required. However, if we do it only on the TableDataContext level it should be not a problem.

webapp/packages/plugin-data-spreadsheet-new/src/DataGridSettingsService.ts
SychevAndrey: created a list with languages + regions

webapp/packages/plugin-data-spreadsheet-new/src/DataGridSettingsService.ts
```diff
@@ -36,7 +41,12 @@ export class DataGridSettingsService {
     return this.settings.getValue('plugin.data-spreadsheet.showDescriptionInHeader');
   }
 
+  get useUserFormatting(): string {
+    return this.settings.getValue('plugin.data-spreadsheet.useUserFormatting');
+  }
+
   readonly settings: SettingsProvider<DataGridSettingsSchema>;
+  readonly supportedLocales = this.getSupportedLocalesWithRegions();
```
Wroud: please use constructor for initial values

webapp/packages/plugin-data-spreadsheet-new/src/DataGridSettingsService.ts
```diff
@@ -55,6 +65,23 @@ export class DataGridSettingsService {
...
+        return false;
+      }
+    });
+  }
+
+  private getLocaleName(localeCode: string) {
+    const languageNameInItsOwnLang = new Intl.DisplayNames(localeCode, { type: 'language' });
+    const name = languageNameInItsOwnLang.of(localeCode);
+    return name && name?.length > 2 ? name.slice(0, 1).toLocaleUpperCase() + name.slice(1) : localeCode;
+  }
```
Wroud: can be declared as separate functions outside of the class

webapp/packages/plugin-data-spreadsheet-new/src/DataGridSettingsService.ts
```diff
@@ -17,10 +17,15 @@ import {
...
 import { DATA_EDITOR_SETTINGS_GROUP } from '@cloudbeaver/plugin-data-viewer';
+import { COMMON_LOCALES } from './commonLocales.js';
+
+export const NO_FORMAT = 'default';
+export const OS_FORMAT = '_OS';
 
 const defaultSettings = schema.object({
   'plugin.data-spreadsheet.hidden': schemaExtra.stringedBoolean().default(false),
   'plugin.data-spreadsheet.showDescriptionInHeader': schemaExtra.stringedBoolean().default(true),
+  'plugin.data-spreadsheet.useUserFormatting': schema.string().default(NO_FORMAT),
```
Wroud: let's rename it to the `formatLocale`

webapp/packages/plugin-data-spreadsheet-new/src/DataGrid/useTableData.tsx
```diff
@@ -83,6 +85,33 @@ export function useTableData(
...
         return Boolean(this.data?.columns?.some(column => (column as SqlResultColumn).description));
       },
+      get useUserFormatting(): IDataGridFormatters | null {
+        const setting = this.dataGridSettingsService.useUserFormatting;
+
+        if (setting === NO_FORMAT) {
+          return null;
+        }
+
+        const locale = setting === OS_FORMAT ? new Intl.DateTimeFormat().resolvedOptions().locale : setting;
```
Wroud: i think this part should be moved to the `dataGridSettingsService` so we will get `null` or `locale`, this will make it possible to not expose inner implementation details like `NO_FORMAT` and `OS_FORMAT`

webapp/packages/plugin-data-spreadsheet-new/src/DataGrid/helpers/detectDateTimeKind.ts
```diff
@@ -0,0 +1,21 @@
...
+ */
+
+import { DateTimeKind } from '../TableDataContext.js';
+
+export function detectDateTimeKind(displayValue: string): DateTimeKind {
+  if (/^\d{4}-\d{2}-\d{2}$/.test(displayValue)) {
+    return DateTimeKind.DateOnly;
+  }
+
+  if (/^\d{2}:\d{2}:\d{2}/.test(displayValue) && !displayValue.includes('-')) {
```
Wroud: please declare regex outside of the function body

webapp/packages/plugin-data-spreadsheet-new/src/DataGrid/useTableData.tsx
```diff
@@ -140,19 +169,42 @@ export function useTableData(
 
         return model.isReadonly(resultIndex) || (this.format.isReadOnly(key) && this.editor?.getElementState(key) !== DatabaseEditChangeType.add);
       },
+      getExtendedDateKind(columnKey: IGridColumnKey): DateTimeKind {
+        if (this.extendedDateKinds.has(columnKey.index)) {
+          return this.extendedDateKinds.get(columnKey.index) as DateTimeKind;
+        }
```
Wroud: probably you want to use `ResultSetCacheAction`

webapp/packages/plugin-data-spreadsheet-new/src/DataGrid/useTableData.tsx
```diff
@@ -28,15 +28,17 @@ import {
...
-import { DataGridSettingsService } from '../DataGridSettingsService.js';
+import { DataGridSettingsService, NO_FORMAT, OS_FORMAT } from '../DataGridSettingsService.js';
 import type { SqlResultColumn } from '@cloudbeaver/core-sdk';
 import { GridConditionalFormattingAction } from '@cloudbeaver/plugin-data-viewer-conditional-formatting';
+import { detectDateTimeKind } from './helpers/detectDateTimeKind.js';
 
 interface ITableDataPrivate extends ITableData {
   dataGridSettingsService: DataGridSettingsService;
   gridDIVElement: React.RefObject<HTMLDivElement | null>;
+  extendedDateKinds: Map<number, DateTimeKind>;
```
Wroud: maybe we can move formatting logic to separate hook+context? Formatting might become more complex over time

webapp/packages/plugin-data-spreadsheet-new/src/DataGrid/Formatters/CellFormatterFactory.tsx
```diff
@@ -38,6 +40,17 @@ export const CellFormatterFactory = observer<ICellFormatterProps>(function CellF
         if (resultColumn && isBooleanValuePresentationAvailable(holder.value, resultColumn)) {
           formatterRef.current = BooleanFormatter;
         }
+
+        if (tableDataContext.useUserFormatting) {
```
Wroud: `else if`

webapp/packages/plugin-data-spreadsheet-new/src/DataGrid/Formatters/CellFormatters/NumberFormatter.module.css
```diff
@@ -0,0 +1,18 @@
...
+.numberFormatter {
+  display: flex;
+  align-items: center;
+  overflow: hidden;
+}
+
+.numberFormatterValue {
+  overflow: hidden;
+  text-overflow: ellipsis;
+}
```
Wroud: can we use tailwind?

Wroud (changes requested):

webapp/packages/plugin-data-spreadsheet-new/src/DataGrid/useTableData.tsx
SychevAndrey: looks like an over-engineering to me since we don't need cell or row-level caching here, only the cache by columns

webapp/packages/plugin-data-spreadsheet-new/src/DataGrid/Formatters/CellFormatters/DateTimeFormatter.tsx
```diff
@@ -0,0 +1,61 @@
...
+      case DateTimeKind.DateTime:
+      case DateTimeKind.TimeOnly:
+        dateFormatter = formattingContext.formatters.dateTime;
+        break;
+      case DateTimeKind.DateOnly:
+        dateFormatter = formattingContext.formatters.dateOnly;
+        break;
+    }
+    const date = new Date(displayValue);
+    value = dateFormatter!.format(date);
```
sergeyteleshev: can we add default formatter here? which is gonna just display something as it is without formatting

cause now this line can break the code:
`value = dateFormatter!.format(date);`

better to have default switch case to prevent that

sergeyteleshev (dismissed):

webapp/packages/plugin-data-spreadsheet-new/src/DataGrid/Formatters/CellFormatterFactory.tsx
```diff
@@ -1,42 +1,86 @@
...
+  tableDataContext: ITableData;
+  hasFormatters: boolean;
+}
+
+type FormatterSelector = (context: IFormatterContext, cell: IGridDataKey) => React.FC<ICellFormatterProps> | null;
+
+const formatterSelectors: FormatterSelector[] = [
+  // Binary
+  (context, cell) => {
+    const holder = context.tableDataContext.getCellHolder(cell);
```
Wroud: you better to pass holder as an argument, because this selector might be slow and this is an optimization (you can get holder once and use multiple times)

webapp/packages/plugin-data-spreadsheet-new/src/DataGrid/FormattingContext.ts
```diff
@@ -0,0 +1,30 @@
...
+  dateOnly: Intl.DateTimeFormat;
+  number: Intl.NumberFormat;
+}
+
+export interface IFormattingContext {
+  formatters: IDataGridFormatters | null;
+  getExtendedDateKind: (columnKey: IGridColumnKey) => DateTimeKind;
+}
+
+export const FormattingContext = createContext<IFormattingContext>(undefined as any);
```
Wroud: `undefined as any` is dirty code, better to use:
```ts
export const FormattingContext = createContext<IFormattingContext | null>(null);

export function useFormattingContext() {
  const context = use(FormattingContext);
  if (!context) {
    throw new Error('Formatting Context required')
  }
  return context;
}
```

webapp/packages/plugin-data-spreadsheet-new/src/DataGrid/useTableData.tsx
Wroud: we need to add `getColumn` like we have `getRow` in it, it's better because it will clear cache when needed and place it in one place

webapp/packages/plugin-data-spreadsheet-new/src/DataGridSettingsService.ts
```diff
@@ -55,6 +78,20 @@ export class DataGridSettingsService {
...
 
+  getFormatLocale(): string | null {
+    const setting = this.formatLocale;
+
+    if (setting === NO_FORMAT) {
+      return null;
+    }
+
+    if (setting === OS_FORMAT) {
+      return new Intl.DateTimeFormat().resolvedOptions().locale;
```
Wroud: this operation is very slow, consider run it in the constructor or lazy (like `getOrCreateOSFormat` that will save it in private parameter and reuse)

Wroud (changes requested):

Wroud (approved):

sergeyteleshev (approved):

PR #4031, opened 2026-01-05
dbeaver/pro#7897 add connection view plugin

devnaumov:
closes [7897](https://github.com/dbeaver/pro/issues/7897)


server/bundles/io.cloudbeaver.server/schema/service.core.graphqls
```diff
@@ -549,6 +549,9 @@ type ConnectionInfo {
     "Supported features (provided etc)"
     features: [ String! ]!
     navigatorSettings: NavigatorSettings!
+    "Original navigator settings from the configuration file"
+    originalNavigatorSettings: NavigatorSettings! @since(version: "25.3.3")
```
Wroud: actually you can add optional argument for the `navigatorSettings` field, something like:
```gql
navigatorSettings(datasourceSettings: Boolean): NavigatorSettings!
```

webapp/packages/core-localization/src/locales/fr.ts
```diff
@@ -141,6 +141,7 @@ export default [
   ['ui_field_is_required', 'Please fill in this field'],
   ['ui_delete_confirmation_message', 'You are about to delete "{arg:item}". Are you sure?'],
   ['ui_not_selected', 'Not selected'],
+  ['ui_reset', 'Reset'],
```
sergeyteleshev, line 144: can be translated it to languages respectively using ai suggestions

server/bundles/io.cloudbeaver.server/schema/service.core.graphqls
Wroud: `userSettings` should be removed from the `NavigatorSettingsInput` type, instead you need to add similar argument to the `setConnectionNavigatorSettings`

`originalNavigatorSettings` better to rename to the `datasourceNavigatorSettings`

webapp/packages/core-sdk/src/queries/fragments/DatabaseConnectionNavigatorViewSettings.gql
```diff
@@ -0,0 +1,10 @@
+fragment DatabaseConnectionNavigatorViewSettings on ConnectionInfo {
+  id
+  projectId
+  navigatorSettings {
+    ...AllNavigatorSettings
+  }
+  originalNavigatorSettings {
+    ...AllNavigatorSettings
+  }
```
Wroud: probably you need to split it into two fragments and two resources, because `originalNavigatorSettings` needed only for connection editing form

webapp/packages/core-root/src/ConnectionNavigatorViewSettings.ts
```diff
@@ -1,14 +1,14 @@
...
+ * Copyright (C) 2020-2026 DBeaver Corp and others
  *
  * Licensed under the Apache License, Version 2.0.
  * you may not use this file except in compliance with the License.
  */
 import type { NavigatorSettingsInput } from '@cloudbeaver/core-sdk';
 
 export type NavigatorView = 'simple' | 'advanced';
-export type NavigatorViewSettings = Partial<NavigatorSettingsInput>;
+export type NavigatorViewSettings = Omit<NavigatorSettingsInput, 'showSystemObjects'> & { showSystemObjects?: boolean };
```
sergeyteleshev, line 11: why did we mark a thing that is required to optional? may be it should be optional first from the begin?

sergeyteleshev: also `DEFAULT_NAVIGATOR_VIEW_SETTINGS` in `ConnectionInfoResource` should be exported here, in this file

Wroud (changes requested):

webapp/packages/plugin-connection-view/src/ConnectionViewPart.ts
```diff
@@ -0,0 +1,61 @@
...
+  }
+
+  protected override async saveChanges(data: IFormState<IConnectionFormState>): Promise<void> {
+    if (!this.optionsPart.connectionKey) {
+      return;
+    }
+
+    await this.connectionViewService.changeConnectionView(this.optionsPart.connectionKey, {
+      ...this.state,
+      userSettings: false,
```
sergeyteleshev, line 59: why do we exclude it from the flow and set it here?

webapp/packages/plugin-connection-view/src/ConnectionViewForm.tsx
```diff
@@ -0,0 +1,44 @@
...
+    viewFormPart.state = { ...viewFormPart.state, ...view };
+  }
+
+  return (
+    <Group form gap>
+      <GroupTitle>{translate('plugin_connection_view')}</GroupTitle>
+      <Container gap dense>
+        <FieldCheckbox id="simple" checked={isSimple} onChange={changeView}>
+          {translate('plugin_connection_view_option_simple')}
+        </FieldCheckbox>
```
sergeyteleshev: I suggest to use select here so we can also see 2 options: simple and advanced like it is in connection context menu

server/bundles/io.cloudbeaver.server/schema/service.core.graphqls
devnaumov: I think that, in this case, we restricted ourselves by using a single interface. Imagine you want to add an additional field only to global settings or remove userSettings field from global interface (as its always false). When you have two entities, navigatorSettings and datasourceSettings, you can do this easily by changing just one interface. I think datasource settings and user settings should be strictly separated at the gql level.

server/bundles/io.cloudbeaver.model/src/io/cloudbeaver/model/WebConnectionInfo.java
```diff
@@ -283,7 +284,11 @@ public String[] getFeatures() {
 
     @Property
     @NotNull
-    public DBNBrowseSettings getNavigatorSettings() {
+    public DBNBrowseSettings getNavigatorSettings(@NotNull DataFetchingEnvironment environment) {
```
serge-rider: we mustn't use GQL-specific types in web services

serge-rider (changes requested):

webapp/packages/plugin-connection-view/src/ConnectionViewPart.ts
devnaumov: In the connection form, we always submit settings as global, and we ensure that this is the case here

webapp/packages/core-root/src/ConnectionNavigatorViewSettings.ts
devnaumov: https://github.com/dbeaver/cloudbeaver/pull/3831/files#diff-b2727084f6c6c4bea4af47bd09ecc3523a52d2b4db85be2b9e9c0f9400e3b977R11

The same approach was used here, this is due to how our navigator view model is working right now

webapp/packages/plugin-connection-view/src/ConnectionViewForm.tsx
devnaumov: I am not sure, its whether simple or not. For example in administration part its just simple or not. Probably we can change it to switch and change label if we need to. Will sync it with a team

webapp/packages/core-sdk/src/queries/fragments/DatabaseConnectionNavigatorViewSettings.gql
devnaumov: The settings are fairly lightweight to resolve (they’re mostly cached, confirmed with the backend team). Only setting new settings is performance heavy. I don’t like the idea of splitting such similar entities into two almost identical resources (when the there is no resolve time issue). Without a separate endpoint for fetching default settings, and with the current userSettings approach, this doesn’t seem like a significant improvement. However, if we want to be strict about separating resources by place of use, we can do it.

Wroud (dismissed):

sergeyteleshev (dismissed):
I verified Copilot's suggestions regarding translations. Could you fix it? It seems reasonable

webapp/packages/plugin-connection-view/src/ConnectionViewResource.ts
```diff
@@ -0,0 +1,137 @@
...
+    this.aliases.add(ConnectionInfoActiveProjectKey, () =>
+      resourceKeyList(this.keys.filter(key => projectsService.activeProjects.some(({ id }) => id === key.projectId))),
+    );
+
+    this.sync(this.connectionInfoResource);
+    this.connectionInfoResource.onItemDelete.addHandler(this.delete.bind(this));
+  }
+
+  async changeConnectionView(key: IConnectionInfoParams, settings: NavigatorViewSettings): Promise<NavigatorViewSettingsInfo> {
+    const connectionNavigatorViewSettings = this.get(key)?.navigatorSettings || DEFAULT_NAVIGATOR_VIEW_SETTINGS;
```
Wroud: do we really need to spread original settings? it seems redundant, this can be done on the backend side because flags is optional and frontend can send only changed flags to the backend

Wroud (changes requested):

Wroud (dismissed):

SychevAndrey (approved):

PR #4041, opened 2026-01-09
dbeaver/pro#5145 add rename action for connection

devnaumov:
closes [5145](https://github.com/dbeaver/pro/issues/5145)


sergeyteleshev (dismissed):

webapp/packages/core-navigation-tree/src/NodesManager/NavTreeResource.ts
```diff
@@ -301,14 +302,18 @@ export class NavTreeResource extends CachedMapResource<string, string[], Record<
...
         this.markLoaded(node.id);
         this.onDataOutdated.execute(parentId);
 
+        if (node.objectFeatures.includes(EObjectFeature.dataSource)) {
+          return node.id;
+        }
+
+        const parts = node.id.split('/');
+        parts.splice(parts.length - 1, 1, name);
+        const path = parts.join('/');
```
Wroud: if it possible we should avoid such conditions in the generic places

webapp/packages/plugin-navigation-tree/src/NodesManager/NavNodeContextMenuService.ts
```diff
@@ -113,7 +114,7 @@ export class NavNodeContextMenuService extends Bootstrap {
 
         if (NodeManagerUtils.isDatabaseObject(node.id) || isConnectionFolder(node)) {
           if (action === ACTION_RENAME) {
-            return node.features?.includes(ENodeFeature.canRename) ?? false;
+            return (node.features?.includes(ENodeFeature.canRename) || node.objectFeatures.includes(EObjectFeature.dataSource)) ?? false;
```
Wroud: seems strange, why backend will not send `canRename` feature?

webapp/packages/core-navigation-tree/src/NodesManager/NavTreeResource.ts
devnaumov: We can ignore it on the subscription level, but then we need to do it in every place, the connection node is just a node which id behaves differently when renaming, i think it would be strange to submit event with invalid nodeId, 

I think the best optoin is to return new id from backend

Wroud: agree, best option is to get new id from the backend

Wroud (approved):

sergeyteleshev (approved):

SychevAndrey (approved):

PR #4051, opened 2026-01-13
[CB] Use the Tab key for autocomplete

sergeyteleshev:
closes https://github.com/dbeaver/pro/issues/7555

https://github.com/user-attachments/assets/7d9ae254-edee-4f3a-97fb-990b71577086


webapp/packages/core-view/src/Action/KeyBinding/createKeyBinding.ts
```diff
@@ -11,7 +11,6 @@ interface IKeyBindingOptions {
   id: string;
   preventDefault?: boolean;
   keys?: string | string[];
-  keysWin?: string | string[];
   keysMac?: string | string[];
```
Wroud, line 15: i don't think that you should remove OS specific keys, because in different OS you might want to have completely different shortcuts (not just Ctrl -> Option mapping)

Wroud (changes requested):

webapp/packages/core-view/src/Action/KeyBinding/getCommonAndOSSpecificKeys.ts
```diff
@@ -14,24 +14,23 @@ export function getCommonAndOSSpecificKeys(keyBinding: IKeyBinding | undefined):
...
+    return getKeys(keyBinding.keysMac);
   }
 
-  return keys;
+  return [];
 }
 
 function getKeys(keys: string[] | string | undefined): string[] {
-  return Array.isArray(keys) ? keys : [keys ?? ''].filter(Boolean);
+  return Array.isArray(keys) ? keys.filter(Boolean) : [keys ?? ''].filter(Boolean);
```
Wroud:
```suggestion
  return (Array.isArray(keys) ? keys : [keys ?? '']).filter(Boolean);
```

webapp/packages/plugin-codemirror6/src/createEditorAutocompletion.ts
```diff
@@ -1,22 +1,32 @@
...
+  preventDefault: true,
+};
+
 const EDITOR_AUTOCOMPLETION_COMPARTMENT = new Compartment();
 
-const EDITOR_AUTOCOMPLETION_KEYMAP = keymap.of([
-  { key: 'Alt-Space', run: startCompletion, preventDefault: true },
-  { key: 'Shift-Ctrl-Space', run: startCompletion, preventDefault: true },
-]);
+const EDITOR_AUTOCOMPLETION_KEYMAP = Prec.high(keymap.of([EDITOR_START_COMPLETION_KEYBINDING, EDITOR_ACCEPT_COMPLETION_KEYBINDING]));
```
Wroud: why `Alt-Space` removed?

webapp/packages/plugin-codemirror6/src/createEditorAutocompletion.ts
```diff
@@ -1,22 +1,32 @@
...
+import { acceptCompletion, autocompletion, startCompletion } from '@codemirror/autocomplete';
+import { Compartment, Prec, type Extension } from '@codemirror/state';
+import { keymap, type KeyBinding } from '@codemirror/view';
 
 export type CompletionConfig = Parameters<typeof autocompletion>[0];
 
+// Shortcuts are case sensitive
+export const CODEMIRROR_SHORTCUT_SPLITTER = '-';
+export const EDITOR_START_COMPLETION_KEYBINDING: KeyBinding = {
+  key: ['Shift', 'Mod', 'Space'].join(CODEMIRROR_SHORTCUT_SPLITTER),
```
Wroud:
```suggestion
  key: 'Shift-Mod-Space',
```

webapp/packages/plugin-codemirror6/src/createEditorAutocompletion.ts
```diff
@@ -1,22 +1,32 @@
...
+  run: startCompletion,
+  preventDefault: true,
+};
+export const EDITOR_ACCEPT_COMPLETION_KEYBINDING: KeyBinding = {
+  key: ['Tab'].join(CODEMIRROR_SHORTCUT_SPLITTER),
```
Wroud:
```suggestion
  key: 'Tab',
```

webapp/packages/plugin-sql-editor/src/actions/bindings/mapCodemirrorShortcuts.ts
```diff
@@ -0,0 +1,17 @@
...
+import { SHORTCUT_DIVIDER, type IKeyBinding } from '@cloudbeaver/core-view';
+import { CODEMIRROR_SHORTCUT_SPLITTER, type KeyBinding } from '@cloudbeaver/plugin-codemirror6';
+
+export function mapCodemirrorShortcuts(keyBinding: KeyBinding): Omit<IKeyBinding, 'id'> {
+  return {
+    keys: keyBinding.key?.split(CODEMIRROR_SHORTCUT_SPLITTER).join(SHORTCUT_DIVIDER) ?? '',
+    keysMac: keyBinding.key?.split(CODEMIRROR_SHORTCUT_SPLITTER).join(SHORTCUT_DIVIDER) ?? '',
+    preventDefault: keyBinding.preventDefault,
+  };
+}
```
Wroud: you can declare `CODEMIRROR_SHORTCUT_SPLITTER` here without exporting it

Wroud (changes requested):

webapp/packages/plugin-codemirror6/src/createEditorAutocompletion.ts
sergeyteleshev: cause I think one shortcut is enough. reverted

Wroud (dismissed):

SychevAndrey (dismissed):

webapp/packages/core-view/src/Action/KeyBinding/getBindingLabel.ts
```diff
@@ -29,12 +29,15 @@ const FORMAT_SHORTCUT_KEYS_MAP: Record<string, string> = {
...
-const SOURCE_DIVIDER_REGEXP = /\+/gi;
-const APPLIED_DIVIDER = ' + ';
 
-function transformKeys(keyBinding: IKeyBinding): string[] {
-  return getCommonAndOSSpecificKeys(keyBinding).map(shortcut =>
+export const SHORTCUT_DIVIDER = '+';
+export const SOURCE_DIVIDER_REGEXP = /\+/gi;
+export const APPLIED_DIVIDER = ` ${SHORTCUT_DIVIDER} `;
+
+export function transformKeys(keyBinding: IKeyBinding): string[] {
```
sergeyteleshev, line 39: `getShortcutLabelKeys` added because we have a case when we add something like this

`{
    keys: 'mod+space',
    macKeys: 'alt+space'
}`

on mac `mod + space` dont work because it is OS reserved shortcut for switching language. and in the docs helper for shortcuts we gonna see 2 shortcuts: `mod+space or alt+space`. and this causes to see for user 1 invalid and 1 valid shortcuts. so I decided we now show only valid os specific shortcuts in helper
but register in `useHotKeys` all of them (not only what showed in helper now)

Wroud: If `keys: 'mod+space'` doesn't work on some OS, then maybe it's better to specify it only for OS where it works (win, Linux). The idea behind `keys` is that it should be used for universal hotkeys that work across different OS

SychevAndrey (approved):

Wroud (approved):

PR #4052, opened 2026-01-13
dbeaver/pro#7769 fix properties mapping for enums

HocKu7:
Closes https://github.com/dbeaver/pro/issues/7769

What changed

Now, values inside ObjectPropertyInfo (validValues and defaultValue) can be objects with { displayName, value }.

I created two additional functions to use when iterating over validValues or when getting a property value without considering the data type as getObjectPropertyValue fn relies on property type (for example, in the properties table).
Also there is a custom interface for object property so we can change type there and get type errors in our codebase. Previously there were not errors as types were always any.

However, it’s becoming clear that ObjectPropertyInfo is getting more complex to work with. Maybe we should consider implementing an ObjectViewerPropertyMapper class and use it like ObjectViewerPropertyMapper(property).defaultValue. If we always map properties through this object, the behavior would be more predictable in theory.


webapp/packages/core-sdk/src/getObjectPropertyValue.ts
```diff
@@ -43,5 +43,21 @@ function getValue(value: any, controlType: ObjectPropertyType) {
...
+
+  return option;
+}
+
+export function getObjectPropertyOptionName(option: ObjectPropertyOption): string {
+  if (typeof option === 'object' && option !== null) {
+    return option.displayName;
+  }
+
+  return String(option);
```
Wroud, line 66: i think that if the property is `null` or `undefined` is better to display empty string (for it's name)

webapp/packages/core-sdk/src/getObjectPropertyValue.ts
```diff
@@ -43,5 +43,21 @@ function getValue(value: any, controlType: ObjectPropertyType) {
     return value.join(', ');
   }
 
-  return value.displayName || value.value || JSON.stringify(value);
+  return value.value || value.displayName || JSON.stringify(value);
```
Wroud, line 46: this change will affect behaviour of the `getObjectPropertyDisplayValue` probably in unexpected way

webapp/packages/plugin-connections/src/ConnectionForm/Options/ConnectionFormOptionsPart.ts
```diff
@@ -548,8 +556,10 @@ function prepareDynamicProperties(
       delete result[propertyInfo.id];
     } else {
       const isDefault = isNotNullDefined(propertyInfo.defaultValue);
+
       if (!(propertyInfo.id in result) && isDefault) {
-        result[propertyInfo.id] = propertyInfo.defaultValue;
+        const defaultValue = getObjectPropertyDefaultValue(propertyInfo);
+        result[propertyInfo.id] = defaultValue;
```
```suggestion
        result[propertyInfo.id] = getObjectPropertyDefaultValue(propertyInfo);
```

webapp/packages/plugin-connections/src/ConnectionForm/SSL/ConnectionFormSSLPart.ts
```diff
@@ -121,7 +121,8 @@ export class ConnectionFormSSLPart extends FormPart<INetworkHandlerConfig, IConn
         const isDefault = isNotNullDefined(descriptorProperty.defaultValue);
 
         if (!(key in handlerConfig.properties) && isDefault) {
-          handlerConfig.properties[key] = descriptorProperty.defaultValue;
+          const defaultValue = getObjectPropertyDefaultValue(descriptorProperty);
+          handlerConfig.properties[key] = defaultValue;
```
Wroud:
```suggestion
          handlerConfig.properties[key] = getObjectPropertyDefaultValue(descriptorProperty);
```

webapp/packages/core-sdk/src/getObjectPropertyValue.ts
sergeyteleshev: Surely will but its right now. We just need to run auto-tests and fix the cases appeared in-place

sergeyteleshev (dismissed):

webapp/packages/core-sdk/src/getObjectPropertyValue.ts
devnaumov: I think this is kind of expected. getObjectPropertyDisplayValue is intended to get property.value, always as a string, not property.displayName. Falling back to displayName is more of a last resort here and probably should be removed, since it’s not related to the property value. 

The good news is that we use getObjectPropertyDisplayValue only in the metadata viewer, so it will be easy to check.

Wroud (dismissed):

yagudin10 (approved):

PR #4056, opened 2026-01-14
Dbeaver/pro#7975 Added row identifier data

Nexus6v2:
Added row identifier state and type/name to support different data editor states
https://dbeaver.com/docs/dbeaver/Data-Editor/#table-status-indicators

Closes Dbeaver/pro#7975


SychevAndrey (dismissed):

server/bundles/io.cloudbeaver.server/src/io/cloudbeaver/service/sql/WebSQLResultSetRowIdentifier.java
```diff
@@ -0,0 +1,31 @@
+package io.cloudbeaver.service.sql;
```
LonwoLonwo, line 17: Copyright

server/bundles/io.cloudbeaver.server/schema/service.sql.graphqls
```diff
@@ -139,6 +143,13 @@ type SQLResultSet {
     readOnlyStatus: String @since(version: "25.0.1")
 }
 
+type SQLRowIdentifier {
```
serge-rider, line 153: I thought we want exact info about the identifier (its type, unique name and set of columns)

server/bundles/io.cloudbeaver.server/src/io/cloudbeaver/service/sql/WebSQLQueryDataReceiver.java
```diff
@@ -182,9 +186,24 @@ public void fetchEnd(@NotNull DBCSession session, @NotNull DBCResultSet resultSe
...
-        webResultSet.setHasRowIdentifier(!rowIdentifiers.isEmpty() && hasRowIdentifier);
+        DBDRowIdentifier rowIdentifier = resultsInfo.getDefaultRowIdentifier();
+        if (rowIdentifier == null) {
+            webResultSet.setRowIdentifierState(WebSQLResultSetRowIdentifier.TABLE_METADATA_NOT_FOUND);
+        } else if (rowIdentifier.getUniqueKey().getConstraintType().equals(DBSEntityConstraintType.VIRTUAL_KEY)) {
+            webResultSet.setRowIdentifierState(WebSQLResultSetRowIdentifier.VIRTUAL_KEY);
+        } else if (!rowIdentifier.isIncomplete() && rowIdentifier.isValidIdentifier()) {
+            webResultSet.setHasRowIdentifier(true);
+            webResultSet.setRowIdentifierState(WebSQLResultSetRowIdentifier.PRIMARY_KEY);
+            String rowIdentifierName = rowIdentifier.getAttributes().stream()
```
serge-rider: It has to be constraint name. not concatenated column names

serge-rider (changes requested):

server/bundles/io.cloudbeaver.server/src/io/cloudbeaver/service/sql/WebSQLQueryDataReceiver.java
Nexus6v2: This is the same logic we have in desktop UI, in case of a composite key, all columns are concatenated for a display
Do you suggest we change it there as well ?

server/bundles/io.cloudbeaver.server/schema/service.sql.graphqls
Nexus6v2: I can extend the API with additional data, but this is all we need for current UI
It follows the same format we have in desktop for rowIdentifier representation

server/bundles/io.cloudbeaver.server/src/io/cloudbeaver/service/sql/WebSQLResultSetRowIdentifier.java
Nexus6v2: Added, thanks

server/bundles/io.cloudbeaver.server/src/io/cloudbeaver/service/sql/WebSQLQueryDataReceiver.java
serge-rider: Please send a link on the code where we use this logic. It doesn't sound right

server/bundles/io.cloudbeaver.server/schema/service.sql.graphqls
serge-rider: Yes, we need full information about identifier.
Desktop has full access to all details of DBDRowIdentifier, it is needed in many different places.

server/bundles/io.cloudbeaver.server/src/io/cloudbeaver/service/sql/WebSQLQueryDataReceiver.java
Nexus6v2: Here's the part I've mentioned
https://github.com/dbeaver/dbeaver/blob/788ac25ca63c5c02d5dae34f154200e1816f8d1b/plugins/org.jkiss.dbeaver.ui.editors.data/src/org/jkiss/dbeaver/ui/controls/resultset/spreadsheet/SpreadsheetPresentation.java#L2128-L2132

And how it looks it UI
image

server/bundles/io.cloudbeaver.server/schema/service.sql.graphqls
```diff
@@ -126,7 +126,11 @@ type SQLResultSet {
     # server always returns hasMoreData = false
     hasMoreData: Boolean!
     "Identifies if result set has row identifier. If true then it is possible update data or load LOB files"
-    hasRowIdentifier: Boolean!
+    hasRowIdentifier: Boolean! @deprecated (reason: "use rowIdentifierState (25.3.4)")
+    "State of a table row identifier: METADATA_NOT_FOUND, NONE, PRIMARY_KEY, VIRTUAL_KEY"
+    rowIdentifierState: String @since(version: "25.3.4")
```
Wroud: seems like we should use `enum` for `rowIdentifierState`

Wroud (changes requested):

server/bundles/io.cloudbeaver.server/schema/service.sql.graphqls
Nexus6v2: Added enum

server/bundles/io.cloudbeaver.server/schema/service.sql.graphqls
Nexus6v2: Extended rowIdentifier type to contain all attributes

Wroud (dismissed):

PR #4058, opened 2026-01-14
[CB] Improve undo/redo behavior for Data Editor

sergeyteleshev:
closes https://github.com/dbeaver/pro/issues/7614


webapp/packages/plugin-data-viewer/src/DatabaseDataModel/Actions/Grid/GridEditAction.ts
```diff
@@ -544,4 +560,52 @@ export class GridEditAction<
       update.type = DatabaseEditChangeType.update;
     }
   }
+
+  undoSelectedCellValue(): void {
```
Wroud, line 564: should not be here, in this action you might have only simple actions like 'undo' / 'redo' optionally for provided cell

PR #4063, opened 2026-01-16
Critical UX Issue: Delete key marks entire row as deleted during inline cell editing

### LINKED ISSUE #4061: Critical UX Issue: Delete key marks entire row as deleted during inline cell editing
### Description

When editing table data directly in CloudBeaver (inline cell editing), pressing the Delete key marks the entire row as deleted instead of deleting a single character in the active cell.
The affected row is visually marked as deleted (red). If this state is not noticed and the user commits the transaction, the complete record is permanently deleted.
This behavior is highly error-prone and can easily lead to unintended data loss, especially for users accustomed to standard text editing behavior.

Impact
High risk of accidental data loss
Violates common UX expectations
Especially dangerous in production databases

Image

### Steps to reproduce

Open a table in CloudBeaver
Enable direct (inline) editing of a cell
Place the cursor inside a text field
Press the Delete key
Observe that the entire row is marked as deleted

### Expected/Desired Behavior

The Delete key should delete a character to the right of the cursor within the active cell, consistent with standard text editing behavior
Row deletion should only be possible via explicit actions (context menu, toolbar action, or dedicated shortcut with confirmation)

**Actual Behavior**
Pressing Delete during inline cell editing marks the entire row as deleted
No warning or confirmation is shown
The deletion is committed silently if the user commits the transaction

**One or more of the following should be implemented:**
Disable row deletion via Delete key during inline cell editing
Restrict Delete key to cell-level editing only
Require an explicit confirmation dialog before marking a row as deleted
Introduce a distinct keyboard shortcut for row deletion
Improve visual feedback and warnings before commit when rows are marked for deletion

### CloudBeaver Version

25.3.2

### Additional context

Currently, deleting characters safely is only possible using the Backspace key, which is non-obvious and undocumented behavior.

sergeyteleshev:
closes https://github.com/dbeaver/cloudbeaver/issues/4061

### DIFF
--- a/webapp/packages/plugin-data-viewer/src/DATA_VIEWER_KEY_BINDINGS.ts
+++ b/webapp/packages/plugin-data-viewer/src/DATA_VIEWER_KEY_BINDINGS.ts
@@ -24,7 +24,7 @@ export const KEY_BINDING_DUPLICATE_ROW: IKeyBinding = {
 
 export const KEY_BINDING_DELETE_ROW: IKeyBinding = {
   id: 'data-viewer-delete-row',
-  keys: ['mod+shift+delete', 'mod+shift+backspace'],
+  keys: ['mod+shift+delete'],
 };
 
 export const KEY_BINDING_SAVE: IKeyBinding = {
--- a/webapp/packages/plugin-data-viewer/src/DATA_VIEWER_KEY_BINDINGS.ts
+++ b/webapp/packages/plugin-data-viewer/src/DATA_VIEWER_KEY_BINDINGS.ts
@@ -24,8 +24,7 @@ export const KEY_BINDING_DUPLICATE_ROW: IKeyBinding = {
 
 export const KEY_BINDING_DELETE_ROW: IKeyBinding = {
   id: 'data-viewer-delete-row',
-  keys: ['Delete'],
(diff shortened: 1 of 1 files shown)


SychevAndrey:
when did we decide to change the shortcut?

Screenshot 2026-01-19 at 10 21 52

And I see the strange _FN + Backspace_ shortcut in in Shortcuts dialog, which doesn't work

sergeyteleshev:
> when did we decide to change the shortcut?

I decided during this PR cause it fixes the original issue, because shortcuts intersect. Also this matches the saving shortcut (mod+shift+s & mod+shift+backspace)

> And I see the strange FN + Backspace shortcut in the Shortcuts dialog, which doesn't work

Yes, we have this issue, but it should have its own ticket. Moreover, my PR fixes this issue in this spot, but the issue is in displaying shortcut mapping.

SychevAndrey:
Actually, _FN + Backspace_ seems to be working on my second keyboard.

SychevAndrey (approved):
I believe this is a significant change to our users and should be approved by the team.

sergeyteleshev:
Agreed on call to keep this version

Wroud (approved):

PR #4072, opened 2026-01-19
dbeaver/pro#4692 show spinner when loading new page

devnaumov:
closes [4692](https://github.com/dbeaver/pro/issues/4692)

We could show spinner based on the key loading state but we want to make sure we are responding to the certain action and not just key loading

CleanShot 2026-01-20 at 17 11 45@2x


SychevAndrey (dismissed):

webapp/packages/plugin-navigation-tree/src/NavigationTree/ElementsTree/NavTreeLimitFilter/elementsTreeLimitRenderer.tsx
```diff
@@ -29,20 +31,35 @@ const NavTreeLimitMessage: NavigationNodeRendererComponent = observer(function N
...
+    return (
+      <div className='tw:px-6 tw:py-1'>
+        <Spinner size='small' />
+      </div>
+    );
+  }
+
   return (
-    <div className={s(styles, { loadMoreBox: true })}>
+    <div className='tw:text-(--theme-text-hint-on-light) theme-typography--caption tw:px-6 tw:py-1'>
```
sergeyteleshev: Let's disable this link (or make it unclickable) if the resource is loading. cause we don't wanna `loadMore` something if the data is about to change
Moreover, this button is probably easy to spam with these `loadMore`s. If so we should fix it also

Wroud: probably the best option will be use `Button` component with `loader` prop

Wroud: why not use `theme-text-text-hint-on-light theme-typography--caption` instead of `tw:text-(--theme-text-hint-on-light) theme-typography--caption`

devnaumov: Changed it to Button

sergeyteleshev (dismissed):

webapp/packages/plugin-navigation-tree/src/NavigationTree/ElementsTree/NavTreeLimitFilter/elementsTreeLimitRenderer.tsx
```diff
@@ -1,20 +1,22 @@
...
+
+import { useState } from 'react';
 import { observer } from 'mobx-react-lite';
 
-import { Link, s, useTranslate } from '@cloudbeaver/core-blocks';
+import { useTranslate } from '@cloudbeaver/core-blocks';
 import { useService } from '@cloudbeaver/core-di';
 import { NavTreeResource } from '@cloudbeaver/core-navigation-tree';
 import { CachedResourceOffsetPageKey, CachedResourceOffsetPageTargetKey, getNextPageOffset } from '@cloudbeaver/core-resource';
+import { Button } from '@dbeaver/ui-kit';
```
Wroud: please use Button from the `@cloudbeaver/core-blocks`

Wroud (changes requested):

webapp/packages/plugin-navigation-tree/src/NavigationTree/ElementsTree/NavTreeLimitFilter/elementsTreeLimitRenderer.tsx
Wroud: Moreover you can pass a prop `loader` and make `onClick` return a promise (already), so you don't need to initiate internal `loading` state in the component

Wroud (approved):

PR #4080, opened 2026-01-22
dbeaver/pro#7383 make pointer target at least 24px

devnaumov:
closes [7383](https://github.com/dbeaver/pro/issues/7383)

There is one ui change I’m not sure about. It’s a context menu in the data viewer. Previously, there was equal spacing between the text on the left and the menu trigger on the right, but now the menu trigger has a box that isn’t visually visible, and it creates the effect that the left and right sides are not aligned. I think it’s fine, though, if we want to show a hover effect on the trigger, it will look fine because the box will be aligned with the cell box.

image
image
image
image


SychevAndrey:
image
Could we move this menu button to the right to place it how it was before?

SychevAndrey (approved):

sergeyteleshev (approved):

Wroud (approved):

PR #4085, opened 2026-01-26
7023-cb-find-and-replace-new

SychevAndrey:
https://github.com/user-attachments/assets/7e90e7a7-e5bb-4cdc-b82f-ee74202d5a19


webapp/common-react/@dbeaver/react-data-grid/src/search/GridSearchEngine.ts
```diff
@@ -0,0 +1,93 @@
...
+    return matches;
+  }
+
+  /** Replace pattern in cell text. Returns new text and whether pattern still matches. */
+  static replaceInCell(cellText: string, pattern: RegExp, replaceValue: string): { newText: string; stillMatches: boolean } {
+    const newText = cellText.replace(pattern, replaceValue);
+
+    if (pattern.global) {
+      pattern.lastIndex = 0;
+    }
```
Wroud: is it safe to mutate object from the argument?

webapp/common-react/@dbeaver/react-data-grid/src/DataGrid.tsx
```diff
@@ -106,6 +117,74 @@ export const DataGrid = forwardRef<DataGridRef, DataGridProps>(function DataGrid
...
+    setSearchOpen(true);
+    setTimeout(() => searchPanelRef.current?.focus(), 0);
+  }
+
+  function handleSearchClose() {
+    setSearchOpen(false);
+    const cached = searchStorage?.get();
+    if (cached) {
+      searchStorage?.set({ ...cached, open: false });
+    }
```
Wroud: let's move to the internal hook

webapp/common-react/@dbeaver/react-data-grid/src/DataGrid.tsx
```diff
@@ -106,6 +117,74 @@ export const DataGrid = forwardRef<DataGridRef, DataGridProps>(function DataGrid
...
+    const el = containerRef.current;
+    if (!el) {
+      return;
+    }
+
+    function handleKeyDown(e: KeyboardEvent) {
+      if ((e.ctrlKey || e.metaKey) && e.key === 'f') {
+        e.preventDefault();
+        setSearchOpen(true);
+        setTimeout(() => searchPanelRef.current?.focus(), 0);
```
Wroud: `setTimeout(() => searchPanelRef.current?.focus(), 0);`
isn't it better to handle this logic inside search panel component in render layout hook?

webapp/common-react/@dbeaver/react-data-grid/src/DataGrid.tsx
```diff
@@ -106,6 +117,74 @@ export const DataGrid = forwardRef<DataGridRef, DataGridProps>(function DataGrid
...
+      }
+    }
+
+    el.addEventListener('keydown', handleKeyDown);
+    return () => el.removeEventListener('keydown', handleKeyDown);
+  }, []);
+
+  function handleSearchOpen() {
+    setSearchOpen(true);
+    setTimeout(() => searchPanelRef.current?.focus(), 0);
```
Wroud: same here

webapp/common-react/@dbeaver/react-data-grid/src/DataGrid.tsx
```diff
@@ -106,6 +117,74 @@ export const DataGrid = forwardRef<DataGridRef, DataGridProps>(function DataGrid
...
+  }
+
+  function handleSearchClose() {
+    setSearchOpen(false);
+    const cached = searchStorage?.get();
+    if (cached) {
+      searchStorage?.set({ ...cached, open: false });
```
Wroud: maybe we can make `update` method that will accept partial state for update?

webapp/common-react/@dbeaver/react-data-grid/src/search/GridSearchPanel.tsx
```diff
@@ -0,0 +1,102 @@
...
+export interface GridSearchPanelRef {
+  focus: () => void;
+  refresh: () => void;
+}
+
+interface GridSearchPanelProps {
+  columnCount: number;
+  scrollToCell: (position: { rowIdx: number; colIdx: number }) => void;
+  replaceCellValue: (rowIdx: number, colIdx: number, value: string) => void;
+  replaceCellValues?: (updates: ICellValueUpdate[]) => void;
```
Wroud: `onReplace`
let's keep only one and rename it

webapp/common-react/@dbeaver/react-data-grid/src/search/GridSearchPanel.tsx
```diff
@@ -0,0 +1,102 @@
...
+    replaceCellValues,
+    onReplacingChange,
+    storage,
+    open,
+  });
+
+  useEffect(() => {
+    onCellClassNameChange(getCellClassName);
+    return () => onCellClassNameChange(undefined);
+  }, [getCellClassName, onCellClassNameChange]);
```
Wroud: you better to expose `getCellClassName` via `useImperativeHandle`

webapp/common-react/@dbeaver/react-data-grid/src/search/GridSearchEngine.ts
```diff
@@ -0,0 +1,93 @@
...
+        const flags = options.caseSensitive ? 'g' : 'gi';
+        return new RegExp(query, flags);
+      }
+
+      let escapedSearch = query.replace(ESCAPE_REGEX, '\\$&');
+      if (options.wholeWord) {
+        escapedSearch = `\\b${escapedSearch}\\b`;
+      }
+
+      const flags = options.caseSensitive ? 'g' : 'gi';
```
sergeyteleshev: duplicated twice. can be moved to the top of function
plus lets move `g` and `gi` to consts with name so no more magic numbers/letters

webapp/common-react/@dbeaver/react-data-grid/src/search/GridSearchEngine.ts
```diff
@@ -0,0 +1,93 @@
...
+      return new RegExp(escapedSearch, flags);
+    } catch {
+      return null;
+    }
```
sergeyteleshev: this is a silent quit. which from the code perspective basically equals to nothing have been found

what if user enables regexp, then tries something that causes an error. as result it sees - nothing have found. but in fact there is an error. maybe we can show the error to the user just to know the search query is screwed and need to try something else. it would improve ux for the users for sure

webapp/common-react/@dbeaver/react-data-grid/src/search/GridSearchEngine.ts
```diff
@@ -0,0 +1,93 @@
...
+      for (let colIdx = 0; colIdx < columnCount; colIdx++) {
+        const cellText = getCellText(rowIdx, colIdx);
+        if (searchPattern.test(cellText)) {
+          matches.push({ rowIdx, colIdx });
+          if (searchPattern.global) {
+            searchPattern.lastIndex = 0;
+          }
+        }
+      }
+    }
```
sergeyteleshev: well for big tables this would just crash the app due to performance issues in the worst case scenario. for most cases it would be like a freeze

we need something tricky here. maybe batch search in new thread (service worker)

I suppose not a bad idea to sync with the team how should we handle that so we do not dig into this real hard

also we have `react-minisearch` lib for searching. maybe it can offer something for us. we need to check the API for ideas

webapp/common-react/@dbeaver/react-data-grid/src/search/GridSearchPanel.tsx
```diff
@@ -0,0 +1,102 @@
...
+
+  useImperativeHandle(ref, () => ({
+    focus: () => panelRef.current?.focus(),
+    refresh: () => actions.refresh(),
```
sergeyteleshev: better to add it to function declaration in order to improve debug experience

webapp/common-react/@dbeaver/react-data-grid/src/search/GridSearchPanel.tsx
```diff
@@ -0,0 +1,102 @@
...
+    <SearchPanel
+      ref={panelRef}
+      className={clsx('rdg-search-panel tw:bg-(--theme-secondary)', isReadOnly && 'tw:px-0!')}
+      isReadOnly={isReadOnly || !replaceCellValue}
+      query={{
+        search: snapshot.query,
+        replace: snapshot.replace,
+        caseSensitive: snapshot.caseSensitive,
+        wholeWord: snapshot.wholeWord,
+        regexp: snapshot.regexp,
```
sergeyteleshev, line 84: maybe it is the same interface? suspiciously similar

webapp/common-react/@dbeaver/react-data-grid/src/search/GridSearchPanel.tsx
```diff
@@ -0,0 +1,102 @@
...
+  function handleClose() {
+    actions.close();
+    onClose();
+  }
+
+  return (
+    <SearchPanel
+      ref={panelRef}
+      className={clsx('rdg-search-panel tw:bg-(--theme-secondary)', isReadOnly && 'tw:px-0!')}
+      isReadOnly={isReadOnly || !replaceCellValue}
```
sergeyteleshev: `replaceCellValue` is always a function and truthy

webapp/common-react/@dbeaver/react-data-grid/src/search/useGridSearch.ts
```diff
@@ -0,0 +1,533 @@
...
+function debouncedSearch(store: GridSearchStore): void {
+  if (store.debounceTimer !== null) {
+    clearTimeout(store.debounceTimer);
+  }
+  store.debounceTimer = setTimeout(() => {
+    if (!store.disposed) {
+      runSearch(store);
+    }
+  }, DEFAULT_DEBOUNCE_MS);
+}
```
sergeyteleshev: we have debounce in core-utils. maybe its time to move it in `common-typescript/js-helpers`

webapp/common-react/@dbeaver/react-data-grid/src/search/useGridSearch.ts
```diff
@@ -0,0 +1,533 @@
...
+      notifyListeners(store);
+      syncState(store);
+      debouncedSearch(store);
+    },
+
+    toggleRegex(): void {
+      store.regexp = !store.regexp;
+      invalidateSnapshot(store);
+      notifyListeners(store);
+      syncState(store);
```
sergeyteleshev: these 3 always go in a row. move it to `update()` or something like that?

webapp/common-react/@dbeaver/react-data-grid/src/search/useGridSearch.ts
```diff
@@ -0,0 +1,533 @@
...
+        return;
+      }
+
+      const pattern = buildPattern(store);
+      if (!pattern) {
+        return;
+      }
+
+      store.options.onReplacingChange?.(true);
+      try {
```
sergeyteleshev: no catch statement so can be removed

webapp/common-react/@dbeaver/react-data-grid/src/search/useGridSearch.ts
```diff
@@ -0,0 +1,533 @@
...
```
sergeyteleshev: same

webapp/common-react/@dbeaver/react-data-grid/src/search/useGridSearch.ts
```diff
@@ -0,0 +1,533 @@
...
+    [store],
+  );
+  const getSnapshotFn = useCallback(() => getSnapshot(store), [store]);
+  const snapshot = useSyncExternalStore(subscribe, getSnapshotFn);
+
+  const [actions] = useState(() => createActions(store));
+  const [getCellClassName] = useState(() => createCellClassName(store));
+
+  return { snapshot, actions, getCellClassName, replaceOpen: store.replaceOpen };
+}
```
sergeyteleshev: all in all I love the functional approach. it would be easy to cover with unit tests like that

webapp/common-react/@dbeaver/react-data-grid/src/DataGrid.tsx
```diff
@@ -106,6 +117,74 @@ export const DataGrid = forwardRef<DataGridRef, DataGridProps>(function DataGrid
...
+    }
+  }
+
+  function handleReplacingChange(value: boolean) {
+    isReplacingRef.current = value;
+  }
+
+  function scrollToCell(position: { rowIdx: number; colIdx: number }) {
+    innerGridRef.current?.scrollToCell({ idx: position.colIdx, rowIdx: position.rowIdx });
+  }
```
sergeyteleshev: please provide correct predefined types for arguments

webapp/common-react/@dbeaver/react-data-grid/src/DataGrid.tsx
```diff
@@ -143,6 +222,10 @@ export const DataGrid = forwardRef<DataGridRef, DataGridProps>(function DataGrid
 
   useImperativeHandle(ref, () => ({
     selectCell: (position: ICellPosition) => {
+      if (isReplacingRef.current) {
+        return;
+      }
```
sergeyteleshev, line 185: why cant we select the cell we are about to replace with new value?

webapp/common-react/@dbeaver/react-data-grid/src/DataGrid.tsx
```diff
@@ -189,6 +276,9 @@ export const DataGrid = forwardRef<DataGridRef, DataGridProps>(function DataGrid
   );
 
   function handleCellFocus(args: CellSelectArgs<IInnerRow, unknown>) {
+    if (isReplacingRef.current) {
+      return;
+    }
```
sergeyteleshev, line 251: same. why?

webapp/packages/plugin-data-spreadsheet-new/src/DataGrid/DataGridTable.tsx
```diff
@@ -429,6 +464,23 @@ export const DataGridTable = observer<IDataPresentationProps>(function DataGridT
     tableData.editor?.set({ row, column }, value);
   }
 
+  function handleCellChangeBatch(changes: Array<{ rowIdx: number; colIdx: number; value: string }>) {
```
sergeyteleshev: strict types pls

webapp/packages/plugin-data-spreadsheet-new/src/DataGrid/DataGridTable.tsx
```diff
@@ -189,13 +199,20 @@ export const DataGridTable = observer<IDataPresentationProps>(function DataGridT
         return;
       }
 
+      if (dataGridRef.current?.isReplacing()) {
+        return;
+      }
```
sergeyteleshev: I think we should revert that new behavior

As user I want to see all changes focused. For example we have huge table and our change is out of the screen right now. Currently if I replace 1 value I wont see what was replaced cause it was not scrolled or selected this cell. so we can miss something important

webapp/packages/plugin-data-viewer/src/DatabaseDataModel/Actions/Document/DocumentEditAction.ts
```diff
@@ -93,6 +93,12 @@ export class DocumentEditAction<
     this.removeUnchanged(key);
   }
 
+  setMany(updates: Array<{ key: TKey; value: TValue }>): void {
+    for (const { key, value } of updates) {
+      this.set(key, value);
+    }
+  }
```
sergeyteleshev, line 98: this will cause a lot of actions dispatched and reactions subscribed to this action. so UI maybe can went bananas
maybe we should create a editBatch where we provide multiple edited cells, edit it in batch and execute 1 action

webapp/packages/plugin-data-viewer/src/DatabaseDataModel/Actions/Grid/GridEditAction.ts
```diff
@@ -195,6 +195,35 @@ export class GridEditAction<
...
+
+    this.historyManager.recordBatchCellEdit({ updates: historyUpdates });
+
+    this.action.execute({
+      resultId: this.result.id,
+      type: DatabaseEditChangeType.update,
+      revert: false,
+      value: actionValues,
+    });
+  }
```
sergeyteleshev, line 223: like this yes

webapp/packages/plugin-data-viewer/src/DatabaseDataModel/Actions/IDatabaseDataCacheAction.ts
```diff
@@ -9,13 +9,20 @@ import { createService } from '@cloudbeaver/core-di';
...
   has(key: TKey, scope: symbol): boolean;
   get<T>(key: TKey, scope: symbol): T | undefined;
   set<T>(key: TKey, scope: symbol, value: T): void;
   delete(key: TKey, scope: symbol): void;
   deleteAll(scope: symbol): void;
+
+  hasGlobal(scope: symbol): boolean;
+  getGlobal<T>(scope: symbol): T | undefined;
+  setGlobal<T>(scope: symbol, value: T): void;
+  deleteGlobal(scope: symbol): void;
```
sergeyteleshev: maybe better to move it to separate cache action: `IDatabaseGlobalCacheAction` where we can store global values for the `data editor`
now it is mixed with cells values so we need to separate abstractions

sergeyteleshev (changes requested):
All in all feature looks awesome! Nice integration with undo/redo btw

Please also consider cases with readonly tables (probably block replace functionality with disabled state) - Session Manager, readonly tables, tables where we cannot edit cells, but can add/delete rows 

Also, keep in mind that if we replace 1 item. We want to scroll and focus this item because it may be outside the visible view, and we need to show the user that this value has changed. For "replace all" the view should remain the same (no selection, no scroll to cell)

Some of the issues can be resolved in the next tickets. We need to decide what we can do now and what later

webapp/common-react/@dbeaver/react-data-grid/src/search/GridSearchPanel.tsx
SychevAndrey: thanks for that catch. that check was from the time when replace function could not be provided. 

Also fixed the condition, replace is not shown in read-only tables:
image

webapp/common-react/@dbeaver/react-data-grid/src/search/GridSearchEngine.ts
SychevAndrey: image
this simple solution still shows decent results even on huge tables. I would take it to separate ticket for further exploration. 

I tried to use Web Worker for that calculations, but the bottleneck here is the getCellText which can't be called somewhere else. Batching, serializing/deserializing, orchestration adds too much overhead and didn't show any significant improvements. Actually, for me it even for worse

webapp/common-react/@dbeaver/react-data-grid/src/search/GridSearchEngine.ts
SychevAndrey: This is how the search works in Value Panel right now. If we want to show something, we need to groom it first.

webapp/common-react/@dbeaver/react-data-grid/src/search/GridSearchPanel.tsx
SychevAndrey: with what interface?

webapp/common-react/@dbeaver/react-data-grid/src/search/useGridSearch.ts
SychevAndrey: discussed, needed for finally

webapp/common-react/@dbeaver/react-data-grid/src/DataGrid.tsx
SychevAndrey: I didn't quite understand, but I updated highlighted code

webapp/common-react/@dbeaver/react-data-grid/src/DataGrid.tsx
SychevAndrey: because it will break the flow, we don't want to loose focus from replace input

webapp/common-react/@dbeaver/react-data-grid/src/DataGrid.tsx
SychevAndrey: same

webapp/packages/plugin-data-viewer/src/DatabaseDataModel/Actions/Document/DocumentEditAction.ts
SychevAndrey: It was not needed actually

webapp/packages/plugin-data-viewer/src/DatabaseDataModel/Actions/IDatabaseDataCacheAction.ts
SychevAndrey: I'm not sure. We might get confused later by two cache actions :) Let's discuss

webapp/common-react/@dbeaver/react-data-grid/src/search/GridSearchEngine.ts
devnaumov: We’ll need to come up with a solution anyway. Users with large datasets will definitely run into issues, but we can address performance problems on large datasets in a separate ticket. The main thing for now is that the new functionality should not affect users who don’t use the search and replace feature

webapp/common-react/@dbeaver/react-data-grid/src/search/GridSearchEngine.ts
```diff
@@ -0,0 +1,97 @@
...
+
+    return matches;
+  }
+
+  /** Replace pattern in cell text. Returns new text and whether pattern still matches. */
+  static replaceInCell(cellText: string, pattern: RegExp, replaceValue: string): { newText: string; stillMatches: boolean } {
```
devnaumov: Why do we choose to use static methods and class approach in general here? Seems like it could be 3 functions instead

webapp/common-react/@dbeaver/react-data-grid/src/search/GridSearchPanel.tsx
```diff
@@ -0,0 +1,110 @@
...
+    scrollToCell,
+    replaceCellValue,
+    replaceCellValues,
+    onReplacingChange,
+    storage,
+    open,
+  });
+
+  useEffect(() => {
+    onCellClassNameChange(getCellClassName);
```
devnaumov: This part is confusing for me. We are reacting to the fn reference change here? Could you describe what its used for please

webapp/common-react/@dbeaver/react-data-grid/src/DataGrid.tsx
```diff
@@ -214,44 +304,61 @@ export const DataGrid = forwardRef<DataGridRef, DataGridProps>(function DataGrid
...
+                  renderRow: rowRenderer,
+                  renderCell: cellRenderer,
+                  noRowsFallback: children,
+                }}
+                onScroll={onScroll}
+                onSelectedCellChange={handleCellFocus}
+                onCellKeyDown={handleCellKeyDown}
+                onColumnWidthsChange={setColumnWidths}
+              />
+              {searchOpen && (
```
devnaumov: We can use `<Activity />`  here to preserve state and improve perfomance

webapp/common-react/@dbeaver/react-data-grid/src/DataGrid.tsx
```diff
@@ -106,6 +117,74 @@ export const DataGrid = forwardRef<DataGridRef, DataGridProps>(function DataGrid
...
+        setSearchOpen(true);
+        setTimeout(() => searchPanelRef.current?.focus(), 0);
+      }
+    }
+
+    el.addEventListener('keydown', handleKeyDown);
+    return () => el.removeEventListener('keydown', handleKeyDown);
+  }, []);
+
+  function handleSearchOpen() {
```
devnaumov: Can we move all search related logic to a separate hook? so we dont implement it all in the entry file?

webapp/common-react/@dbeaver/react-data-grid/src/DataGrid.tsx
```diff
@@ -51,13 +53,19 @@ export interface DataGridProps extends IDataGridCellContext, IDataGridRowContext
   onEditorOpen?: (position: ICellPosition) => void;
   onCellKeyDown?: (position: ICellPosition, event: DataGridCellKeyboardEvent) => void;
   className?: string;
+  searchReadOnly?: boolean;
```
Wroud: do we need `searchReadOnly`? we have `getCellEditable`

webapp/packages/plugin-data-viewer/src/DatabaseDataModel/Actions/Grid/GridEditHistoryManager.ts
```diff
@@ -44,6 +45,15 @@ export class GridEditHistoryManager<TKey extends IGridDataKey, TCell> {
...
 
+  recordBatchCellEdit(data: IGridHistoryBatchCellUpdateData<TKey, TCell>): void {
+    this.compressLastEditedCellHistory();
+
+    this.history.add({
+      source: GRID_HISTORY_SOURCE.BATCH_EDIT_CELLS,
+      data,
+    });
+  }
+
```
Wroud: let's rework `GRID_HISTORY_SOURCE.EDIT_CELL` to accept array of changes, there is no need in the multiple event types for every type of change

Wroud: moreover, `batch` is not an a special event, any event can be batched, so it usually solved on the event structure side or event emmiter

webapp/packages/plugin-data-viewer/src/DatabaseDataModel/Actions/ResultSet/ResultSetEditAction.ts
```diff
@@ -73,6 +73,27 @@ export class ResultSetEditAction extends GridEditAction<SqlResultColumn, SqlResu
...
+          value = createResultSetContentValue({
+            text: String(value),
+            contentLength: String(value).length,
+            contentType: prevValue.contentType ?? 'text/plain',
+          });
+        }
+      }
+
+      return { key, value };
+    });
```
Wroud: please avoid logic duplication, this is very important to have same code for `set` and `setMany`, make a private method to transform data

webapp/packages/plugin-data-viewer/src/DatabaseDataModel/Actions/IDatabaseDataCacheAction.ts
Wroud: i don't like suffix `Global` but agree that it's better to have single cache action

webapp/common-react/@dbeaver/react-data-grid/src/search/GridSearchEngine.ts
SychevAndrey: You are right, it's not needed here, was just an effect of previous version of the search.

webapp/common-react/@dbeaver/react-data-grid/src/search/GridSearchEngine.ts
SychevAndrey: It looked safe, but I'm not 100% sure, so copied a regexp.

webapp/common-react/@dbeaver/react-data-grid/src/search/GridSearchPanel.tsx
SychevAndrey: We are "registering" getCellClassName in DataGrid to enable highlighting on panel mount, and cleaning it on unmount to remove highlighting.

webapp/common-react/@dbeaver/react-data-grid/src/search/GridSearchPanel.tsx
SychevAndrey: a ref can't trigger the parent's re-render needed to propagate the value into context, so we would end up needing a callback or effect on the parent side anyway

webapp/common-react/@dbeaver/react-data-grid/src/DataGrid.tsx
SychevAndrey: searchReadOnly is a global param that we may use to hide replace functionality. I don't think `getCellEditable` will help here

webapp/packages/plugin-data-viewer/src/DatabaseDataModel/Actions/IDatabaseDataCacheAction.ts
SychevAndrey: Was choosing between Meta and Shared suffixes and decided to go with the latest. Do you like it more?

webapp/common-react/@dbeaver/react-data-grid/src/search/GridSearchPanel.tsx
Wroud: this is not true, you can pass callback to the `ref` and use `useState` to track ref changes, moreover `getCellClassName` seems to be stable callback (it is not changing during re-renders) so there is no need to track it (it will be tracked correctly with ref)

webapp/common-react/@dbeaver/react-data-grid/src/DataGrid.tsx
```diff
@@ -169,13 +202,16 @@ export const DataGrid = forwardRef<DataGridRef, DataGridProps>(function DataGrid
...
-    const previousRowCount = rowsCountRef.current;
-    rowsCountRef.current = rowsCount;
+  if (prevRowsCount !== rowsCount) {
+    setPrevRowsCount(rowsCount);
 
-    if (previousRowCount === 0) {
+    if (prevRowsCount === 0) {
       setColumnWidths(new Map<string, ColumnWidth>());
     }
   }
```
Wroud, line 238: why do we need this changes? it seems not related to the initial issue

webapp/packages/plugin-data-viewer/src/DatabaseDataModel/Actions/DatabaseEditAction.ts
```diff
@@ -54,6 +54,7 @@ export abstract class DatabaseEditAction<
   abstract getElementState(key: TKey): DatabaseEditChangeType | null;
   abstract get(key: TKey): TValue | undefined;
   abstract set(key: TKey, value: TValue): void;
+  abstract setMany(updates: Array<{ key: TKey; value: TValue }>): void;
```
Wroud, line 56: maybe it's better to use `Array<[TKey, TValue]>`?

webapp/packages/plugin-data-viewer/src/DatabaseDataModel/Actions/IDatabaseDataEditAction.ts
```diff
@@ -55,6 +55,7 @@ export interface IDatabaseDataEditAction<
   getElementState: (key: TKey) => DatabaseEditChangeType | null;
   get: (key: TKey) => TValue | undefined;
   set: (key: TKey, value: TValue) => void;
+  setMany: (updates: Array<{ key: TKey; value: TValue }>) => void;
```
Wroud, line 58: same here

webapp/packages/plugin-data-spreadsheet-new/src/DataGrid/DataGridTable.tsx
```diff
@@ -190,13 +198,20 @@ export const DataGridTable = observer<IDataPresentationProps>(function DataGridT
...
 
     tableData.editor?.action.addHandler(syncEditor);
 
     function syncFocus(data: DatabaseDataSelectActionsData<Partial<IGridDataKey>>) {
       if (data.type === 'focus') {
+        if (dataGridRef.current?.isReplacing()) {
+          return;
+        }
         // TODO: we need this delay to update focus after render rows update
         setTimeout(() => {
```
Wroud: i belive this should be handled on the grid component side, if you don't want to change focus or any, you can:
1. Block focus change event in the `Grid` component
2. Block focus change method in the `Grid` component

webapp/common-react/@dbeaver/react-data-grid/src/DataGrid.tsx
SychevAndrey: https://github.com/dbeaver/cloudbeaver/pull/4085/changes/99861f8128bf426d9eb2e2be1dab3ae0ca956054

`Cannot access refs during render`

webapp/packages/plugin-data-viewer/src/DatabaseDataModel/Actions/Grid/GridEditHistoryManager.ts
```diff
@@ -36,7 +36,8 @@ export class GridEditHistoryManager<TKey extends IGridDataKey, TCell> {
   }
 
   recordCellEdit(data: IGridHistoryCellUpdateData<TKey, TCell>): void {
-    this.compressLastEditedCellHistory(data.key);
+    const singleUpdateKey = data.updates.length === 1 ? data.updates[0]!.key : undefined;
+    this.compressLastEditedCellHistory(singleUpdateKey);
```
Wroud: can we use `if` here and restrict passing `undefined` to the `compressLastEditedCellHistory` because it make no sense

webapp/common-react/@dbeaver/react-data-grid/src/DataGrid.tsx
Wroud: https://react.dev/reference/react/useRef
see pitfall

webapp/packages/plugin-data-viewer/src/DatabaseDataModel/Actions/DatabaseEditAction.ts
SychevAndrey: I'm not sure that tuple is better here.
Even if we look at the usage, compare:

    updates.push({ key: { row, column }, value });
    updates.push([{ row, column }, value ]);

For me the first option is more readable.

And if we look at edit actions payloads they are also shaped like objects:

    this.historyManager.recordCellEdit({
      updates: [{ key, value, prevValue }],
    });

webapp/common-react/@dbeaver/react-data-grid/src/search/useGridSearch.ts
```diff
@@ -0,0 +1,452 @@
...
+  });
+  const matchedSetRef = useRef(initialDerived.matchedSet);
+  const activeMatchKeyRef = useRef(initialDerived.activeMatchKey);
+  const queryRef = useRef(state.query);
+
+  const lastGridSizeRef = useRef({ rowCount: options.rowCount, columnCount: options.columnCount });
+  const isInitialMountRef = useRef(true);
+
+  const runSearchAndScroll = useCallback(
+    (searchState: Pick<GridSearchState, 'query' | 'caseSensitive' | 'wholeWord' | 'regexp'>, preserveActiveIndex: boolean): void => {
```
Wroud: `Pick<GridSearchState, 'query' | 'caseSensitive' | 'wholeWord' | 'regexp'>` -> `GridSearchQueryState`

webapp/common-typescript/@dbeaver/js-helpers/src/keyboardEventFlags.ts
```diff
@@ -0,0 +1,21 @@
...
+  [STOP_CAPTURE_VIEW_HOTKEYS_FLAG]?: boolean;
+};
+
+export function stopCaptureViewHotkeys(event: object): void {
+  (event as TEventWithFlags)[STOP_CAPTURE_VIEW_HOTKEYS_FLAG] = true;
+}
+
+export function isCaptureViewHotkeysStopped(event: object): boolean {
+  return Boolean((event as TEventWithFlags)[STOP_CAPTURE_VIEW_HOTKEYS_FLAG]);
+}
```
Wroud: there is `EventContext` for such purpose, i think you can move it from the `cloudbeaver` codebase to `js-helpers`

webapp/packages/core-view/src/View/CaptureView.tsx
```diff
@@ -69,7 +70,7 @@ export const CaptureView = observer<React.PropsWithChildren<ICaptureViewProps>>(
       enableOnFormTags: ['INPUT', 'SELECT', 'TEXTAREA'],
       preventDefault(event, handler) {
         // Don't prevent default if event was already handled by a child view
-        if (EventContext.has(event, EventStopPropagationFlag)) {
+        if (isCaptureViewHotkeysStopped(event) || EventContext.has(event, EventStopPropagationFlag)) {
```
Wroud: maybe you can use `EventStopPropagationFlag`?

webapp/common-typescript/@dbeaver/js-helpers/src/keyboardEventFlags.ts
```diff
@@ -0,0 +1,21 @@
+/*
+ * CloudBeaver - Cloud Database Manager
+ * Copyright (C) 2020-2026 DBeaver Corp and others
+ *
+ * Licensed under the Apache License, Version 2.0.
+ * you may not use this file except in compliance with the License.
+ */
+
+const STOP_CAPTURE_VIEW_HOTKEYS_FLAG = '__cloudbeaver_stop_capture_view_hotkeys__';
```
Wroud: you can't have `capture view` in the common codebase because there is no `capture view` there

Wroud (changes requested):

webapp/common-react/@dbeaver/react-data-grid/src/search/useGridSearch.ts
SychevAndrey: nice catch, that code appeared here before QueryState :D

webapp/common-react/@dbeaver/react-data-grid/src/search/GridSearchPanel.tsx
```diff
@@ -0,0 +1,104 @@
...
+) {
+  const panelRef = useRef<SearchPanelRef>(null);
+  const { cellText } = useContext(DataGridCellContext) ?? {};
+  const rowContext = useContext(DataGridRowContext);
+  const rowCount = useGridReactiveValue(rowContext?.rowCount);
+
+  const { snapshot, actions, getCellClassName, replaceOpen } = useGridSearch({
+    rowCount: rowCount ?? 0,
+    columnCount,
+    getCellText: (r, c) => cellText?.get(r, c) ?? '',
```
Wroud: i think it's better to pass reactive value here

webapp/common-react/@dbeaver/react-data-grid/src/search/state.ts
```diff
@@ -0,0 +1,135 @@
...
+  }
+
+  for (let i = 0; i < left.length; i++) {
+    const leftMatch = left[i];
+    const rightMatch = right[i];
+
+    if (!leftMatch || !rightMatch || leftMatch.rowIdx !== rightMatch.rowIdx || leftMatch.colIdx !== rightMatch.colIdx) {
+      return false;
+    }
+  }
```
Wroud:
```suggestion
  if (left.some((left, i) => left.rowIdx !== right[i]!.rowIdx || left.colIdx !== right[i]!.colIdx)) {
    return false;
  }
```

webapp/common-react/@dbeaver/react-data-grid/src/search/useGridSearch.ts
```diff
@@ -0,0 +1,452 @@
...
+        opts.columnCount,
+        opts.getCellText,
+      );
+
+      dispatch({ type: 'SET_MATCHES', matchedCells: matches, preserveActiveIndex });
+
+      const newActiveIdx = computeActiveIdx(matches, currentState.activeMatchIdx, preserveActiveIndex);
+      scrollToMatch(opts.scrollToCell, matches[newActiveIdx]);
+    },
+    [dispatch],
```
Wroud: > The dispatch function has a stable identity, so you will often see it omitted from Effect dependencies, but including it will not cause the Effect to fire. If the linter lets you omit a dependency without errors, it is safe to do.
https://react.dev/reference/react/useReducer#caveats

webapp/common-react/@dbeaver/react-data-grid/src/search/useGridSearch.ts
```diff
@@ -0,0 +1,452 @@
...
+  useEffect(() => {
+    if (isInitialMountRef.current) {
+      isInitialMountRef.current = false;
+      return;
+    }
+
+    if (!state.query) {
+      dispatch({ type: 'SET_MATCHES', matchedCells: [], preserveActiveIndex: false });
+      return;
+    }
```
Wroud, line 413: i think this should be part of the `SET_QUERY`

webapp/common-react/@dbeaver/react-data-grid/src/search/useGridSearch.ts
```diff
@@ -0,0 +1,452 @@
...
+    if (st.matchedCells.length > 0) {
+      return;
+    }
+
+    if (!st.query) {
+      return;
+    }
+
+    // Defer to allow grid to render first
+    const timeoutId = setTimeout(() => {
```
Wroud: Why? Can you please describe the behaviour without this code

Wroud (changes requested):

webapp/common-react/@dbeaver/react-data-grid/src/search/useGridSearch.ts
SychevAndrey: yes, good idea

webapp/common-react/@dbeaver/react-data-grid/src/search/state.ts
SychevAndrey: I thought you prefer for loops with explicit checks instead of functional code like here. It's definitely shorter, but I'm not sure it's really better

webapp/packages/plugin-data-spreadsheet-new/src/DataGrid/DataGridTable.tsx
```diff
@@ -197,11 +209,7 @@ export const DataGridTable = observer<IDataPresentationProps>(function DataGridT
 
     function syncFocus(data: DatabaseDataSelectActionsData<Partial<IGridDataKey>>) {
       if (data.type === 'focus') {
-        // TODO: we need this delay to update focus after render rows update
-        setTimeout(() => {
-          handlers.focusCell(data.key);
-          setTimeout(() => restoreFocus(), 1);
-        }, 1);
+        handlers.focusCell(data.key, false, true);
```
devnaumov, line 212: Did we fixed the old behaviour in different way? 

1. Choose cell
2. Press duplicate row
3. Focus should be on the row
4. We can navigate through editor using keyboard

webapp/common-react/@dbeaver/react-data-grid/src/search/state.ts
Wroud: https://jsbm.dev/fNo9Kkd079pHF

webapp/common-react/@dbeaver/react-data-grid/src/search/useGridSearch.ts
```diff
@@ -0,0 +1,447 @@
...
+    const matches = searchGrid(
+      searchState.query,
+      {
+        caseSensitive: searchState.caseSensitive,
+        wholeWord: searchState.wholeWord,
+        regexp: searchState.regexp,
+      },
+      opts.rowCount,
+      opts.columnCount,
+      (rowIdx, colIdx) => getCellText(opts, rowIdx, colIdx),
```
Wroud, line 150: you have `cellListenersRef`, when you run search you probably want to add temporary list of subscriptions for every `getCellText` used in the search, so you can re-run search or remove changed cells from matches (this is optional behaviour and should be added only if some bugs or issues found with cells change value while searching, it's better to add comment that this way cell changes not tracked)

webapp/common-react/@dbeaver/react-data-grid/src/search/useGridSearch.ts
```diff
@@ -0,0 +1,447 @@
...
+        const st = stateRef.current;
+        if (st.matchedCells.length === 0) {
+          if (st.query) {
+            runSearchNow(false);
+          }
+          return;
+        }
+        const prevIdx = st.activeMatchIdx === 0 ? st.matchedCells.length - 1 : st.activeMatchIdx - 1;
+        dispatch({ type: 'NAVIGATE_PREVIOUS' });
+        scrollToMatch(optionsRef.current.scrollToCell, st.matchedCells[prevIdx]);
```
Wroud: it seems like 
`const nextIdx = (st.activeMatchIdx + 1) % st.matchedCells.length;`
and
`const prevIdx = st.activeMatchIdx === 0 ? st.matchedCells.length - 1 : st.activeMatchIdx - 1;`

should be in sync with `NAVIGATE_NEXT` and `NAVIGATE_PREVIOUS` actions, i will recommend one of:
1. move this logic to the function that will return next or previous Idx based on the state argument and use in in all places
2. find a way to run `scrollToMatch` in reaction to the state change
3. replace `NAVIGATE_NEXT` and `NAVIGATE_PREVIOUS` with `SET_ACTIVE_MATCH` action that will accept match index

webapp/common-react/@dbeaver/react-data-grid/src/useDataGridSearch.ts
```diff
@@ -0,0 +1,95 @@
...
+  function handleSearchClose() {
+    setSearchOpen(false);
+    containerRef.current?.querySelector<HTMLDivElement>('[aria-selected="true"]')?.focus();
+  }
+
+  function onReplace(updates: { rowIdx: number; colIdx: number; value: string }[]) {
+    const validUpdates = updates.filter(u => getCellEditable?.(u.rowIdx, u.colIdx) !== false);
+    if (validUpdates.length > 0) {
+      onCellChangeBatch?.(validUpdates);
+    }
```
Wroud, line 77: i think that in find and replace mode cells that can't be edited should be skipped from search

Wroud (changes requested):

webapp/packages/plugin-data-spreadsheet-new/src/DataGrid/DataGridTable.tsx
SychevAndrey: https://github.com/user-attachments/assets/f2e2ca88-f051-4015-b2e5-dcac926159b8

Yes, it was exactly like you described

webapp/common-react/@dbeaver/react-data-grid/src/useDataGridSearch.ts
SychevAndrey: discussed, we decided to keep showing the cells that we can't update and later make an improvement and show notifications for the user to let them know that some cells require manual updates.

webapp/common-react/@dbeaver/react-data-grid/src/search/useGridSearch.ts
```diff
@@ -433,8 +454,9 @@ export function useGridSearch(options: IGridSearchOptions): IGridSearchResult {
   useEffect(
     () => () => {
       cellListenersRef.current.clear();
+      clearTrackedCellSubscriptions();
     },
-    [],
+    [clearTrackedCellSubscriptions],
   );
```
Wroud, line 459: you have this logic in the `useTrackedGridSearchCellSubscriptions` already so you don't need it here

Wroud (dismissed):

webapp/packages/core-events/src/EventContext/EventContext.ts
```diff
@@ -1,71 +1,9 @@
...
-};
-
-function addContextToEvent(obj: any): IEventWithContext {
-  if (!obj[eventContext]) {
-    obj[eventContext] = [];
-  }
-
-  return obj;
-}
+export { EventContext, type IEventFlag } from '@dbeaver/js-helpers';
```
devnaumov, line 9: Are we sure this is part of the dbeaver package?

devnaumov (dismissed):

sergeyteleshev (dismissed):

webapp/packages/core-events/src/EventContext/EventContext.ts
SychevAndrey: I moved it there according to Wroud's suggestion

sergeyteleshev (approved):

PR #4087, opened 2026-01-26
[CB] remove shortcuts intersection for layouts

sergeyteleshev:
closes https://github.com/dbeaver/pro/issues/8104

I added a scoping feature:
https://react-hotkeys-hook.vercel.app/docs/documentation/useHotkeys/scoping-hotkeys


webapp/packages/core-view/src/View/CaptureView.tsx
```diff
@@ -32,24 +32,20 @@ export const CaptureView = observer<React.PropsWithChildren<ICaptureViewProps>>(
   const viewContext = useViewContext(view, parentContext);
   const actionService = useService(ActionService);
   const activeView = useActiveView(view);
-  const [ref, state] = useFocus<HTMLDivElement>({ onFocus: activeView.focusView, onBlur: activeView.blurView });
+  const [focusRef] = useFocus<HTMLDivElement>({ onFocus: activeView.focusView, onBlur: activeView.blurView });
   const style = useS(styles);
+  const allActionItems = view.actions.map(action => actionService.getAction(viewContext, action)).filter(Boolean) as IActionItem[];
+  const enabledActionItems = allActionItems.filter(action => action.binding && !action.isDisabled());
```
SychevAndrey, line 38: seems like the first filter is redundant (it was the last before and was also redundant)

Here we filter using boolean constructor, but on the very next line we filter again. I'm not sure we saving something by this approach + we allocate 2 arrays instead of one which usually is more expensive then just N checks.

I would suggest combine all the filtering to some predicate function and use it once like


```suggestion
function isActionEnabled (action): action is IActionItem {
    return action !== null && !!action.binding && !action.isDisabled()
  }
  filter(isActionEnabled);
```

webapp/packages/core-view/src/View/CaptureView.tsx
```diff
@@ -32,24 +32,20 @@ export const CaptureView = observer<React.PropsWithChildren<ICaptureViewProps>>(
...
+  const allKeys = allActionItems.filter(action => action.binding).flatMap(item => getCommonAndOSSpecificKeys(item.binding?.binding));
```
SychevAndrey, line 39: why do we need to use the keys of disabled actions now?

Wroud: same question

webapp/packages/core-view/src/View/CaptureView.tsx
```diff
@@ -58,31 +54,33 @@ export const CaptureView = observer<React.PropsWithChildren<ICaptureViewProps>>(
         });
       });
 
+      event.stopImmediatePropagation();
```
Wroud, line 57: why do we need this?

Wroud (changes requested):

webapp/packages/core-view/src/View/CaptureView.tsx
sergeyteleshev: For the fix, we need 2 arrays:
all shortcuts - whole context for the hook
enabled shortcuts - to execute the enabled one during the action handler

Another thing is we have 2 for cycles: `map` and `filter`. They are not nested, so there is no like n^2 time complexity
From a time complexity view, we have 2N, which is equal to N - it is the basis for algorithms and structure courses

webapp/packages/core-view/src/View/CaptureView.tsx
sergeyteleshev: explained in the frontend chat with examples

SychevAndrey (approved):

PR #4102, opened 2026-01-30
dbeaver/pro#8151 fix: reset password and key fields on change detection

SychevAndrey:


webapp/packages/plugin-connections/src/ConnectionForm/SSH/ConnectionFormSSHPart.ts
```diff
@@ -130,6 +130,13 @@ export class ConnectionFormSSHPart extends FormPart<INetworkHandlerConfig, IConn
       handlerConfig = getTrimmedSSHConfig(handlerConfig);
       this.optionsPart.state.networkHandlersConfig!.push(handlerConfig);
     }
+
+    if (passwordChanged) {
+      this.state.password = '';
+    }
+    if (keyChanged) {
+      this.state.key = '';
```
Wroud: isn't it will save empty key and password?

Wroud (dismissed):

webapp/packages/plugin-connections/src/ConnectionForm/SSH/ConnectionFormSSHPart.ts
sergeyteleshev: tried another solution

We had 2 bugs: 
1. `saveChanges` for the `SSH` part did not execute at all because of the incorrect `this.loading`state. We had a race condition when a part was loading and, at the same time, `saveChanges` was called, leading to an early return. So I changed it to actually wait until it's loaded and then tried to save changes
2. Also, backend mutates the state we have sent for security reasons: it erases the password and some other fields. So I added a flag which basically says: "okay, we trust the data from the backend, so put it in the initial and current state when `loader` executed".

devnaumov (approved):

PR #4106, opened 2026-02-02
dbeaver/pro#8119 Add CLI commands for project management (create, rename, delete, list)

labanovichttps:


server/bundles/io.cloudbeaver.model/src/io/cloudbeaver/model/session/WebSessionWorkspace.java
```diff
@@ -143,6 +142,22 @@ public DBPImage getResourceIcon(DBPAdaptable resourceAdapter) {
...
 
+    @NotNull
+    @Override
+    public WebSessionProjectImpl createProject(@NotNull String name, @Nullable String description) throws DBException {
+        throw new DBException("Project creation is not supported in web session workspace");
+    }
+
+    @Override
+    public void deleteProject(@NotNull DBPProject project) throws DBException {
+        removeProject((WebSessionProjectImpl) project);
```
alexander-skoblikov: thow not supported exception, because it's not a real deletion

alexander-skoblikov (changes requested):

PR #4107, opened 2026-02-02
[CB] Ability to clone connections

sergeyteleshev:
closes https://github.com/dbeaver/pro/issues/6998


webapp/packages/plugin-connections/src/ContextMenu/ConnectionMenuBootstrap.ts
```diff
@@ -152,6 +162,27 @@ export class ConnectionMenuBootstrap extends Bootstrap {
...
+            try {
+              const existingConnections = this.connectionInfoResource.values.filter(
+                c => c.projectId === connection.projectId && c.folder === connection.folder,
+              );
+              const newName = getUniqueName(
+                connection.name.concat(` ${COPY_NAME_SUFFIX}`),
+                existingConnections.map(c => c.name),
+              );
+
+              if (!connection.nodePath) {
```
devnaumov, line 177: Its strange that we are relying on the nodePath and its optional, probably we need a new method where we can use connection Id to clone it

devnaumov (dismissed):

HocKu7 (dismissed):

webapp/packages/plugin-connections/src/ContextMenu/ConnectionMenuBootstrap.ts
```diff
@@ -18,16 +19,20 @@ import { NotificationService } from '@cloudbeaver/core-events';
...
 
 import { PublicConnectionFormService } from '../PublicConnectionForm/PublicConnectionFormService.js';
 import { ACTION_CONNECTION_CHANGE_CREDENTIALS } from './Actions/ACTION_CONNECTION_CHANGE_CREDENTIALS.js';
+import { ACTION_CONNECTION_CLONE } from './Actions/ACTION_CONNECTION_CLONE.js';
 import { ACTION_CONNECTION_DISCONNECT } from './Actions/ACTION_CONNECTION_DISCONNECT.js';
 import { ACTION_CONNECTION_DISCONNECT_ALL } from './Actions/ACTION_CONNECTION_DISCONNECT_ALL.js';
 import { ACTION_CONNECTION_EDIT } from './Actions/ACTION_CONNECTION_EDIT.js';
 import { MENU_CONNECTIONS } from './MENU_CONNECTIONS.js';
 
+const COPY_NAME_SUFFIX = 'copy';
```
SychevAndrey: Don't we want to localize it?

webapp/packages/plugin-connections/src/ContextMenu/ConnectionMenuBootstrap.ts
```diff
@@ -152,6 +163,30 @@ export class ConnectionMenuBootstrap extends Bootstrap {
             await this.connectionsManagerService.requireConnection({ connectionId: connection.id, projectId: connection.projectId }, true);
             break;
           }
+          case ACTION_CONNECTION_CLONE: {
+            try {
+              // Load all connections for the project first to ensure we have them all to generate a unique name for the cloned connection
+              await this.connectionInfoResource.load(ConnectionInfoProjectKey(connection.projectId));
```
SychevAndrey: looks heavy for that purpose (we now for sure some customers have thousands of connections). We might want to use different query that will return only names of the connections or we could ask backend to provide an endpoint for name generation.

sergeyteleshev: agreed. seems like we should address this feature in the next ticket due to backend is too busy to help

webapp/packages/plugin-connections/src/ContextMenu/ConnectionMenuBootstrap.ts
sergeyteleshev: We should. I would probably revert that

webapp/packages/plugin-connections/src/ContextMenu/ConnectionMenuBootstrap.ts
```diff
@@ -164,7 +165,10 @@ export class ConnectionMenuBootstrap extends Bootstrap {
           }
           case ACTION_CONNECTION_CLONE: {
             try {
-              const existingConnections = this.connectionInfoResource.values.filter(
+              // Load all connections for the project first to ensure we have them all to generate a unique name for the cloned connection
+              await this.connectionInfoResource.load(ConnectionInfoProjectKey(connection.projectId));
+
+              const existingConnections = this.connectionsManagerService.projectConnections.filter(
```
devnaumov: To be aligned with connection creation logic

```
const connections = await this.connectionInfoResource.load(ConnectionInfoProjectKey(connection.projectId);
const connectionNames = connections.map(connection => connection.name);
```

webapp/packages/plugin-connections/src/ContextMenu/ConnectionMenuBootstrap.ts
```diff
@@ -152,6 +163,30 @@ export class ConnectionMenuBootstrap extends Bootstrap {
...
+          case ACTION_CONNECTION_CLONE: {
+            try {
+              // Load all connections for the project first to ensure we have them all to generate a unique name for the cloned connection
+              await this.connectionInfoResource.load(ConnectionInfoProjectKey(connection.projectId));
+
+              const existingConnections = this.connectionsManagerService.projectConnections.filter(
+                c => c.projectId === connection.projectId && c.folder === connection.folder,
+              );
+              const newName = getUniqueName(
+                connection.name.concat(` ${COPY_NAME_SUFFIX}`),
```
devnaumov: Please use localization service and 'ui_copy' token

devnaumov (changes requested):

webapp/packages/plugin-connections/src/ContextMenu/ConnectionMenuBootstrap.ts
```diff
@@ -152,6 +164,27 @@ export class ConnectionMenuBootstrap extends Bootstrap {
             await this.connectionsManagerService.requireConnection({ connectionId: connection.id, projectId: connection.projectId }, true);
             break;
           }
+          case ACTION_CONNECTION_CLONE: {
+            try {
+              // Load all connections for the project first to ensure we have them all to generate a unique name for the cloned connection
+              const projectConnections = await this.connectionInfoResource.load(ConnectionInfoProjectKey(connection.projectId));
```
Wroud, line 170: This is possibly a heavy operation; maybe we should pass it to the backend side, so we can call the `clone` method without specifying the name, and the backend will choose the right one

devnaumov (dismissed):

webapp/packages/plugin-connections/src/ContextMenu/ConnectionMenuBootstrap.ts
sergeyteleshev: we've talked about it on standup call. backend said this is not a heavy calculation

devnaumov (approved):

PR #4108, opened 2026-02-02
dbeaver/pro#6306 api for original query

yagudin10:


HocKu7 (dismissed):

devnaumov (dismissed):

SychevAndrey (dismissed):

webapp/packages/plugin-data-viewer/src/ContainerDataSource.ts
```diff
@@ -87,9 +87,11 @@ export class ContainerDataSource extends ResultSetDataSource<IDataContainerOptio
 
     try {
       const response = await this.currentTask;
+      const executedQuery = response.fullQuery || '';
 
       this.requestInfo = {
-        originalQuery: response.fullQuery || '',
+        originalQuery: response.originalQuery || executedQuery,
```
devnaumov: I don’t think we should fall back to something else here. The client of this code should receive the original query. If it doesn’t exist, it’s expected to return undefined, and the client can handle any fallback if needed.

devnaumov (approved):

Nexus6v2 (approved):

PR #4117, opened 2026-02-05
[CB] SQL Query generation: front-end part

sergeyteleshev:
closes https://github.com/dbeaver/pro/issues/8048


webapp/packages/plugin-data-spreadsheet-new/src/DataGrid/DataGridContextMenu/DataGridContextMenuGenerateSqlService.ts
```diff
@@ -0,0 +1,268 @@
...
+      root: true,
+      menus: [],
+      contexts: [DATA_CONTEXT_DV_DDM, DATA_CONTEXT_DV_DDM_RESULT_INDEX, DATA_CONTEXT_DV_RESULT_KEY],
+      isApplicable: context => {
+        const model = context.get(DATA_CONTEXT_DV_DDM);
+        const resultIndex = context.get(DATA_CONTEXT_DV_DDM_RESULT_INDEX)!;
+
+        return isResultSetDataModel(model) && !model.source.isReadonly(resultIndex);
+      },
+      getItems: (context, items) => [...items, MENU_DATA_GRID_GENERATE_SQL],
```
SychevAndrey, line 65: why don't we let to generate selects for readony?

webapp/packages/plugin-data-spreadsheet-new/src/DataGrid/DataGridContextMenu/DataGridContextMenuGenerateSqlService.ts
```diff
@@ -0,0 +1,268 @@
...
+          ACTION_DATA_GRID_GENERATE_SQL_SELECT_MANY,
+        ].includes(action),
+      isDisabled: context => {
+        const model = context.get(DATA_CONTEXT_DV_DDM)!;
+
+        if (model.isLoading()) {
+          return true;
+        }
+
+        return hasLobValuesInSelectedRows(context);
```
SychevAndrey: typo

webapp/packages/plugin-data-spreadsheet-new/src/DataGrid/DataGridContextMenu/DataGridContextMenuGenerateSqlService.ts
```diff
@@ -0,0 +1,268 @@
...
+    default:
+      return SqlResultSetGeneratorId.DataInsert;
+  }
+}
+
+function getRowKeys(
+  selectedElements: IGridDataKey<IGridRowKey, IGridColumnKey>[],
+  key: IGridDataKey<IGridRowKey, IGridColumnKey> | undefined,
+): IGridRowKey[] {
+  const rowKeysSet = new Map<string, IGridRowKey>(selectedElements.map(element => [GridDataKeysUtils.serialize(element.row), element.row]));
```
SychevAndrey: named as a set but uses a map

SychevAndrey (changes requested):

webapp/packages/plugin-data-spreadsheet-new/src/DataGrid/DataGridContextMenu/DataGridContextMenuGenerateSqlService.ts
sergeyteleshev: It's not. LOB - Large OBject

webapp/packages/plugin-data-spreadsheet-new/src/DataGrid/DataGridContextMenu/DataGridContextMenuGenerateSqlService.ts
sergeyteleshev: Cause we decided we don't want this feature for the `Grouping panel `and `Session Manager`
We want this feature for: `data editor`, `SQL result sets`, and `datasets`

webapp/packages/plugin-data-spreadsheet-new/src/DataGrid/DataGridContextMenu/DataGridContextMenuGenerateSqlService.ts
```diff
@@ -0,0 +1,268 @@
...
+        data: rowValue,
+        metaData: rowMetadata,
+      });
+    }
+  }
+
+  return result;
+}
+
+function hasLobValuesInSelectedRows(context: IDataContextProvider): boolean {
```
SychevAndrey: if Lob means LargeObject, maybe it's better to explicitly name a function like that because when we quickly check the implementation and see isBinary and isBlobTrancated checks it looks like we are cheching for a blob and they sound very similar

webapp/packages/core-utils/src/executeAsyncSilently.ts
```diff
@@ -0,0 +1,25 @@
...
+ * Licensed under the Apache License, Version 2.0.
+ * you may not use this file except in compliance with the License.
+ */
+
+interface Result<T> {
+  result: T | null;
+  error: Error | null;
+}
+
+export async function executeAsyncSilently<T>(asyncFunction: () => Promise<T>): Promise<Result<T>> {
```
devnaumov: Do we really need this?

We should have one approach on how we handle errors, whether using try/catch of with [] = tryCatch(), our codebase is using try/catch, i think we should stick with it

also its better to return array as result cause you can give different names in the place of use, like this

```
export async function tryCatch<T, E = Error>(promise: T | Promise<T>) {
  try {
    const data = await promise;
    return [data, null] as const;
  } catch (error) {
    return [null, error as E] as const;
  }
}
```

webapp/packages/plugin-data-spreadsheet-new/src/DataGrid/Actions/GenerateSQL/ACTION_DATA_GRID_GENERATE_SQL_DELETE.ts
```diff
@@ -0,0 +1,13 @@
+/*
+ * CloudBeaver - Cloud Database Manager
+ * Copyright (C) 2020-2025 DBeaver Corp and others
```
devnaumov: 2026

devnaumov (changes requested):

webapp/packages/plugin-data-spreadsheet-new/src/DataGrid/DataGridContextMenu/DataGridContextMenuGenerateSqlService.ts
Wroud: this is hard to understand from code, that `model.source.isReadonly(resultIndex)` is about excluding `Grouping panel` and `Session Manager`

webapp/packages/plugin-data-spreadsheet-new/src/DataGrid/DataGridContextMenu/DataGridContextMenuGenerateSqlService.ts
```diff
@@ -0,0 +1,268 @@
...
+    });
+
+    this.menuService.addCreator({
+      menus: [MENU_DATA_GRID_GENERATE_SQL],
+      contexts: [DATA_CONTEXT_DV_DDM, DATA_CONTEXT_DV_DDM_RESULT_INDEX, DATA_CONTEXT_DV_RESULT_KEY],
+      isApplicable: context => {
+        const model = context.get(DATA_CONTEXT_DV_DDM);
+        const resultIndex = context.get(DATA_CONTEXT_DV_DDM_RESULT_INDEX)!;
+        return isResultSetDataModel(model) && !model.source.isReadonly(resultIndex);
+      },
```
Wroud: you don't need to duplicate this check if it's required for submenu itself

webapp/packages/plugin-data-spreadsheet-new/src/DataGrid/DataGridContextMenu/DataGridContextMenuGenerateSqlService.ts
```diff
@@ -0,0 +1,268 @@
...
+      menus: [MENU_DATA_GRID_GENERATE_SQL],
+      contexts: [DATA_CONTEXT_DV_DDM, DATA_CONTEXT_DV_DDM_RESULT_INDEX, DATA_CONTEXT_DV_RESULT_KEY],
+      isActionApplicable: (context, action) =>
+        [
+          ACTION_DATA_GRID_GENERATE_SQL_INSERT,
+          ACTION_DATA_GRID_GENERATE_SQL_UPDATE,
+          ACTION_DATA_GRID_GENERATE_SQL_DELETE,
+          ACTION_DATA_GRID_GENERATE_SQL_SELECT,
+          ACTION_DATA_GRID_GENERATE_SQL_SELECT_MANY,
+        ].includes(action),
```
Wroud: just specify:
```
actions:  [
          ACTION_DATA_GRID_GENERATE_SQL_INSERT,
          ACTION_DATA_GRID_GENERATE_SQL_UPDATE,
          ACTION_DATA_GRID_GENERATE_SQL_DELETE,
          ACTION_DATA_GRID_GENERATE_SQL_SELECT,
          ACTION_DATA_GRID_GENERATE_SQL_SELECT_MANY,
        ]
```

webapp/packages/core-utils/src/executeAsyncSilently.ts
Wroud: i don't think that we need this helper at all

webapp/packages/plugin-data-spreadsheet-new/src/DataGrid/DataGridContextMenu/DataGridContextMenuGenerateSqlService.ts
```diff
@@ -0,0 +1,268 @@
...
+        resultsId: resultId,
+        generatorId,
+        selectedRows: rows,
+      }),
+    );
+
+    await this.commonDialogService.open(GeneratedSqlDialog, {
+      query: result ?? '',
+      exception: error,
+      nodeId: connectionId,
```
Wroud: why we want to open dialog? why we can't show error notification instead?

webapp/packages/plugin-data-spreadsheet-new/src/DataGrid/DataGridContextMenu/DataGridContextMenuGenerateSqlService.ts
Wroud: the main purpose of the `IAction` is to provide additional behaviour such as shortcuts or reusability in different contexts, `IAction` is some of kind standarterized action in the application that can be used in different contexts (menus, menu bars), you don't need to create action for every item in the menu, sometimes it's more convenient to use `MenuBaseItem`

webapp/packages/plugin-data-spreadsheet-new/src/DataGrid/DataGridContextMenu/DataGridContextMenuGenerateSqlService.ts
devnaumov: We can use helpers such as isDatabaseSessionManagerDataSource that is already exist to exclude specific data sources

sergeyteleshev: this is what we already do:
https://github.com/dbeaver/cloudbeaver/blob/b180f971478c4aa49fcc4b1bbe1375f642133f63/webapp/packages/plugin-data-spreadsheet-new/src/DataGrid/DataGridContextMenu/DataGridContextMenuCellEditingService.ts#L67

Wroud (dismissed):

SychevAndrey (dismissed):

PR #4134, opened 2026-02-12
ctrl + z is unstable (lib artefact fix)

### LINKED ISSUE #3815: ctrl + z is unstable
### Description

### Steps to reproduce

### Expected/Desired Behavior

### CloudBeaver Version

25.2.2

### Additional context

_No response_

sergeyteleshev:

The bug is on the lib side. The issue here is that it fires the event handler on almost every key press because it does not account for modifiers (Ctrl, Cmd, Alt, etc.).

What I've done here:
1. fixed bug and created issue & PR for the library https://github.com/JohannesKlauss/react-hotkeys-hook/issues/1318
2. Build the lib and these artefacts put them into our codebase
3. Then, I implemented the fix for the task. Added `useKey:true` so it parses shortcuts not by the code but for the symbol itself. So, for example, the symbol "s" is the same on any type of keyboard (QWERTY, AZERTY, etc.).

As a third alternative solution, we can just wait for the lib to fix the issue. Then update the lib


webapp/common-react/@dbeaver/react-hotkeys/artifacts/lib/BoundHotkeysProxyProvider.d.ts
```diff
@@ -0,0 +1,14 @@
+import { ReactNode } from 'react';
```
devnaumov, line 1: Please revert. Modifying a library’s internal code should be a last resort. We need to be absolutely certain the problem originates in the library. To confirm that, we should open an issue with a question, wait for the maintainer’s response, and then we can make a decision.

I’m not convinced the problem is on the library side, it seems they are aware of the layout differences and already support them.

https://react-hotkeys-hook.vercel.app/docs/documentation/useHotkeys/ignore-layouts

devnaumov (changes requested):

webapp/common-react/@dbeaver/react-hotkeys/artifacts/lib/BoundHotkeysProxyProvider.d.ts
sergeyteleshev: > I’m not convinced the problem is on the library side

This is definitely a library side issue. All modifiers (ctrl, shift, cmd, etc.) DON'T WORK at all. So you are unavailable to do something like `ctrl+shift+s` or `ctrl+S` (as an alternative with `useKey: true` flag)

I'm not the only one who thinks like that:
https://github.com/JohannesKlauss/react-hotkeys-hook/issues/1318
https://github.com/JohannesKlauss/react-hotkeys-hook/issues/1316
https://github.com/JohannesKlauss/react-hotkeys-hook/issues/1300

Moreover, there are recent changes in this area according to `git blame`. I've checked, and previously it worked before the refactor they did

PR #4135, opened 2026-02-12
ctrl + z is unstable (alternative)

### LINKED ISSUE #3815: ctrl + z is unstable

sergeyteleshev:

### DIFF
--- a/webapp/packages/core-view/src/View/CaptureView.tsx
+++ b/webapp/packages/core-view/src/View/CaptureView.tsx
@@ -20,6 +20,7 @@ import styles from './CaptureView.module.css';
 import { CaptureViewContext } from './CaptureViewContext.js';
 import type { IView } from './IView.js';
 import { parseHotkey } from './parseHotkey.js';
+import { matchesHotkeyModifiers } from './matchesHotkeyModifiers.js';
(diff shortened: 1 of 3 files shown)


webapp/packages/core-view/src/View/CaptureView.tsx
```diff
@@ -57,14 +58,16 @@ export const CaptureView = observer<React.PropsWithChildren<ICaptureViewProps>>(
         return commonAndSpecificKeys.some(key => {
           const hotkey = parseHotkey(key);
 
-          return isObjectsEqual(hotkey, handler);
+          // modifiers are not checked by the library cause it has a bug: https://github.com/JohannesKlauss/react-hotkeys-hook/issues/1318
+          return isObjectsEqual(hotkey, handler) && matchesHotkeyModifiers(event, hotkey);
```
sergeyteleshev, line 62: The bug is on the lib side. The issue here is that it fires the event handler on almost every key press because it does not consider modifiers (Ctrl, Cmd, Alt, etc.).

SychevAndrey (approved):

PR #4140, opened 2026-02-13
dbeaver/pro#8068 [CB] Fix: form reload on save and AI settings language

SychevAndrey:
Don't autoselect on Enter if allowCustomValue


webapp/packages/core-ui/src/Form/FormPart.ts
```diff
@@ -113,8 +113,15 @@ export abstract class FormPart<TPartState extends object, TFormState = any> impl
...
+      // loaded=false ensures setInitialState resets both initialState and state.
       this.loaded = false;
-      this.exception = null;
+
+      try {
+        await this.loader();
+        this.exception = null;
+      } finally {
+        this.loaded = true;
+      }
```
Wroud: This should not be here, form will be automatically loaded after saving in another process. Loading is not a part of saving so loading error should not be handled in this place. And loading should not be triggered here

Wroud (changes requested):

Wroud:
You described a bug with previous fix with no proper observable, this can be fixed by modifying one of the loading state interface flags (you can set `isOutdated` to `false` and `isLoaded` to `true`) there is already mechanism for such cases, so you need to fix a bug, not modify behavior. In some cases there is no need to load data after saving if form closed or data was modified imperatively.

webapp/packages/core-ui/src/Form/FormPart.ts
```diff
@@ -113,7 +113,7 @@ export abstract class FormPart<TPartState extends object, TFormState = any> impl
         return;
       }
 
-      this.loaded = false;
+      this.setInitialState(toJS(this.state));
```
Wroud: this is very bad idea, because backend might return different value for state

Wroud (changes requested):
I will suggest the next solution:
in the `ILoadableState` interface:
1. add optional field `isLoadable` (you might choose a different name, but keep in mind that `CachedResource` has this field already, so if you will use different name, it's better to rename `isLodable` in the `CachedResource` to that name)

in the `useAutoLoad`:
1. In `getComputed`, add a  call to the `isLoadable` to trigger re-render when this flag changes
2. In the `loadFunctionName` function, skip loading the  call if `isLoadable?.() === false`

in the `FormPart` class:
1. Implement the `isLoadable` function `return !this.formState.savingPromise`
2. Use this new function in the `load` function instead of `this.formState.savingPromise`

motivation:
1. There might be many parts with heavy data in a single form, we don't want to load all this data until the user switches to the specified part
2. We can't just set the current `state` to the `initialState` because there are sometimes bugs on the backend side or concurrent collisions resulting in different frontend and backend state

webapp/packages/core-blocks/src/Loader/useAutoLoad.ts
```diff
@@ -31,7 +31,7 @@ export function useAutoLoad(
   for (const loader of state as ReadonlyArray<ILoadableState>) {
     getComputed(
       // activate mobx subscriptions
-      () => (!loader.isLoaded() || loader.isOutdated?.() === true) && !loader.isError(),
+      () => (!loader.isLoaded() || loader.isOutdated?.() === true) && !loader.isError() && loader.isLoadable?.() !== false,
```
devnaumov, line 34: `loader.isLoadable?.() === true` ?
as !== false will include undefined

Wroud: no, thats fine, because isLoadable === undefined is also means that resource can be loaded (it's not limited by this flag)

Wroud (approved):

webapp/packages/core-blocks/src/Loader/useAutoLoad.ts
sergeyteleshev: thats a bit confusing I agree
probably if `isLoadable` was not optional in the interface, and would be defined as `isLoadable() {return true};` with possibility to override the method, it will make more sense

sergeyteleshev (approved):

PR #4141, opened 2026-02-16
dbeaver/pro#8196 validate projects on copying connection

yagudin10:


server/bundles/io.cloudbeaver.server/src/io/cloudbeaver/service/core/impl/WebServiceCore.java
```diff
@@ -464,6 +464,11 @@ public WebConnectionInfo copyConnectionFromNode(
                 throw new DBException("Node '" + nodePath + "' is not a datasource node");
             }
             DBPDataSourceContainer dataSourceTemplate = dbnDataSource.getDataSourceContainer();
+            if (!dataSourceTemplate.isExternallyProvided()) {
```
Nexus6v2, line 467: I think you can join conditions here. Also, maybe put it in separate named variables for readability

PR #4144, opened 2026-02-16
dbeaver/pro#3822 make selection accountable

### LINKED ISSUE #3822: alt + x is bugged and does not behave like in dbeaver
### Description

Hello, there are several issues with the alt+x shortcut.

1) When you open a new sql editor containing queries and play alt +x, the queries are executed. If you run alt+x again, the queries are executed AND cloudbeaver asks if we want to close the tabs. It should just play the queries.

2) alt+x does not behave the same way in dbeaver. In dbeaver, it only executes the queries that have been highlighted. There it plays all queries from the tab.(it can be annoying in production when you're used to dbeaver :/) You would either have to make the behavior of alt+x identical to that of dbeaver, or add a new shortcut allowing only highlighted queries to be played.

Thank you !

### Steps to reproduce

1) Open a new sql editor
2) Write queries
3) alt + x
4) alt + x

### Expected/Desired Behavior

1) No bug with alt + x (played more than one time)
2) only run highlighted queries

### CloudBeaver Version

25.2.2

### Additional context

_No response_

devnaumov:
closes [3822](https://github.com/dbeaver/cloudbeaver/issues/3822)


webapp/packages/plugin-sql-editor/src/SqlEditor/useSqlEditor.ts
```diff
@@ -241,10 +241,14 @@ export function useSqlEditor(state: ISqlEditorTabState): ISQLEditorData {
...
         this.onExecute.execute(true);
+
         try {
           this.executingScript = true;
-          await this.model.getResolvedSegment();
-          const queries = this.model.parser.scripts;
+
+          const segment = await this.model.getResolvedSegment();
+
+          const hasSelection = segment && this.model.cursor.anchor !== this.model.cursor.head;
```
Wroud: you don't need segment, you only need cursor

Wroud (changes requested):

Wroud (dismissed):

Wroud (approved):

PR #4164, opened 2026-02-23
dbeaver/pro#8298 fix: preserve query param order, fix incorrect repla…

SychevAndrey:
…cement, add tooltip


webapp/packages/core-blocks/src/PropertiesTable/PropertiesTable.tsx
```diff
@@ -36,6 +36,7 @@ interface Props {
   className?: string;
   staticProperties?: boolean;
   filterable?: boolean;
+  ordered?: boolean;
```
Wroud: `orderByName` / `sortByName` and it should be enabled by default

webapp/packages/plugin-sql-editor/src/QueryDataSource.ts
```diff
@@ -300,12 +300,22 @@ export class QueryDataSource<TOptions extends IDataQueryOptions = IDataQueryOpti
...
+
+    for (const parameter of queryParamsEvent.parameters) {
+      if (!parameter?.name || parameterValues.has(parameter.name)) {
+        continue;
+      }
+
+      parameterValues.set(parameter.name, parameter.value);
+    }
+
+    const paramNames = Array.from(parameterValues.keys());
```
Wroud, line 313: i think you can just use `queryParamsEvent.parameters` as is, without `parameterValues` map

SychevAndrey: I can't, because the params array has duplicates like:
```parameters: [{name: "b"}, {name: "col2"}, {name: "a"}, {name: "col2"}]```

Wroud (approved):

sergeyteleshev (approved):

PR #4165, opened 2026-02-23
dbeaver/cloudbeaver#3815 upgrades lib to version where it works

### LINKED ISSUE #3815: ctrl + z is unstable

sergeyteleshev:


SychevAndrey (dismissed):

webapp/packages/core-blocks/src/useHotkeys.ts
```diff
@@ -1,54 +1,108 @@
...
 declare module 'react-hotkeys-hook' {
-  export type FormTags = 'input' | 'textarea' | 'select' | 'INPUT' | 'TEXTAREA' | 'SELECT';
+  export type FormTags =
+    | 'input'
+    | 'textarea'
+    | 'select'
+    | 'INPUT'
+    | 'TEXTAREA'
+    | 'SELECT'
+    | 'searchbox'
```
devnaumov, line 21: Why we extend the initial form tags here? slider and, spinbutton and menuitem looks strange here

devnaumov (dismissed):

webapp/packages/core-blocks/src/useHotkeys.ts
sergeyteleshev: it Is copy pasted types from the lib itself. added it for consistency

SychevAndrey (dismissed):

devnaumov (dismissed):

webapp/packages/core-view/src/View/parseHotkey.ts
```diff
@@ -82,10 +82,9 @@ export function isBindingPressed<T extends HTMLElement>(event: KeyboardEvent<T>,
 }
 
 export function parseHotkey(hotkey: string, combinationKey = '+'): Hotkey {
-  const keys = hotkey
-    .toLocaleLowerCase()
-    .split(combinationKey)
-    .map(k => mapKey(k));
+  hotkey = hotkey.toLocaleLowerCase();
```
devnaumov, line 85: Dont understand why do we need these changes. Logic seems to be the same, also its better to not mutate fn args and create local variable instead

sergeyteleshev: because the lib now transforms it inside to lowercase
and basically we compare it like this: "Alt+R" (ours) === "alt+r" (libs)

its okay to mutate if it is not an object. nothing is gonna happen. primitives don't have a links as an objects

devnaumov (approved):

PR #4171, opened 2026-02-24
Confusing "something went wrong" in case of network issues

sergeyteleshev:
closes https://github.com/dbeaver/pro/issues/8468


webapp/packages/core-events/src/INotification.ts
```diff
@@ -51,6 +51,7 @@ export interface INotification<TProps extends INotificationExtraProps<any> = INo
   type: ENotificationType;
   title: string;
   message?: string;
+  pinned?: boolean;
```
Wroud, line 54: how many cases do we have with pinned notifications? do we really need to extend it?

webapp/packages/core-events/src/NotificationService.ts
```diff
@@ -182,15 +180,15 @@ export class NotificationService {
...
+  logException(exception: Error | GQLError | undefined | null, title: string, message?: string, silent?: boolean): void {
     const errorDetails = errorOf(exception, DetailsError);
 
     if (!silent) {
       const hasDetails = errorDetails?.hasDetails() ?? false;
 
       if (hasDetails) {
         if (!title) {
-          title = exception?.name;
+          title = exception?.name || 'ui_unexpected_error';
```
Wroud, line 191: You've changed behaviour:
Previously if the developer hasn't specified a title for the notification, this code will try to get the title from the message (especially for some errors returned by the backend, we want to show this message)

But right now, `title` is required, and `if (!title) {` will never be the case (except you will pass an empty string, but it's strange), and this mechanism is broken

I believe you should revert these changes:
1. We don't want to specify the notification title for every error notification by default (by default, we want to showthe  error from the backend as-is)
2. We still want to be able to override the notification title in some cases

webapp/packages/core-root/src/ServerHealthCheckService.ts
```diff
@@ -0,0 +1,129 @@
...
+const HEALTH_CHECK_ABORT_TIMEOUT = 2000;
+
+export enum ServerHealthStatus {
+  Unknown = 'unknown',
+  Alive = 'alive',
+  Unavailable = 'unavailable',
+}
+
+@injectable(() => [GraphQLService])
+export class ServerHealthCheckService extends Bootstrap {
```
Wroud: Do we really need all this to show a better error message for the user? I don't think so, the initial problem is: "I see 'something went wrong' instead of 'network error' or 'server is unavailable'". I believe that you can achieve this with fewer changes in the codebase without a new mechanism for tracking server availability

Wroud (changes requested):

PR #4172, opened 2026-02-25
Confusing "something went wrong" in case of network issues

sergeyteleshev:
closes https://github.com/dbeaver/pro/issues/8468

The case when the server died:
https://github.com/user-attachments/assets/096e7def-732e-45e4-b722-e24fda5cdadb

The case when timeout triggers (for example, VPN is turned off):

Screenshot 2026-02-27 at 13 04 31

https://github.com/user-attachments/assets/8e5d54f4-5202-4af5-94d8-783a9fa91350


webapp/packages/core-root/src/NetworkStateService.ts
```diff
@@ -54,7 +58,7 @@ export class NetworkStateService extends Bootstrap {
         this.graphQLService.enableRequests();
       }
     } else {
-      this.graphQLService.blockRequests(new NetworkError('Network connection was lost'));
+      this.graphQLService.blockRequests(new NetworkError(this.localizationService.translate('core_root_network_connection_lost_error')));
```
Wroud: please avoid using localization in the base app initialization logic, because this errors might appear before localization initialized, so better to keep unlocalized messages in such cases

webapp/packages/core-blocks/src/ExceptionMessage.tsx
```diff
@@ -56,7 +59,6 @@ export const ExceptionMessage = observer<Props>(function ExceptionMessage({ exce
             </h2>
             <div className={s(styles, { errorMessage: true })}>
               {(error.hasDetails && error.message) || translate('core_blocks_exception_message_error_message')}{' '}
-              {onRetry && translate('ui_please_retry')}
```
devnaumov, line 59: Can you explain please why are we removing it?

webapp/packages/core-root/src/NetworkStateService.ts
```diff
@@ -64,9 +64,18 @@ export class NetworkStateService extends Bootstrap {
     try {
       return await request;
     } catch (exception: any) {
+      const gqlError = errorOf(exception, GQLError) ?? errorOf(exception, PlainGQLError);
```
devnaumov: Its a little bit strange that we add this logic inside sessionExpiredInterceptor, i think it would be difficult to find. Also we have a ServerInternalError, that seems perfect place for status code >= 500. Probably we should rename it?

webapp/packages/core-blocks/src/ExceptionMessage.tsx
sergeyteleshev: 2 reasons:
1. The button "retry" already calls to action without extra text
2. With the text I just added, it seems like duplication. See the video and imagine this text at the end of the sentence

devnaumov (dismissed):

devnaumov (dismissed):

SychevAndrey (dismissed):

SychevAndrey (dismissed):

webapp/packages/core-root/src/NetworkStateService.ts
```diff
@@ -60,13 +62,27 @@ export class NetworkStateService extends Bootstrap {
...
+      const isGqlProxyError = gqlError?.response.status === 500 && (!gqlError.response.body || Object.keys(gqlError.response.body).length === 0);
+      /* 
+        fetch() API throws a TypeError when a network error occurs, and the message is 'Failed to fetch'. 
+        This is a common way to detect network issues in web applications in prod mode (not dev frontend build) 
+      */
+      const isCommonNetworkError = exception instanceof TypeError && exception.message === NO_NETWORK_MESSAGE;
+
+      if (isCommonNetworkError || isGqlProxyError) {
+        throw new NetworkError('Server is not available. Please check your network connection and try again.', { cause: exception });
```
devnaumov, line 68: Why we use only one message for both cases now?  I think we should keep the old if block too 

```
if (exception instanceof TypeError && exception.message === 'Failed to fetch') {
        throw new NetworkError('Error while processing request', { cause: exception });
}
```

Wroud: I think that's okay, because the initial request was about "I don't understand this message" in same case

Wroud (dismissed):

webapp/packages/core-root/src/NetworkStateService.ts
Wroud: the `ServerInternalError` is for different case

devnaumov (dismissed):

SychevAndrey (approved):

PR #4174, opened 2026-02-25
dbeaver/pro#7657 avoid blicking by debounce loading state

devnaumov:
closes [7657](https://github.com/dbeaver/pro/issues/7657)

delay showing
instant hiding


webapp/packages/core-blocks/src/useDebounceCallback.ts
```diff
@@ -0,0 +1,51 @@
...
+
+/* eslint-disable react-hooks/refs */
+
+import { useMemo, useRef } from 'react';
+
+type TDebouncedCallback<Params extends unknown[]> = ((...args: Params) => void) & {
+  cancel: () => void;
+};
+
+export function useDebounceCallback<Params extends unknown[], Return>(
```
Wroud: There is also `useStateDelay` and `useActivationDelay` with similar logic

webapp/packages/core-blocks/src/Loader/Loader.tsx
```diff
@@ -162,24 +164,26 @@ export const Loader = observer<Props>(function Loader({
   }
 
   const style = useS(styles);
-  const [isVisible, setVisible] = useState(loading);
+
+  const debouncedLoading = useDebounceValue(loading, loading === true ? 500 : 0);
+  const [isVisible, setVisible] = useState(debouncedLoading);
```
Wroud: this is debounce over debounce (`setVisible` called in the `useEffect` with a timeout to achieve debounce behaviour)

So this is kinda strange

Wroud:
in general, if loader mounted with loading state it should immediately display loading state, if loader changed state to the loading state it should delay it

webapp/packages/core-blocks/src/Loader/Loader.tsx
devnaumov: According to the code, the visible only affect loader when overlay is true, so it does not debounce general loading state (verified)
CleanShot 2026-02-26 at 12 11 33@2x

devnaumov:
> in general, if loader mounted with loading state it should immediately display loading state, if loader changed state to the loading state it should delay it

yes, it works that way now

webapp/packages/core-blocks/src/useDebounceCallback.ts
devnaumov: The naming for such hooks in react ecosystem is useDebounceValue useDebounceCallback so i think we should stick with it. Yes, logic is similar in these hooks, I will check them

Wroud: yep, it's better not to add more similar hooks, we can replace old ones

webapp/packages/core-blocks/src/Loader/Loader.tsx
Wroud: if so, maybe we don't need `isVisible` anymore if it's also debounced? (you can keep only debounced loading state and use instead of `isVisible`)

webapp/packages/core-blocks/src/useDebounceCallback.ts
```diff
@@ -0,0 +1,51 @@
...
+  callback: (...args: Params) => Return,
+  delay: number,
+): TDebouncedCallback<Params> {
+  const internalCallbackRef = useRef(callback);
+  const timerRef = useRef<ReturnType<typeof setTimeout> | null>(null);
+  const delayRef = useRef(delay);
+
+  internalCallbackRef.current = callback;
+  delayRef.current = delay;
```
SychevAndrey: we should not update refs during render, as we have discussed that before. You can read or write refs from event handlers or effects instead.

SychevAndrey (changes requested):

webapp/packages/core-blocks/src/useDebounceCallback.ts
devnaumov: Changed to useStateDelay

webapp/packages/core-blocks/src/Loader/Loader.tsx
```diff
@@ -237,24 +225,24 @@ export const Loader = observer<Props>(function Loader({
...
     }
 
-    if (!loading) {
+    if (!debouncedLoading) {
       return null;
     }
   }
 
-  if ((!isVisible && overlay) || !loading) {
+  if (!debouncedLoading) {
```
Wroud: do we sure about this changes? why not `(!debouncedLoading && overlay) || !loading`

Wroud (approved):

sergeyteleshev (approved):

SychevAndrey (approved):

PR #4184, opened 2026-03-02
dbeaver/pro#7566 feat: add settings functionality to the new tree

SychevAndrey:
Implemented settings as a part of Toolbar. The key reason was that the TreeToolbar sets different contexts for all other actions related to tree and easily extendible as a menu service. It might limit in some weird cases, but for now it looks like a valid decision

Settings are made as unknown intentionally to be polymorphic. We could use Record<string, unknown> instead, but I don't see it really improving anything.


webapp/packages/plugin-navigation-tree/src/TreeNew/DATA_CONTEXT_TREE.ts
```diff
@@ -16,3 +16,4 @@ export type TreeRefreshHandler = () => Promise<void>;
 export const DATA_CONTEXT_TREE_DATA = createDataContext<ITreeData>('tree-new-data');
 export const DATA_CONTEXT_TREE_FILTER = createDataContext<ITreeToolbarFilter>('tree-new-filter');
 export const DATA_CONTEXT_TREE_REFRESH = createDataContext<TreeRefreshHandler>('tree-new-refresh');
+export const DATA_CONTEXT_TREE_SETTINGS = createDataContext<unknown>('tree-new-settings');
```
Wroud: it's better to put some interface:
```
interface ITreeSettings {
  get<T>(key: string): T | undefined;
  set<T>(key: string, value: T): void;
}
```

Wroud (changes requested):

webapp/packages/plugin-navigation-tree/src/TreeNew/DATA_CONTEXT_TREE.ts
sergeyteleshev: more like: `Record<string, unknown> | undefined`

sergeyteleshev (dismissed):

webapp/packages/plugin-navigation-tree/src/TreeNew/useTreeFilter.ts
```diff
@@ -116,3 +147,51 @@ export function useTreeFilter(options: ITreeFilterOptions = {}): Readonly<ITreeF
     ['setFilter', 'isNodeMatched', 'transformer', 'stateTransformer'],
   );
 }
+
+export function useTreeFilter(
```
Wroud: this hook seems to be useless

webapp/packages/plugin-navigation-tree/src/TreeNew/useTreeFilter.ts
```diff
@@ -25,7 +27,36 @@ export interface ITreeFilter {
...
+  childrenTransformers: Array<TreeDataTransformer<string[]>>;
+  stateTransformers: Array<TreeDataTransformer<INodeState>>;
+  toolbarFilter: ITreeFilter | undefined;
+}
+
+export type ITreeFilterWithState = Readonly<ITreeFilter & ITreeFilterState>;
+
+export const TREE_SETTINGS_FILTER_ENABLED = 'tree.filter.enabled';
+
+export function useTreeFilterState(
```
Wroud: this hook also seems to be useless

webapp/packages/plugin-navigation-tree/src/TreeNew/useTreeFilter.ts
```diff
@@ -25,7 +27,36 @@ export interface ITreeFilter {
...
+    enabled,
+    childrenTransformers,
+    stateTransformers,
+    toolbarFilter: enabled ? treeFilter : undefined,
+  };
+}
+
+function useTreeFilterBase(options: ITreeFilterOptions = {}): Readonly<ITreeFilter> {
   options = useObjectRef(options);
   const matchCache = new Map<string, boolean>();
```
Wroud, line 46: you can get enabled state here in one line:
```
const enabled = settings?.get<boolean>(TREE_SETTINGS_FILTER_ENABLED) ?? true;
```
and use it in the `transformer` and `stateTransformer` to ignore logic if filter not enabled

i will also recommend to not add `enabledKey` variable for now and only introduce it if it will be needed

webapp/packages/plugin-navigation-tree/src/TreeNew/useTreeSettings.ts
```diff
@@ -0,0 +1,45 @@
...
+                return this.settings.get(key) as T | undefined;
+            },
+            set<T>(key: string, value: T): void {
+                this.settings.set(key, value);
+                this.onSet?.(key, value);
+            },
+        }),
+        {
+            onSet: observable.ref,
+            settings: observable.ref,
```
Wroud: actually you can remove both lines, because `onSet` will not impact "read" data logic so you don't care if it changed
and `settings` initialized in constructor and already observable because of `observable.map`

Wroud (changes requested):

webapp/packages/plugin-navigation-tree/src/TreeNew/useTreeSettings.ts
```diff
@@ -0,0 +1,45 @@
...
+import { useObservableRef } from '@cloudbeaver/core-blocks';
+
+export interface ITreeSettings {
+    get<T>(key: string): T | undefined;
+    set<T>(key: string, value: T): void;
+}
+
+export interface ITreeSettingsOptions {
+    initialSettings?: Record<string, unknown>;
+    onSet?<T>(key: string, value: T): void;
```
Wroud: i think it's better to change `onSet` to `onChange` and pass `entities` as an argument, so it will be easier to handle key deletion and `clear` actions on settings map

Wroud (changes requested):

sergeyteleshev (dismissed):

webapp/packages/plugin-navigation-tree/src/TreeNew/useTreeFilter.ts
```diff
@@ -57,9 +76,22 @@ export function useTreeFilter(options: ITreeFilterOptions = {}): Readonly<ITreeF
...
+      settings,
       filter: '',
+      get enabled() {
+        return this.settings?.get<boolean>(TREE_SETTINGS_FILTER_ENABLED) ?? true;
+      },
+      get childrenTransformers() {
+        return this.enabled ? [this.transformer] : [];
+      },
+      get stateTransformers() {
+        return this.enabled ? [this.stateTransformer] : [];
```
Wroud: i don't think that you need this

Wroud (changes requested):

sergeyteleshev (dismissed):

webapp/packages/plugin-navigation-tree/src/TreeNew/useTreeFilter.ts
```diff
@@ -57,9 +74,16 @@ export function useTreeFilter(options: ITreeFilterOptions = {}): Readonly<ITreeF
...
+  return useObservableRef<ITreeFilterStateObject>(
     () => ({
+      settings,
       filter: '',
+      get enabled() {
+        return this.settings?.get<boolean>(TREE_SETTINGS_FILTER_ENABLED) ?? true;
+      },
+      get toolbarFilter(): ITreeFilter | undefined {
+        return this.enabled ? this : undefined;
+      },
```
Wroud: This is very strange, because you can access all fields from `ITreeFilter` from `ITreeFilterWithState`, but you have `toolbarFilter` that can be undefined, the developer will just use methods from `ITreeFilterWithState` because they are always presented, so he doesn't need to check for undefined

(i will recommend to remove this field)

webapp/packages/plugin-navigation-tree/src/TreeNew/useUserTreeSettings.ts
```diff
@@ -0,0 +1,41 @@
...
+        if (!newSettings.has(key)) {
+          delete persistedSettings[key];
+        }
+      }
+
+      for (const [key, value] of newSettings) {
+        persistedSettings[key] = value;
+      }
+    },
+  });
```
Wroud: You should handle the case when the user re-logins, in this case, `persistedSettings` will be changed, while `useTreeSettings` will remain the same, see `useUserData` usages

Wroud (changes requested):

Wroud (dismissed):

sergeyteleshev (dismissed):

sergeyteleshev (approved):

PR #4194, opened 2026-03-05
[CB] Error in the console when opening the admin panel

sergeyteleshev:
closes https://github.com/dbeaver/pro/issues/5495


webapp/packages/plugin-administration/src/ConfigurationWizard/ServerConfiguration/ServerConfigurationFormStateManager.ts
```diff
@@ -28,7 +28,7 @@ export class ServerConfigurationFormStateManager {
     });
   }
 
-  create(): IFormState<null> {
+  open(): IFormState<null> {
```
devnaumov: We are not opening anything here, we are creating formState instance here, why change name to open?
Usually you need to create separate method in service, name it open and call create from stateManager.create from there. Manager is just for state, it does not know anything about panels and etc so cant have open method

webapp/packages/plugin-administration/src/ConfigurationWizard/ServerConfiguration/ServerConfigurationPage.tsx
```diff
@@ -53,7 +53,7 @@ export const ServerConfigurationPage: AdministrationItemContentComponent = obser
   const serverConfigurationFormStateManager = useService(ServerConfigurationFormStateManager);
   const configurationWizardService = useService(ConfigurationWizardService);
 
-  const formState = serverConfigurationFormStateManager.formState!;
+  const formState = serverConfigurationFormStateManager.open();
```
Wroud: i think you need to remove `!` and change logic to respect that fact that formState might be not initialized, you can do it by moving all dependent logic into different component that will accept `formState` as a prop and in this component you will display some message instead of form

Wroud (changes requested):

webapp/packages/plugin-administration/src/ConfigurationWizard/ServerConfiguration/ServerConfigurationPage.tsx
sergeyteleshev: Okay!

image

SychevAndrey (approved):

Wroud (approved):

devnaumov (approved):

PR #4207, opened 2026-03-10
[CB] refactor data editor features readonly states

sergeyteleshev:
closes https://github.com/dbeaver/pro/issues/8299

introduces new abstraction editable model/source in order to simplify understanding how ui enabling/hiding works for different models and sources

### DIFF
--- a/webapp/packages/plugin-data-viewer/src/ResultSet/isEditableDataSource.ts
+++ b/webapp/packages/plugin-data-viewer/src/ResultSet/isEditableDataSource.ts
@@ -1,0 +1,29 @@
+/*
+ * CloudBeaver - Cloud Database Manager
+ * Copyright (C) 2020-2026 DBeaver Corp and others
+ *
+ * Licensed under the Apache License, Version 2.0.
+ * you may not use this file except in compliance with the License.
+ */
+import { type IDatabaseDataSource } from '../DatabaseDataModel/IDatabaseDataSource.js';
+import { type ResultSetDataSource, isResultSetDataSource } from './ResultSetDataSource.js';
+
+/**
+ * Type guard that checks if a data source supports editing operations at the specified result index.
+ * Returns true for data sources that inherently support editing (e.g., ContainerDataSource, QueryDataSource,
+ * DatasetDataSource, SubCollectionDataSource) and are not marked as readonly for the given result.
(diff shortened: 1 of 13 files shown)


webapp/packages/plugin-data-viewer/src/ValuePanelPresentation/TextValue/isTextValueReadonly.ts
```diff
@@ -23,15 +24,13 @@ interface Args {
...
+export function isTextValueReadonly({ contentAction, formatAction, model, resultIndex, cellHolder, editAction }: Args): boolean {
   if (!cellHolder) {
     return true;
   }
 
   return (
-    // TODO add more proper way to define to what features it should be added https://github.com/dbeaver/pro/issues/8299
-    model.isReadonly(resultIndex) ||
-    model.isDisabled(resultIndex) ||
+    !isEditableDataSource(model.source, resultIndex) ||
```
devnaumov: Previously it was isReadonly || isDisabled, seems like now isEditableDataSource checks only isReadonly case. In general, we have different sources for different features like session manager data source, DatasetDataSource and etc. So we can specify particular data source to exclude action from certain feature? It does not seem like we get much value from the helper that only checks dataSource.isReadonly(resultIndex);

webapp/packages/plugin-data-viewer/src/ResultSet/isEditableDataSource.ts
```diff
@@ -0,0 +1,29 @@
...
+ * Returns false for readonly-by-nature sources (e.g., GroupingDataSource, DatabaseSessionManagerDataSource)
+ * or when the result index is marked as readonly.
+ *
+ * Used throughout the application to determine whether editing features should be enabled:
+ * - Cell editing menu items (add, duplicate, delete rows)
+ * - SQL generation commands (INSERT, UPDATE, DELETE)
+ * - Undo/Redo history actions
+ * - Data modification operations
+ */
+export function isEditableDataSource<TOptions = unknown>(
```
Wroud: i think this makes no sense, `isEditableDataSource` is not correct naming, because it's only for `ResultSetDataSource` type

i will suggest to remove this helpers at all

Wroud (changes requested):

PR #4214, opened 2026-03-12
dbeaver/pro#8515 sync properties state

devnaumov:
closes [8515](https://github.com/dbeaver/pro/issues/8515)

We should not sync state in the format method. However, since we already do this in our SSL, SSH, and other parts, it’s better to add it to the properties as well to keep things consistent. We should create a tech ticket and remove all logic except trim and similar operations from the format method.


webapp/packages/plugin-connections/src/ConnectionForm/DriverProperties/ConnectionFormDriverPropertiesPart.ts
```diff
@@ -69,16 +71,37 @@ export class ConnectionFormDriverPropertiesPart extends FormPart<ConnectionPrope
...
+    const properties = await this.dbDriverResource.load(this.optionsPart.state.driverId, ['includeDriverProperties']);
+
+    for (const [key, value] of Object.entries(config)) {
+      const property = properties?.driverProperties.find(property => property.id === key);
+      if (property && value === getObjectPropertyOptionValue(property.defaultValue)) {
+        delete config[key];
+      }
+    }
+
+    return config;
```
Wroud, line 108: why we can't set value to null?

Wroud (changes requested):

webapp/packages/plugin-connections/src/ConnectionForm/DriverProperties/ConnectionFormDriverPropertiesPart.ts
devnaumov: The backend returns an empty string if a null value is passed (they cant fix this easily) They reviewed their logic and asked that properties reset to default values not be returned. Therefore, we should only send properties that were actually changed

Wroud: okay, but this seems like a but on the backend side, because null is null not empty string

Wroud (dismissed):

SychevAndrey (dismissed):
Have you created the tech debt ticket?

webapp/packages/plugin-connections/src/ConnectionForm/DriverProperties/ConnectionFormDriverPropertiesPart.ts
sergeyteleshev: pls leave at least a comment for this method describing what issue it fixes

sergeyteleshev (dismissed):

sergeyteleshev (approved):

PR #4223, opened 2026-03-18
VQB: the comment is lost when returning to the VQB tab from the SQL Editor tab

sergeyteleshev:
closes https://github.com/dbeaver/pro/issues/5574


webapp/packages/core-ui/src/Tabs/TabsState.tsx
```diff
@@ -140,9 +135,11 @@ export const TabsState = observer(function TabsState<T = Record<string, any>>({
           return;
         }
         dynamic.selectedId = data.tabId;
-        if (dynamic.store.getState().selectedId !== data.tabId) {
-          dynamic.store.select(data.tabId);
+        if (dynamic.state.selectedId !== data.tabId) {
+          dynamic.state.setCurrentId(data.tabId);
+          dynamic.state.setSelectedId(data.tabId);
         }
+        dynamic.container?.select(data.tabId, dynamic.props);
```
devnaumov, line 142: What are you trying to achive here? We should not call container's select from TabsState. We can use useTabStore hook to .select tab if this is the case

devnaumov (changes requested):

webapp/packages/core-ui/src/Tabs/TabsContainer/ITabsContainer.ts
```diff
@@ -49,6 +49,7 @@ export interface ITabsContainer<TProps = void, TOptions extends Record<string, a
   readonly areaLabel: string;
   readonly tabInfoList: Array<ITabInfo<TProps, TOptions>>;
   has: (tabId: string) => boolean;
+  select: (tabId: string, props: TProps) => void;
```
Wroud, line 52: this method was previously remove, i think you should use react context and tab hooks for that

Wroud (changes requested):

PR #4224, opened 2026-03-18
dbeaver/pro#7806 remove MenuBarSmallItem

devnaumov:
closes [7806](https://github.com/dbeaver/pro/issues/7806)


webapp/packages/core-blocks/src/Button.css
```diff
@@ -7,7 +7,6 @@
  */
 @layer components {
   .dbv-kit-button {
-    --dbv-kit-btn-padding-inline: 15px;
```
devnaumov, line 10: The padding depends on the button size, otherwise it overrides

Wroud (dismissed):

webapp/packages/core-blocks/src/Button.css
SychevAndrey, line 10: what happened to this style?

webapp/common-react/@dbeaver/ui-kit/src/Button/Button.tsx
```diff
@@ -37,3 +38,6 @@ export function ButtonBaseIcon({ className, children, placement }: ButtonIconPro
 
 export const Button = componentProviderWrapper('Button', ButtonBase);
 export const ButtonIcon = componentProviderWrapper('ButtonIcon', ButtonBaseIcon);
+
+export const UnstyledButton = AriaButton;
```
SychevAndrey, line 42: I'm not a fan of this alias. What was the purpose of creating it? Just to hide Aria naming?

devnaumov: We dont expose lib naming to the core repo. UnstyledButton is common name for buttons that are headless
https://mantine.dev/core/unstyled-button/

webapp/packages/core-blocks/src/Button.css
devnaumov: Padding applies along with size prop, previously size prop did not work in terms of padding, for example buttons with small size. Every size applies its own padding, seems like we dont need this one

sergeyteleshev (dismissed):

SychevAndrey (approved):

PR #4233, opened 2026-03-23
dbeaver/pro#8518 adds message for string parameters

sergeyteleshev:
closes https://github.com/dbeaver/pro/issues/8518

- Added alert for quotes
- Removed extra margins cause it looked ugly. See the comments to compare this spacing. I checked all other places where it is integrated. It looks fine without it

image


webapp/packages/core-blocks/src/PropertiesTable/PropertiesTable.tsx
```diff
@@ -165,7 +165,6 @@ export const PropertiesTable = observer<Props>(function PropertiesTable(props) {
             onRemove={removeProperty}
           />
         ))}
```
sergeyteleshev, line 168: no need for this overflow. it eats the space and layout look asymmetric. I checked every implementation it looks fine to me now

SychevAndrey (commented):
I'm hesitating a bit. One the one hand we don't have this deletion mentioned anywhere. It's not stated in the task, we don't have in commit messages, only small github comment. And it is not related to the main task, so it will be easy to loose. On the other hand, if it really makes our app better-looking, I would like to keep it. So, maybe update the task and PR Description + add some screenshots?

sergeyteleshev:
>It's not stated in the task

It is okay to make adjustments if it is not complicated and improves something

>don't have in commit messages

And don't have to

> only small github comment

And it speaks for itself what these changes are about. Seems cool enough to me

> And it is not related to the main task,

It is related. Cause I added a new layout. And with the new layout, the whole pop-up looks wrong with margins and paddings
before:
Screenshot 2026-03-25 at 12 32 16

after:
Screenshot 2026-03-25 at 12 31 43

> On the other hand, if it really makes our app better-looking, I would like to keep it.

It is!

> So, maybe update the task and PR Description + add some screenshots?

okay!

SychevAndrey (dismissed):

webapp/packages/core-blocks/src/PropertiesTable/PropertiesTable.module.css
```diff
@@ -53,13 +53,3 @@
...
-
-.propertiesListOverflow {
-  composes: branding-overflow from global;
-  position: sticky;
-  flex-shrink: 0;
-  bottom: -4px;
-  width: 100%;
-  height: 24px;
-  pointer-events: none;
-}
```
Wroud, line 65: this was styles provided by the designer. I recommend discussing such changes on Slack and providing proper comparison screenshots (where you actually can see this overflow effect)

This effect was added to highlight user that there is some data overflow, and he can scroll (because on macos there are no scrollbars by default)

if you think that in some places this effect can be safely disabled, please make it 'opt-in' so you can disable it by component prop

Wroud (changes requested):

webapp/packages/core-blocks/src/PropertiesTable/PropertiesTable.tsx
```diff
@@ -37,10 +37,11 @@ interface Props {
   staticProperties?: boolean;
   filterable?: boolean;
   sortByName?: boolean;
+  disableBottomPadding?: boolean;
```
Wroud: `disableOverflowEffect?: boolean`

Wroud (dismissed):

webapp/packages/core-blocks/src/PropertiesTable/PropertiesTable.tsx
```diff
@@ -37,10 +37,11 @@ interface Props {
   staticProperties?: boolean;
   filterable?: boolean;
   sortByName?: boolean;
+  disableOverflowEffect?: boolean;
 }
 
 export const PropertiesTable = observer<Props>(function PropertiesTable(props) {
-  const { className, onAdd, readOnly, propertiesState } = props;
+  const { className, onAdd, readOnly, propertiesState, disableOverflowEffect = false } = props;
```
devnaumov, line 44: no need for default value

devnaumov (dismissed):

devnaumov (approved):

PR #4234, opened 2026-03-23
[CB] New and repeat passwords can be different

sergeyteleshev:
closes https://github.com/dbeaver/pro/issues/8572

https://github.com/user-attachments/assets/4742cd52-4b0b-43b2-9d84-76047df1276f


webapp/packages/plugin-user-profile/src/UserProfileForm/UserAuthenticationPart/ChangePassword.tsx
```diff
@@ -62,6 +62,20 @@ export const ChangePassword = observer(function ChangePassword() {
     return null;
   });
 
+  function onPasswordBlur() {
```
devnaumov: We already have passwordRepeatRef that basically doing the same. The native notification should appear on form save. This is behaviour we had before and it worked. Do we know what happend?

sergeyteleshev: The issue was that executords didnt execute due to a lack of the form context

Also added the fix to keep the validity state actual because after the 1st fix above when we saved the form validity state was not actuall and form didnt allow to save it but passwords were equal at the same time

SychevAndrey: I still don't understand what happened to original `useContext(FormContext)` and why we started to loose it.

sergeyteleshev: This behavior worked only for forms that were already wrapped with Form. So only for children
now I added this behavior also for parent forms

webapp/packages/core-blocks/src/FormControls/useCustomInputValidation.ts
```diff
@@ -59,6 +61,19 @@ export function useCustomInputValidation<T = void, TType extends HTMLInputElemen
...
+      function revalidateOnChange() {
+        if (!inputRef.current || inputRef.current.validity.valid) {
+          return;
+        }
+
+        validate(inputRef.current);
+      },
+    ],
+  });
+
```
Wroud: this is something new that will affect all forms with this hook, do we really want this changes? why? what bugs/problems it will cover?

Wroud (changes requested):

webapp/packages/plugin-user-profile/src/UserProfileForm/UserAuthenticationPart/ChangePassword.tsx
Wroud: While it's logical to add form as a property so it can be used in the same component with `<Form/>` and hook itself, I think this will not be obvious for developers and easily can be lost or misunderstood in refactoring

For example, when the hook is moved together with the input to the parent component with `<Form/>` developer can forget to add the form as an argument to this hook

Or when the developer decides to move the hook & input to a separate component, he can forget that he can omit this argument and will pass it to that new component as a prop

webapp/packages/core-blocks/src/FormControls/useCustomInputValidation.ts
Wroud: try this in a complex form with more than a single input with this hook (or try to change other inputs while one input is invalid)

webapp/packages/plugin-user-profile/src/UserProfileForm/UserAuthenticationPart/ChangePassword.tsx
sergeyteleshev: Okay! Added check for that case

webapp/packages/core-blocks/src/FormControls/useCustomInputValidation.ts
sergeyteleshev: I had this bug. So I decided to keep the validation state more accurate. I'll test it and see if it affected something

https://github.com/user-attachments/assets/468fde8f-0cab-4cab-a827-45e737633637

sergeyteleshev: I've tested it. All cool here

I basically fixed the case when validation of 1 field can depend on the content of another input
so each time we update something, we validate the form once again to keep the validation state actual

Wroud: Hmm, I see, but I think this should be solved in another way, because not any field that changed should trigger any field with custom validation, because it can trigger re-validation. And on your video, you can see that the invalid field will get focus, so imagine:

You have a few fields with custom validation that are invalid. You are trying to change something in the form, but this triggers validation, and the focus jumps to another field. Native field validation should not be triggered by changes in other fields

devnaumov: I think we should understand why behaviour that worked stopped working before making any further changes. It would be great to have a commit that broke it

sergeyteleshev: I'm not sure this is a regression

sergeyteleshev: Okay, the main issue is: the `password repeat` field depends semantically on the `password` field. So validation depends on the same way too

The solution is to synchronize validations when the `password` changes

At first, I tried a more automatic solution, but it affects the entire form. I see your concerns here
Then I suggest handling it manually and seeing if any other cases appear. Then, with new information, we can figure out something else

devnaumov: I checked prod, the main case is working, but its a side effect that we probably did not check, probably not regression

webapp/packages/core-blocks/src/FormControls/useCustomInputValidation.ts
```diff
@@ -11,36 +11,57 @@ import { ExecutorInterrupter } from '@cloudbeaver/core-executor';
...
+  formContext?: IFormContext,
+): ICustomInputValidation<TType> {
+  const reactContext = useContext(FormContext);
+  const context = formContext ?? reactContext;
   const inputRef = useRef<TType | null>(null);
   const translate = useTranslate();
 
+  if (!context) {
+    throw new Error('useCustomInputValidation must be used within a FormContext provider');
+  }
```
Wroud: i suggest to make new hook `useFormCustomInputValidation` with that check that will pass it as an argument to the `useCustomInputValidation` and you can remove use of `FormContext` in the `useCustomInputValidation`, because `useCustomInputValidation` can be used without form

Wroud (changes requested):

Wroud (dismissed):

Wroud (approved):

devnaumov (approved):

PR #4238, opened 2026-03-25
dbeaver/pro#7562 [CB] keep data editor filter state after reconnect

SychevAndrey:


webapp/packages/plugin-data-viewer/src/DatabaseDataModel/Actions/Grid/GridViewAction.ts
```diff
@@ -232,4 +251,50 @@ export class GridViewAction<
...
+          newOrder.push(index);
+          usedIndices.add(index);
+        }
+      }
+
+      for (const key of this.data.columns.map((_, i) => i)) {
+        if (!usedIndices.has(key)) {
+          newOrder.push(key);
+        }
+      }
```
sergeyteleshev, line 315: What does this check stand for?
What columns didn't we see in the previous cycle above?

webapp/packages/plugin-data-viewer/src/DataViewerTableState/IDataViewerPersistedState.ts
```diff
@@ -0,0 +1,22 @@
...
+ * Copyright (C) 2020-2026 DBeaver Corp and others
+ *
+ * Licensed under the Apache License, Version 2.0.
+ * you may not use this file except in compliance with the License.
+ */
+
+export interface IDataViewerPersistedState {
+  constraints: IPersistedConstraint[];
+  whereFilter: string;
+  pinnedColumns: string[];
```
sergeyteleshev: I'm a bit confused that we use column names as unique identifiers. This feels tricky and has a bug potential
In most of DBs it is forbidden to name 2 columns with the same name it is true. But is it in all DBs? I have a little doubt here

Also we can have this kind of case:
`SELECT a AS col, b AS col FROM table;`

We execute this in sql editor. See the table. What would happen? Is there a chance one of the columns disappears? Or any other related bug happened?

My suggestion is to add unique index for each column and it will resolve a lot of potential issues

webapp/packages/plugin-data-viewer/src/DataViewerPage/useDataViewerPanel.ts
```diff
@@ -78,5 +106,82 @@ export function useDataViewerPanel(tab: ITab<IObjectViewerTabState>) {
...
+          return;
+        }
+
+        try {
+          dataViewerTableStateService.saveState(tab.handlerState.objectId, dbModel);
+        } catch (exception: any) {
+          console.warn('[DataViewerTableState] Auto-save failed', exception);
+        }
+      },
+      { delay: 1000, fireImmediately: false, equals: comparer.structural },
```
sergeyteleshev: maybe we can not `throttle` it after any changes, but `debounce`? like save the state once in 5 seconds after any change?

webapp/packages/plugin-data-viewer/src/DataViewerTableState/DataViewerTableStateService.ts
```diff
@@ -0,0 +1,183 @@
...
+    if (!Array.isArray(data.pinnedColumns) || !data.pinnedColumns.every(c => typeof c === 'string')) {
+      return false;
+    }
+
+    if (data.columnOrder !== undefined && (!Array.isArray(data.columnOrder) || !data.columnOrder.every(c => typeof c === 'string'))) {
+      return false;
+    }
+
+    return true;
+  }
```
sergeyteleshev: seems like `zod` is gonna be quite handy here

webapp/packages/plugin-data-viewer/src/DataViewerTableState/IDataViewerPersistedState.ts
SychevAndrey: what will be the index then? How would you resolve it if someone added a column or two while you was working on something else and then reopen the table?

webapp/packages/plugin-data-viewer/src/DatabaseDataModel/Actions/Grid/GridViewAction.ts
SychevAndrey: The answer is the same as for indecies. Someone could edit our table and add some columns. So we will not have them in our orderColumns array. Here we put them in the end (I am open to other solutions)

webapp/packages/plugin-data-viewer/src/DataViewerPage/useDataViewerPanel.ts
SychevAndrey: why?

sergeyteleshev: Does it spam the table with requests every second if you're actively working with it? if so debounce would reduce amount of requests to save the whole state

webapp/packages/plugin-data-viewer/src/DataViewerTableState/IDataViewerPersistedState.ts
sergeyteleshev: > what will be the index then?

The backend sends the `id` for each column

> How would you resolve it if someone added a column or two while you were working on something else and then reopened the table? 

Maybe append it to the end or the beginning of the current view?

Nexus6v2 (dismissed):

webapp/packages/plugin-data-viewer/src/DataViewerPage/useDataViewerPanel.ts
SychevAndrey: ok I see. I don't think it's a big problem with comparing on each action, but I guess the deep equal is overhead here, so I just simplify it a bit and let MobX compare by pointer. You should make 1000 operation/sec to make it lag :) effect is already delayed, I guess 5 sec is too much, localstorage is fast enough and the size of saved object is relatively small until someone has thousands of columns and pinned half of them

webapp/packages/plugin-data-viewer/src/DataViewerTableState/DataViewerTableStateService.ts
```diff
@@ -0,0 +1,173 @@
...
+      clearState: action,
+    });
+
+    this.storageService.registerSettings(
+      STORAGE_KEY,
+      this.states,
+      () => new Map(),
+      (map: Map<string, IDataViewerPersistedState>) => this.validateMap(map),
+    );
+  }
```
Wroud: this data should be stored within tab's state, not as an independent state, please check tabs state and tab handlers state

Wroud (changes requested):

webapp/packages/plugin-data-viewer/src/DataViewerTableState/DataViewerTableStateService.ts
SychevAndrey: 🤦🏻 agree

webapp/packages/plugin-data-viewer/src/DatabaseDataModel/Actions/Grid/GridViewAction.ts
```diff
@@ -205,6 +206,13 @@ export class GridViewAction<
     return this.pinnedColumns.size > 0;
   }
 
+  getPinnedColumnNames(): string[] {
+    return this.columnKeys
+      .filter(key => this.isColumnPinned(key))
+      .map(key => this.getColumnName(key))
+      .filter((name): name is string => name !== undefined);
```
devnaumov: we can use isDefined for that

webapp/packages/plugin-data-viewer/src/DatabaseDataModel/Actions/Grid/GridViewAction.ts
```diff
@@ -232,4 +240,50 @@ export class GridViewAction<
...
+
+  restoreViewState(pinnedColumnNames: string[], columnOrderNames?: string[]): void {
+    if (columnOrderNames?.length) {
+      const nameToIndex = new Map<string, number>();
+
+      for (const key of this.columnKeys) {
+        const name = this.getColumnName(key);
+
+        if (name) {
+          nameToIndex.set(name, key.index);
```
devnaumov, line 295: seems like we need to add new function, something like getColumnIndexByName

webapp/packages/plugin-data-viewer/src/DatabaseDataModel/Actions/Grid/GridViewAction.ts
```diff
@@ -232,4 +240,50 @@ export class GridViewAction<
...
+
+      if (newOrder.length === this.columnsOrder.length) {
+        this.columnsOrder = newOrder;
+      }
+    }
+
+    if (pinnedColumnNames.length > 0) {
+      const keys = pinnedColumnNames
+        .map(name => this.columnKeys.find(k => this.getColumnName(k) === name))
+        .filter((k): k is IGridColumnKey => k !== undefined);
```
devnaumov: isDefined?

webapp/packages/plugin-data-viewer/src/DatabaseDataModel/Actions/DatabaseDataConstraintAction.ts
```diff
@@ -257,38 +261,52 @@ export class DatabaseDataConstraintAction
   override updateResult(result: IDatabaseResultSet): void {
     updateConstraintsForResult(this.source, result);
   }
+
+  private getColumnNameByPosition(attributePosition: number): string | undefined {
```
devnaumov, line 278: probably it should be implemented on the GridViewAction level as we might need it in different places, we have similar logic there, in general, logic that find data column by certain criteria should be gathered in one place and be accessible from different places. Also we have similar logic inside useTableData

webapp/packages/plugin-data-viewer/src/DatabaseDataModel/Actions/Grid/GridViewAction.ts
```diff
@@ -232,4 +240,50 @@ export class GridViewAction<
       this.columnsOrder = this.data.columns.map((key, index) => index);
     }
   }
+
+  restoreViewState(pinnedColumnNames: string[], columnOrderNames?: string[]): void {
```
Wroud: i think it's better to pass an interface as a single argument to this function, so it can be easily extended if needed

webapp/packages/plugin-data-viewer/src/DatabaseDataModel/Actions/DatabaseDataConstraintAction.ts
Wroud: i think it's better to implement on the `GridDataResultAction` side, but there is problem `GridDataResultAction` is abstraction over grid structures, but this function is requires specific ResultSet data type, so it should be implemented in the `ResultSetDataAction` and can be abstract in the `GridDataResultAction`

webapp/packages/plugin-data-viewer/src/DataViewerPage/useDataViewerPanel.ts
```diff
@@ -78,5 +106,88 @@ export function useDataViewerPanel(tab: ITab<IObjectViewerTabState>) {
...
+
+    const pageState = dataViewerTabService.page.getState(tab);
+    const persistedState = pageState?.persistedState;
+    const hasViewState =
+      persistedState && validatePersistedState(persistedState) && (persistedState.pinnedColumns.length > 0 || persistedState.columnOrder?.length);
+
+    let viewStateRestored = false;
+
+    const cancelViewRestore = hasViewState
+      ? reaction(
```
Wroud: since you using reaction here, it will be much easier to have some state interface, and read-mutate it directly in the database data actions, so it will be automatically synced and restored

i think you can introduce the new database data action for persisted state and re-use it in the actions

Wroud (changes requested):

PR #4241, opened 2026-03-25
dbeaver/pro#8746 add useQuery hook

devnaumov:
closes [8746](https://github.com/dbeaver/pro/issues/8746)


webapp/packages/core-blocks/src/useQuery.ts
```diff
@@ -0,0 +1,51 @@
...
+
+  useEffect(() => {
+    abortRef.current?.abort();
+    abortRef.current = new AbortController();
+
+    queryFnRef
+      .current()
+      .then(data => setState({ data, loading: false, error: null }))
+      .catch(error => {
+        if (error.name === 'AbortError') {
```
sergeyteleshev: Shouldn't we change the state here as well?

setState({ data: null, loading: false, error: null });

sergeyteleshev (changes requested):

webapp/packages/core-blocks/src/useQuery.ts
sergeyteleshev: The solution seems pretty basic. I'm sure we can just use some lib for that which already considered all of the cases
But for right now seems fine

devnaumov: Yeah, we need to understand whether we need such helper or not. Probably its better to put it inside resource and just outdate it on the route unmount. But then we add resource overhead just for the simple fn call, which it not ideal too

sergeyteleshev (dismissed):

webapp/packages/core-blocks/src/useQuery.ts
```diff
@@ -0,0 +1,52 @@
...
+  data: T | null;
+  loading: boolean;
+  error: Error | null;
+}
+
+/** We can use this if we don’t have ILoadableState and we fetch data that is not supposed to be kept in the resource */
+export function useQuery<T>(queryFn: () => Promise<T>, deps: any[] = []): QueryState<T> {
+  const [state, setState] = useState<QueryState<T>>({
+    data: null,
+    loading: true,
```
Wroud: the initial state probably should be `null, false, null`

webapp/packages/core-blocks/src/useQuery.ts
```diff
@@ -0,0 +1,52 @@
...
+  const abortRef = useRef<AbortController | null>(null);
+  const queryFnRef = useRef(queryFn);
+
+  useLayoutEffect(() => {
+    queryFnRef.current = queryFn;
+  }, [queryFn]);
+
+  useEffect(() => {
+    abortRef.current?.abort();
+    abortRef.current = new AbortController();
```
Wroud: `abortRef` not used

Wroud (changes requested):

webapp/packages/core-blocks/src/useQuery.ts
```diff
@@ -0,0 +1,221 @@
...
+ * @param {Data | (() => Data)} [options.initialData] The initial data for the hook
+ * @param {Data | (() => Data)} [options.placeholderData] The placeholder data for the hook
+ * @param {number} [options.refetchInterval] The refetch interval
+ * @param {boolean | number} [options.retry] The retry count of requests
+ * @returns {UseQueryReturn<Data>} An object with the state of the query
+ *
+ * @example
+ * const { data, isFetching, isLoading, isError, isSuccess, error, refetch, isRefetching, abort, aborted } = useQuery(() => fetch('url'));
+ */
+export const useQuery = <QueryData, Data = QueryData>(
```
Wroud: is there any change in the codestyle in the project? why arrow functions used for hooks?

webapp/packages/core-blocks/src/useQuery.ts
```diff
@@ -0,0 +1,221 @@
...
+          }
+
+          return request(action);
+        }
+        options?.onError?.(error);
+        setData(undefined);
+        setIsSuccess(false);
+        setError(error);
+        setIsError(true);
+        setIsFetching(false);
```
Wroud: why not use single state object instead?

```
setState({
  data: undefined,
  isSuccess: false,
  error,
  isError: true,
  isFetching: false
})
```

this hook also seems like AI generated, i will recommend to write unit test for such code and use more strict instructions to follow codestyle and coding patterns

webapp/packages/core-blocks/src/useDidUpdate.ts
```diff
@@ -0,0 +1,39 @@
...
+  const mountedRef = useRef(false);
+
+  useLayoutEffect(
+    () => () => {
+      mountedRef.current = false;
+    },
+    [],
+  );
+
+  useLayoutEffect(() => {
```
Wroud: i believe that logic can be implemented without adding such specific hook, especially it refers to old class components lifecycle

Wroud (changes requested):

webapp/packages/core-blocks/src/useQuery.ts
devnaumov: This hook from [reactuse](https://github.com/siberiacancode/reactuse), test cases are written there, we just need a simple helper to fetch data, if we will understand that we have many cases, we will use dedicated library for this

webapp/packages/core-blocks/src/useQuery.ts
devnaumov: Fixed

webapp/packages/core-blocks/src/useDidUpdate.ts
devnaumov: Its classic hook
https://mantine.dev/hooks/use-did-update/

webapp/packages/core-blocks/src/useQuery.ts
Wroud: why not use it as a dependency then?

webapp/packages/core-blocks/src/useDidUpdate.ts
Wroud: mantine is very old components library, therefore not all things from it follows best practices. React team intentially removed all class component lifcycle methods and replaced them with few hooks. Normally you should be able to get desired behaviour with only `useEffect`

webapp/packages/core-blocks/src/useQuery.ts
```diff
@@ -0,0 +1,221 @@
...
+          setIsRefetching(false);
+        }
+      })
+      .catch((error: Error) => {
+        if (retryCountRef.current > 0) {
+          retryCountRef.current -= 1;
+          const retryDelay = typeof options?.retryDelay === 'function' ? options?.retryDelay(retryCountRef.current, error) : options?.retryDelay;
+
+          if (retryDelay) {
+            setTimeout(() => request(action), retryDelay);
```
sergeyteleshev: I think we need to handle `clearTimeout` scenario also to avoid future bugs

sergeyteleshev (approved):

Wroud (approved):

PR #4243, opened 2026-03-26
dbeaver/pro#8644 add temp icon for base sql tab

devnaumov:
closes [8644](https://github.com/dbeaver/pro/issues/8644)

use different icon for base sql scripts and rm scripts


webapp/packages/plugin-sql-editor-navigation-tab-script/src/ResourceSqlDataSource.ts
```diff
@@ -149,7 +149,7 @@ export class ResourceSqlDataSource extends BaseSqlDataSource {
     private readonly projectsService: ProjectsService,
     state: IResourceSqlDataSourceState,
   ) {
-    super();
+    super('/icons/sql_script_m.svg');
```
sergeyteleshev, line 152: That's weird, we define an icon on the model layer
I think we should do it on the component level using helpers with `instanceof DataSource` to define which data source is that and then decide which icon to show

sergeyteleshev (changes requested):

webapp/packages/plugin-sql-editor-navigation-tab-script/src/ResourceSqlDataSource.ts
devnaumov: No, component level uses dataSource icon, please take a look at SqlEditorTab

sergeyteleshev (dismissed):

sergeyteleshev (approved):

PR #4245, opened 2026-03-27
dbeaver/pro#7251 feat: add skip navigation feature for improved acces…

SychevAndrey:
…sibility

https://github.com/user-attachments/assets/ec3fc10c-22e0-496b-80e7-5462e28987c5


SychevAndrey:
image

webapp/packages/core-app/src/AppScreen/SkipNavLinks.tsx
```diff
@@ -0,0 +1,35 @@
...
+ *
+ * Licensed under the Apache License, Version 2.0.
+ * you may not use this file except in compliance with the License.
+ */
+import { Placeholder, useTranslate } from '@cloudbeaver/core-blocks';
+import { useService } from '@cloudbeaver/core-di';
+
+import { PANEL_ID_LEFT_SIDEBAR, PANEL_ID_MAIN_CONTENT } from './AppScreenService.js';
+import { SkipNavService } from './SkipNavService.js';
+import './SkipNavLinks.css';
```
devnaumov: please use css modules, dont import css like that

webapp/packages/core-app/src/AppScreen/SkipNavLinks.tsx
```diff
@@ -0,0 +1,35 @@
...
+  element?.focus();
+}
+
+export function SkipNavLinks(): React.ReactElement {
+  const translate = useTranslate();
+  const skipNavService = useService(SkipNavService);
+
+  return (
+    <nav aria-label={translate('app_skip_nav_label')} className="dbv-kit-skip-nav">
+      <button type="button" className="dbv-kit-skip-nav__link" onClick={() => focusPanel(PANEL_ID_LEFT_SIDEBAR)}>
```
devnaumov: please use UnstyledButton instead so it will have extended accesibility features

webapp/packages/plugin-help/src/SkipNavShortcutsLink.tsx
```diff
@@ -0,0 +1,27 @@
...
+export function SkipNavShortcutsLink(): React.ReactElement {
+  const translate = useTranslate();
+  const commonDialogService = useService(CommonDialogService);
+
+  function handleClick() {
+    commonDialogService.open(ShortcutsDialog, undefined);
+  }
+
+  return (
+    <button type="button" className="dbv-kit-skip-nav__link" onClick={handleClick}>
```
devnaumov: UnstyledButton

webapp/packages/core-app/src/AppScreen/SkipNavBootstrap.ts
```diff
@@ -0,0 +1,24 @@
...
+ */
+import { Bootstrap, injectable } from '@cloudbeaver/core-di';
+
+import { AppScreenService } from './AppScreenService.js';
+import { importLazyComponent } from '@cloudbeaver/core-blocks';
+
+const SkipNavLinks = importLazyComponent(() => import('./SkipNavLinks.js').then(m => m.SkipNavLinks));
+
+@injectable(() => [AppScreenService])
+export class SkipNavBootstrap extends Bootstrap {
```
devnaumov: Move it to a separate plugin, plugin-focus-skip, mention "focus" also in the file names

devnaumov (changes requested):

webapp/packages/core-app/src/AppScreen/SkipNavLinks.tsx
SychevAndrey: skip-nav styles are inherently global: SkipNavShortcutsLink in plugin-help needs the same styles as the main container. With CSS modules this will leed to duplicating the same styles in two .module.css files (in plugin-focus-skip and plugin-help), whereas a single plain CSS file with a BEM class covered both components with zero duplication.

webapp/packages/core-app/src/AppScreen/SkipNavLinks.tsx
SychevAndrey: What accessibility features UnstyledButton adds on top of what a native button already provides?

webapp/packages/core-app/src/AppScreen/SkipNavBootstrap.ts
SychevAndrey: skip-nav is essentially part of the app shell layout (just like TopNavBar, which also lives in appScreenService.placeholder). It's 3 files of actual code. Extracting it into a standalone package results in ~10 boilerplate files for 3 files of logic, plus a new dependency from plugin-help to plugin-focus-skip.

webapp/packages/plugin-help/src/SkipNavShortcutsLink.tsx
SychevAndrey: even ariakit docs suggest using native button instead their component

webapp/packages/core-app/src/AppScreen/SkipNavLinks.tsx
devnaumov: We are not going to use styles like this. We want to stick with css modules. Please be consistent with a codebase patterns, if you want to reuse the same styles, you can use` export { default as CellStyles } from './Cell.module.css'; ` Or you can use tailwind

webapp/packages/core-app/src/AppScreen/SkipNavLinks.tsx
devnaumov: You can keep native button but if we will have the desire to change something about all unstyled buttons in the app, we will not be able to

webapp/packages/plugin-help/src/SkipNavShortcutsLink.tsx
devnaumov: I think they are talking about render prop, you can render button as different html element, they suggest to keep it native. Ariakit uses native button underneath, so they are equal in terms of browser support

webapp/packages/plugin-help/src/SkipNavShortcutsLink.tsx
```diff
@@ -0,0 +1,27 @@
...
+ */
+import { importLazyComponent, useTranslate } from '@cloudbeaver/core-blocks';
+import { useService } from '@cloudbeaver/core-di';
+import { CommonDialogService } from '@cloudbeaver/core-dialogs';
+
+const ShortcutsDialog = importLazyComponent(() => import('./Shortcuts/ShortcutsDialog.js').then(m => m.ShortcutsDialog));
+
+export function SkipNavShortcutsLink(): React.ReactElement {
+  const translate = useTranslate();
+  const commonDialogService = useService(CommonDialogService);
```
Wroud: same here

webapp/packages/core-app/src/AppScreen/SkipNavLinks.tsx
```diff
@@ -0,0 +1,35 @@
...
+import './SkipNavLinks.css';
+
+function focusPanel(panelId: string) {
+  const element = document.querySelector<HTMLElement>(`[data-panel-id="${panelId}"]`);
+  element?.focus();
+}
+
+export function SkipNavLinks(): React.ReactElement {
+  const translate = useTranslate();
+  const skipNavService = useService(SkipNavService);
```
Wroud, line 24: you should use `observable` with `useTranslate`

webapp/packages/core-app/src/AppScreen/SkipNavBootstrap.ts
devnaumov: As a general tip, if you are creating a bootstrap class, it is a good indication that the logic you are implementing should be in a separate plugin. The number of lines does not matter, we have plugins with even fewer lines. Our goal is to keep the app modular, so you should be able to remove a plugin from the declaration file and have the feature removed along with it.

Wroud: i think it's better to keep it here for now

webapp/packages/core-app/src/AppScreen/SkipNavLinks.css
```diff
@@ -0,0 +1,40 @@
+@layer components {
+  .dbv-kit-skip-nav {
+    position: absolute;
+    top: 8px;
+    left: 8px;
+    z-index: 10000;
+    display: flex;
+  }
+
+  .dbv-kit-skip-nav__link {
```
devnaumov: Why we use dbv kit prefixes here?

devnaumov (changes requested):

webapp/packages/core-app/src/AppScreen/SkipNavLinks.tsx
Wroud: i agree, that in this case it's better to use tailwind css or css-modules, because this components does not differ much from our regular components in the app

webapp/packages/core-app/src/AppScreen/SkipNavLinks.tsx
```diff
@@ -0,0 +1,35 @@
...
+
+  return (
+    <nav aria-label={translate('app_skip_nav_label')} className="dbv-kit-skip-nav">
+      <button type="button" className="dbv-kit-skip-nav__link" onClick={() => focusPanel(PANEL_ID_LEFT_SIDEBAR)}>
+        {translate('app_skip_nav_navigator')}
+      </button>
+      <button type="button" className="dbv-kit-skip-nav__link" onClick={() => focusPanel(PANEL_ID_MAIN_CONTENT)}>
+        {translate('app_skip_nav_main_content')}
+      </button>
+      <Placeholder container={skipNavService.extraLinks} />
```
Wroud, line 34: is there can be only buttons? if so, isn't it better to make special container that will accept: panelId, label and will render buttons same way

webapp/packages/core-app/src/AppScreen/SkipNavBootstrap.ts
SychevAndrey: so TopNavBar should be a separate plugin as well?

Wroud (changes requested):

webapp/packages/core-app/src/AppScreen/SkipNavBootstrap.ts
devnaumov: You can create a method inside NavService, something like registerLinks and call it from AppScreenBootstrap register method so we only have one bootstrap per package

webapp/packages/core-app/src/AppScreen/SkipNavLinks.tsx
SychevAndrey: actually, you can put anything here. And adjust it to your workflow, for example. And it's not about only focusing panels. For example, plugin-help registers button, but it doesn't focus anything, it opens shortcuts dialog. Placeholder is a good fit here.

Wroud (approved):

sergeyteleshev (approved):

devnaumov (approved):

PR #4246, opened 2026-03-30
[CB] DB Navigator - add subgroup to context menu to reduce the list of actions

sergeyteleshev:
closes https://github.com/dbeaver/pro/issues/8236


Wroud (dismissed):

devnaumov (dismissed):

webapp/packages/core-ui/src/ContextMenu/SubMenuElement.tsx
Wroud: you can remove `const hasVisibleItems = subMenuData.items.some(item => !item.hidden);` in the `const hidden = getComputed(() => {` block and reuse one you've added above

webapp/packages/core-ui/src/ContextMenu/SubMenuElement.tsx
```diff
@@ -193,6 +194,7 @@ export const SubMenuElement = observer<ISubMenuElementProps>(function SubMenuEle
       render={
         <MenuItem
           id={subMenu.id}
+          disabled={disabled}
```
Wroud: you don't need to pass it here, it's already passed to the `MenuComponent`

Wroud (changes requested):

webapp/packages/core-ui/src/ContextMenu/SubMenuElement.tsx
sergeyteleshev: no I need it to do so 
image

this behavior didnt work

sergeyteleshev: i cannot delete this manage right now cause it is empty bacuse of legacy bug in submenus and loaders. I created a ticket for that

webapp/packages/core-ui/src/ContextMenu/SubMenuElement.tsx
sergeyteleshev: fixed

webapp/packages/core-ui/src/ContextMenu/SubMenuElement.tsx
devnaumov: So now we show disabled state with arrows in case if there is no subitems?

sergeyteleshev: okay I added a local solution to fix this issue
But in general, we should refactor the logic of loaders and nested loaders for submenus. cause right now we have here some hacks to have behavior we want

webapp/packages/core-ui/src/ContextMenu/SubMenuElement.tsx
```diff
@@ -111,11 +112,32 @@ export const SubMenuElement = observer<ISubMenuElementProps>(function SubMenuEle
...
+  }
+
   // TODO: need to be fixed, in case when menu depend on data from loaders this may be always true
   if (hidden && !action) {
     return null;
   }
 
+  const visibleItems = getComputed(() => subMenuData.items.filter(item => !item.hidden));
+
+  if (!action && visibleItems.length === 1) {
```
devnaumov: Please revert, if we have only one item in menu, we show it as subitem, we probably will implement such behvaiour but not in the scope of this ticket

devnaumov (changes requested):

webapp/packages/plugin-connections/src/ContextMenu/ConnectionMenuBootstrap.ts
```diff
@@ -63,17 +65,16 @@ export class ConnectionMenuBootstrap extends Bootstrap {
   override register(): void {
     this.addConnectionsMenuToTopAppBar();
 
+    this.menuService.addCreator({
```
devnaumov, line 68: Please refer to the MENU_DATA_GRID_FILTERS, if there is no registered actions for the menu, root menu should not be shown

devnaumov (changes requested):

Wroud (approved):

devnaumov (approved):

PR #4264, opened 2026-04-06
dbeaver/pro#4158 add special filter operators

### LINKED ISSUE #4158: Navigation tree advanced filter
**Describe the solution you'd like**

When filtering databases in would be nice to have a regex filter.

In dbeaver i often use "product|category" style.

devnaumov:
closes [4158](https://github.com/dbeaver/cloudbeaver/issues/4158)

Added special filter operators, |,*
| -> OR (HIGH PRIORITY), , -> AND, * -> wildcard
applies to Nav tree, scripts, cloud storage, datasets

CleanShot 2026-04-06 at 14 48 22@2x

### DIFF
--- a/webapp/packages/plugin-navigation-tree/src/NavigationTree/ElementsTree/elementsTreeNameFilter.ts
+++ b/webapp/packages/plugin-navigation-tree/src/NavigationTree/ElementsTree/elementsTreeNameFilter.ts
@@ -76,8 +76,8 @@ export function elementsTreeNameFilterNode(tree: IElementsTree, node: NavNode, f
   return createFilter(filterToLower)(nodeName) ? EEquality.partially : EEquality.none;
 }
 
-function wildcardMatch(str: string, pattern: string): boolean {
-  const s = str.toLowerCase();
+function wildcardMatch(value: string, pattern: string): boolean {
+  const s = value.toLowerCase();
   const p = pattern.trim().toLowerCase();
   const parts = p.split('*');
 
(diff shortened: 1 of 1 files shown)


SychevAndrey (approved):

webapp/packages/plugin-navigation-tree/src/NavigationTree/ElementsTree/elementsTreeNameFilter.ts
```diff
@@ -73,5 +73,67 @@ export function elementsTreeNameFilterNode(tree: IElementsTree, node: NavNode, f
...
+
+  return groups.every(group => {
+    const alts = group
+      .split('|')
+      .map(a => a.trim())
+      .filter(Boolean);
+
+    return alts.some(alt => wildcardMatch(value, alt));
+  });
+}
```
sergeyteleshev, line 135: Maybe use RegExp match check itself? Why do we have just a few cases of regexp check instead of the whole tool? user required regexp search at first place, with an example using the `|` operator

My suggestion is to implement:
includes search + fuzzy search + regexp check

sergeyteleshev: `packages/core-blocks/src/fuzzyMatch.ts`

devnaumov: No need for that complexity. We want to have ability to search using and or and wildcard operators, we dont need groups and etc, also we dont want any special toggles that will change how filter works. Basically  this is how desktop works, it has these three operators. Decision not to use regex  was agreed with the team

PR #4274, opened 2026-04-08
dbeaver/cloudbeaver#4249 added session rotation

### LINKED ISSUE #4249: No session rotation on login
### Description

Couldn't find settings for cb-session-id rotation on login. I have the same session cookie before and after successful authentication.  I suppose this can cause session fixation attack

### Steps to reproduce

1. open cloudbeaver without authentication in one browser and take its session id
2. open cloudbeaver in incognito mode, paste session id and log in
3. refresh page opened in first step - you'll be logged in as a user from second step

### Expected/Desired Behavior

Provide new cb-session-id after successful login

### CloudBeaver Version

CloudBeaver Community 26.0.1

### Additional context

_No response_

Nexus6v2:


yagudin10 (dismissed):

yagudin10 (dismissed):

server/bundles/io.cloudbeaver.model/src/io/cloudbeaver/model/session/BaseWebSession.java
```diff
@@ -161,6 +161,10 @@ public String getSessionId() {
         return id;
     }
 
+    public void setSessionId(@NotNull String id) {
```
serge-rider: Not sure we have to "rotate" existing sessions.
The idea was just to change cookie value. Thus new WebSession must be created automatically.

serge-rider (changes requested):

server/bundles/io.cloudbeaver.model/src/io/cloudbeaver/model/session/BaseWebSession.java
Nexus6v2: Updated so that we create new session before each login

server/bundles/io.cloudbeaver.service.auth/src/io/cloudbeaver/service/auth/impl/WebServiceAuthImpl.java
```diff
@@ -67,14 +67,19 @@ public class WebServiceAuthImpl implements DBWServiceAuth {
...
+        @NotNull HttpServletRequest httpRequest,
+        @NotNull WebSession inputWebSession,
         @NotNull String providerId,
         @Nullable String providerConfigurationId,
         @Nullable Map<String, Object> authParameters,
         boolean linkWithActiveUser,
         boolean forceSessionsLogout
     ) throws DBWebException {
         try {
+            // Rotate web session during each login attempt to prevent session fixation attacks
```
serge-rider: What if this is an additional login (e.g. AWS)?
The same for federatedLogin.

server/bundles/io.cloudbeaver.model/src/io/cloudbeaver/model/session/BaseWebSession.java
```diff
@@ -116,6 +116,13 @@ public void removeEventHandler(@NotNull CBWebSessionEventHandler handler) {
         }
     }
 
+    public void migrateEventHandlersTo(@NotNull BaseWebSession target) {
```
serge-rider, line 119: How will it work in case of multi-node env?
And why do we need to migrate event handlers at all? Generally new login means that old session is no more valuable, we should just close it

serge-rider (changes requested):

server/bundles/io.cloudbeaver.model/src/io/cloudbeaver/model/session/BaseWebSession.java
Nexus6v2: > How will it work in case of multi-node env?

I believe it will only be different in case if sticky-sessions routing based on this cookie is used. As far as I know it's not, so it should behave the same.

> And why do we need to migrate event handlers at all? Generally new login means that old session is no more valuable, we should just close it

There is an open web socket that we need to migrate to new session, otherwise the flow breaks and UI waits for auth with an endless spinner

server/bundles/io.cloudbeaver.service.auth/src/io/cloudbeaver/service/auth/impl/WebServiceAuthImpl.java
Nexus6v2: Good question.
My guess is session should also be rotated for additional logins, since we would have the same vulnerability:
1. User logs in
2. Session cookie is compromised
3. User does additional login
4. Compromised session cookie gains access to new auth

Rotated session should also keep the existing auth

I will update the PR

Nexus6v2: Decided on the call to keep rotation only for the first login (anonymous only)

PR #4275, opened 2026-04-08
[CB] SQL Query generation: add settings for the Generated SQL window

sergeyteleshev:
closes https://github.com/dbeaver/pro/issues/8328


webapp/packages/plugin-data-spreadsheet-new/src/DataGrid/DataGridContextMenu/DataGridContextMenuGenerateSqlService.ts
```diff
@@ -40,6 +40,9 @@ import type { IDataContextProvider } from '@cloudbeaver/core-data-context';
 import { GeneratedSqlDialog, SqlGeneratorsResource } from '@cloudbeaver/plugin-sql-generator';
 import { isNotNullDefined } from '@dbeaver/js-helpers';
 
+const DEFAULT_USE_FULLY_QUALIFIED_NAMES = true;
+const DEFAULT_COMPACT_SQL = false;
```
Wroud: why do you have this constants in multiple files?

Wroud (dismissed):

webapp/packages/plugin-sql-generator/src/SqlGenerators/SqlGeneratorsBootstrap.ts
```diff
@@ -81,13 +84,21 @@ export class SqlGeneratorsBootstrap extends Bootstrap {
                 {
                   onSelect: async () => {
                     try {
-                      const query = await this.sqlGeneratorsResource.generateEntityQuery(action.id, node.id);
+                      const query = await this.sqlGeneratorsResource.generateEntityQuery(
+                        action.id,
+                        node.id,
+                        DEFAULT_USE_FULLY_QUALIFIED_NAMES,
```
devnaumov: Declare this defaults constants in sqlGeneratorsResource

devnaumov (dismissed):

Wroud (dismissed):

webapp/packages/plugin-sql-generator/src/SqlGenerators/GeneratedSqlDialog.tsx
```diff
@@ -48,20 +70,73 @@ export const GeneratedSqlDialog = observer<DialogComponentProps<Payload>>(functi
     extensions.set(...sqlDialect);
   }
 
+  async function handleOpenInEditor() {
+    try {
+      await sqlEditorNavigatorService.openNewEditor({
```
devnaumov, line 97: If the editor fails to open, it’s not a good idea to close this dialog. It’s better to keep it open, since the user’s intention is to copy content into the editor. This way, the user can copy it manually if opening the editor fails.

webapp/packages/plugin-sql-generator/src/SqlGenerators/GeneratedSqlDialog.tsx
```diff
@@ -48,20 +70,73 @@ export const GeneratedSqlDialog = observer<DialogComponentProps<Payload>>(functi
...
+        connectionKey: connection ? createConnectionParam(connection) : undefined,
+        query: state.query,
+      });
+    } catch (error: any) {
+      notificationService.logException(error, 'app_shared_sql_generators_error_open_editor');
+    }
+    rejectDialog();
+  }
+
+  async function handleOptionChange<T extends keyof typeof state>(key: T, value: (typeof state)[T]) {
```
devnaumov: Move these functions to the state, otherwise the state kind of useless or you can just use react state

webapp/packages/plugin-data-spreadsheet-new/src/DataGrid/DataGridContextMenu/DataGridContextMenuGenerateSqlService.ts
```diff
@@ -137,31 +137,77 @@ export class DataGridContextMenuGenerateSqlService {
...
+      connectionId,
+      contextId,
+      resultsId: resultId,
+      generatorId,
+      selectedRows: rows,
+      generatorOptions: options,
+    });
+
+    if (!query) {
+      this.notificationService.logError({
```
devnaumov: I think we should just throw here, does it make sense to return null? You fallback to the empty string later in the components anyway. 
`Promise<string> `

webapp/packages/plugin-sql-generator/src/SqlGenerators/SqlGeneratorsResource.ts
```diff
@@ -8,9 +8,18 @@
...
-import { GraphQLService, type SqlGenerateResultSetQueryQueryVariables, type SqlQueryGenerator } from '@cloudbeaver/core-sdk';
+import {
+  GraphQLService,
+  type SqlGenerateResultSetQueryQueryVariables,
+  type SqlQueryGenerator,
+  type SqlQueryGeneratorOptions,
+} from '@cloudbeaver/core-sdk';
 
 export const MAX_GENERATORS_LENGTH = 15;
+export const getDefaultQueryGeneratorOptions = (): SqlQueryGeneratorOptions => ({
```
devnaumov, line 19: As we always pass default options ,we can simplify it, make it just an object, const DEFAULT_QUERY_GENERATOR_OPTIONS and use in generateEntityQuery like this. In this case there will be no need to expose it even

`generatorOptions: {...DEFAULT_QUERY_GENERATOR_OPTIONS, ...options}`

devnaumov (changes requested):

webapp/packages/plugin-data-spreadsheet-new/src/DataGrid/DataGridContextMenu/DataGridContextMenuGenerateSqlService.ts
sergeyteleshev: In this particular case throwing is fine so I implemented it

>  does it make sense to return null

In general, yes. This is a proper way to say that "no value" returned on purpose. Knowing that in the implementation of the method we can handle this case properly

> You fallback to the empty string later in the components anyway.

Yes, this is the part of handling the the `null` case

All in all I simplified the code a bit since we use throwing now

webapp/packages/plugin-sql-generator/src/SqlGenerators/SqlGeneratorsResource.ts
sergeyteleshev: Let's leave it as it is. This is the same technique we apply for the default state of the form parts, for example, and the reason we need it is that if we use just an object, this means it can be mutated. Instead, with a function, we create a new object each time we call it, reducing the probability of a stinky bug caused by mutations. So the current implementation is more predictable cause someone, somewhere, can just forget to use the spread operator

webapp/packages/plugin-sql-generator/src/SqlGenerators/GeneratedSqlDialog.tsx
```diff
@@ -34,32 +34,51 @@ interface Payload {
...
+          });
+        } catch (error: any) {
+          this.notificationService.logException(error, 'app_shared_sql_generators_error_title');
+        } finally {
+          this.loading = false;
+        }
+      },
     }),
     {
+      handleOptionChange: observable.ref,
```
devnaumov: handleOptionChange is action, so -> action.bound

devnaumov (dismissed):

webapp/packages/plugin-sql-generator/src/SqlGenerators/GeneratedSqlDialog.tsx
```diff
@@ -117,7 +119,7 @@ export const GeneratedSqlDialog = observer<DialogComponentProps<Payload>>(functi
                 name="useFullyQualifiedNames"
                 disabled={state.loading}
                 label={translate('app_shared_sql_generators_use_fully_qualified_names')}
-                onChange={value => handleOptionChange('useFullyQualifiedNames', value)}
+                onChange={value => state.handleOptionChange('useFullyQualifiedNames', value)}
```
devnaumov, line 125: You are already using state and name pattern, so, even without onChange the state object will be updated.  What you can do to simplify it is to wrap it inside Form and use form onChange event to fire side effect such as regenerate query. 

If you are using onChange, then you dont need state and name, only checked prop

devnaumov (approved):

PR #4278, opened 2026-04-12
dbeaver/pro#8352 do not unlink connection

devnaumov:
closes [8352](https://github.com/dbeaver/pro/issues/8352)


webapp/packages/plugin-datasource-context-switch/src/ConnectionSchemaManager/ConnectionSchemaManagerBootstrap.ts
```diff
@@ -125,7 +125,25 @@ export class ConnectionSchemaManagerBootstrap extends Bootstrap {
...
+          {
+            isDisabled: () => !this.connectionSchemaManagerService.currentConnectionKey,
+          },
+        );
+
+        const fixed: typeof items = [new ContextMenuSearchItem()];
+
+        if (!this.connectionSchemaManagerService.currentConnectionRequired) {
+          fixed.push(noneItem, new MenuSeparatorItem());
+        }
```
Wroud:
```suggestion
        const fixed: typeof items = [new ContextMenuSearchItem()];

        if (!this.connectionSchemaManagerService.currentConnectionRequired) {
          fixed.push(
            new MenuBaseItem(
              { id: 'none', label: 'core_connections_no_connection', icon: '/icons/database_sm.svg' },
              {
                onSelect: () => {
                this.connectionSchemaManagerService.selectConnection(null);
              },
            },
            {
              isDisabled: () => !this.connectionSchemaManagerService.currentConnectionKey,
            },
          ),
          new MenuSeparatorItem());
        }
```

webapp/packages/plugin-sql-editor/src/MenuBootstrap.ts
```diff
@@ -266,14 +266,20 @@ export class MenuBootstrap extends Bootstrap {
...
         if (EXECUTIONS_ACTIONS.includes(action)) {
           return data.isDisabled || data.isScriptEmpty;
         }
 
         switch (action) {
-          case ACTION_SQL_EDITOR_FORMAT:
-            return data.isDisabled || data.isScriptEmpty || data.readonly;
+          case ACTION_SQL_EDITOR_FORMAT: {
+            const source = this.sqlDataSourceService.get(state.editorId);
+            const context = source?.executionContext;
```
Wroud: you can use `sqlEditorData.model.dataSource.executionContext`

webapp/packages/plugin-sql-editor/src/MenuBootstrap.ts
```diff
@@ -266,14 +266,20 @@ export class MenuBootstrap extends Bootstrap {
...
+            const connection = context ? this.connectionInfoResource.get(createConnectionParam(context.projectId, context.connectionId)) : null;
+
+            return data.isDisabled || data.isScriptEmpty || data.readonly || !connection?.connected;
+          }
```
Wroud, line 288: i don't think that you need to check for `connection?.connected` because `executionContext` will not be available without connection

Wroud (dismissed):

webapp/packages/plugin-sql-editor/src/MenuBootstrap.ts
devnaumov: Connection might exist, but its not connected. In this case we should disable it because backend throw error if connection is not connected and we try format something

Wroud: my point is that you can check only for execution context, because if there is no connection it will be no execution context

Wroud (dismissed):

sergeyteleshev (approved):

webapp/packages/plugin-sql-editor/src/MenuBootstrap.ts
devnaumov: Context will be still there, it will disappear only when we will reset the connection.

PR #4282, opened 2026-04-14
dbeaver/pro#5680 removed swagger deps for rest api documentation

HocKu7:
Closes https://github.com/dbeaver/pro/issues/5680


ShadelessFox (commented):
Maybe we'll start documenting new APIs, especially annotations?

PR #4285, opened 2026-04-15
dbeaver/pro#8234 only admin can overwrite existing udbt handler

HocKu7:
Closes https://github.com/dbeaver/pro/issues/8234


server/bundles/io.cloudbeaver.model/src/io/cloudbeaver/model/rm/local/LocalResourceController.java
```diff
@@ -411,6 +414,107 @@ public boolean updateProjectDataSources(
...
+                continue;
+            }
+
+            if (!areSameHandlerConfiguration(storedHandler, updatedHandler)) {
+                throw new DBException("No permissions to modify network handler '" + handlerId + "'");
+            }
+        }
+    }
+
+    private boolean areSameHandlerConfiguration(
```
serge-rider, line 505: add compare function to handler itself. Or just compare by ID, why so complicated?

serge-rider (changes requested):

PR #4287, opened 2026-04-16
dbeaver/pro#8711 sql execution plan improvements

HocKu7:
Closes https://github.com/dbeaver/pro/issues/8711


server/bundles/io.cloudbeaver.server/src/io/cloudbeaver/service/sql/WebSQLExecutionPlanNode.java
```diff
@@ -73,6 +76,26 @@ public String getDescription() {
         return node.getNodeDescription();
     }
 
+    @Nullable
+    public Double getCost() {
+        return getMetric(DBCPlanCostNode::getNodeCost);
```
serge-rider: Why not just `node instanceof DBCPlanCostNode cn ? cn.getNodeCost() : null`?

It much more readable.
Or create WebSQLExecutionPlanCostNode class to avoid type checks.

serge-rider (changes requested):

yagudin10 (approved):

PR #4298, opened 2026-04-21
[CB] Support real tab characters in SQL Editor

sergeyteleshev:
closes https://github.com/dbeaver/pro/issues/7864

### DIFF
--- a/webapp/packages/plugin-codemirror6/src/useEditorDefaultExtensions.ts
+++ b/webapp/packages/plugin-codemirror6/src/useEditorDefaultExtensions.ts
@@ -1,6 +1,6 @@
 /*
  * CloudBeaver - Cloud Database Manager
- * Copyright (C) 2020-2025 DBeaver Corp and others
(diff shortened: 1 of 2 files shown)


webapp/packages/plugin-codemirror6/src/useEditorDefaultExtensions.ts
```diff
@@ -109,6 +111,7 @@ const extensionMap = {
   rectangularSelection,
   keymap: () => keymap.of(DEFAULT_KEY_MAP),
   lineWrapping: () => EditorView.lineWrapping,
+  tabIndentation: () => indentUnit.of('\t'),
```
devnaumov: To keep things in sync, it’s better to declare the indent unit as a constant and export it. Then, in the query builder, use this variable instead of hardcoded values. This way, if we change it later, it will automatically be applied to the query builder as well.

`export const INDENT_UNIT = '\t';`

devnaumov (approved):

PR #4299, opened 2026-04-22
dbeaver/pro#8978 paste multiple values

devnaumov:
closes [8978](https://github.com/dbeaver/pro/issues/8978)


webapp/packages/plugin-data-spreadsheet-new/src/DataGrid/useTableData.tsx
```diff
@@ -179,6 +179,7 @@ export function useTableData(
       editor: observable.ref,
       view: observable.ref,
       gridDIVElement: observable.ref,
+      isCellEditable: action.bound,
```
sergeyteleshev: `isCellEditable` does not mutate the state so it is not necessary to add `action.bound` here

webapp/packages/plugin-data-spreadsheet-new/src/DataGrid/useGridSelectedCellsPaste.ts
```diff
@@ -43,7 +56,79 @@ export function useGridSelectedCellsPaste(
...
+      for (let r = 0; r < rows; r++) {
+        for (let c = 0; c < cols; c++) {
+          const target = targetMap.get(`${r}:${c}`);
+          const value = matrix[r]?.[c];
+
+          if (target && isDefined(value) && tableData.isCellEditable(target)) {
+            updates.push({ key: target, value });
+          }
+        }
+      }
```
sergeyteleshev, line 132: we should do something about this complicated case:

https://github.com/user-attachments/assets/9c2d2e53-cce2-4e5c-8d81-9622f60a81b5

devnaumov: We should not, it works as expected

webapp/packages/plugin-data-spreadsheet-new/src/DataGrid/useTableData.tsx
devnaumov: We bind it so the context isn’t lost and it can be passed as a callback. In general, its better to keep any function inside an useObservableRef as action.bound for simplicity

Wroud: any action is not trackable by mobx, so if your code depends on the `isCellEditable` in observable context it will not work as expected

webapp/packages/plugin-data-spreadsheet-new/src/DataGrid/useGridSelectedCellsPaste.ts
```diff
@@ -43,7 +56,79 @@ export function useGridSelectedCellsPaste(
...
+
+      if (rows === 1 && cols === 1) {
+        const value = matrix[0]?.[0];
+
+        if (!isDefined(value)) {
+          return;
+        }
+
+        const updates = targets.filter(key => tableData.isCellEditable(key)).map(key => ({ key, value }));
+        tableData.editor.setMany(updates);
```
devnaumov, line 98: we should return here

webapp/packages/plugin-data-spreadsheet-new/src/DataGrid/useTableData.tsx
Wroud: please use manual js bind instead, or better do not spread this method from the object

Wroud (dismissed):

webapp/packages/plugin-data-spreadsheet-new/src/DataGrid/useTableData.tsx
devnaumov: its not needed anymore, i will remove it

webapp/packages/plugin-data-spreadsheet-new/src/DataGrid/useGridSelectedCellsPaste.ts
```diff
@@ -43,7 +56,80 @@ export function useGridSelectedCellsPaste(
...
+      const matrix = parseTSV(clipboardText);
+      const rows = matrix.length;
+      const cols = matrix[0]?.length ?? 0;
+
+      if (rows === 0 || cols === 0) {
+        return;
+      }
+
+      const getRowIdx = (key: IGridDataKey) => tableData.getRowIndexFromKey(key.row);
+      const getColIdx = (key: IGridDataKey) => tableData.getColumnIndexFromColumnKey(key.column);
```
sergeyteleshev, line 72: We need to consider the cases with the pin columns & reordering of columns:

https://github.com/user-attachments/assets/adfa2405-23da-4bfd-b067-1ab03f6b4a59

sergeyteleshev (changes requested):

webapp/packages/plugin-data-spreadsheet-new/src/DataGrid/useGridSelectedCellsPaste.ts
devnaumov: Its considered, getRowIndexFromKey gives you real index, the same as for copying values

sergeyteleshev (approved):

PR #4300, opened 2026-04-23
dbeaver/pro#8107 [CB] increase the contrast of some elements

SychevAndrey:


webapp/packages/core-blocks/src/Button.css
```diff
@@ -23,10 +23,10 @@
     --dbv-kit-btn-secondary-background: transparent;
     --dbv-kit-btn-secondary-background-hover: color-mix(in srgb, transparent, var(--theme-primary) 10%);
     --dbv-kit-btn-secondary-background-active: color-mix(in srgb, transparent, var(--theme-primary) 20%);
-    --dbv-kit-btn-secondary-foreground: var(--theme-link-color);
-    --dbv-kit-btn-secondary-foreground-hover: var(--theme-link-color);
-    --dbv-kit-btn-secondary-foreground-active: var(--theme-link-color);
-    --dbv-kit-btn-secondary-border-color: var(--theme-link-color);
+    --dbv-kit-btn-secondary-foreground: #2674ab;
```
devnaumov, line 26: I think its better to handle these colors through theme vairables because otherwise if we change theme colors, it will not affect button and we have to adjust it separatly, the same with blob ad etc

devnaumov (approved):

PR #4302, opened 2026-04-24
dbeaver/pro#9047 added sql plan explain to CB

Nexus6v2:
Closes dbeaver/pro#9047


server/bundles/io.cloudbeaver.server/src/io/cloudbeaver/service/sql/WebSQLExecutionPlan.java
```diff
@@ -40,6 +40,10 @@ public WebSQLExecutionPlan(WebSession webSession, DBCPlan plan) {
         this.plan = plan;
     }
 
+    public DBCPlan getPlan() {
```
serge-rider, line 47: Nullability

serge-rider (changes requested):

PR #4305, opened 2026-04-27
[CB] Configure AI functions and MCP in the administration part

sergeyteleshev:
closes dbeaver/pro#8924


webapp/packages/core-utils/src/parseHeaders.ts
```diff
@@ -0,0 +1,38 @@
...
+export function headersToText(headers: Record<string, string> | undefined): string | undefined {
+  if (!headers || Object.keys(headers).length === 0) {
+    return undefined;
+  }
+  return Object.entries(headers)
+    .map(([key, value]) => `${key}=${value}`)
+    .join('\n');
+}
+
+export function textToHeaders(text: string | undefined): Record<string, string> | undefined {
```
Wroud: I really don’t like the idea of having these methods. Please try to refactor the logic so they are no longer needed, or use a dedicated library (or native JavaScript features) that guarantees correct parsing

PR #4331, opened 2026-05-08
dbeaver/cloudbeaver#4249 added session rotation (#4274)

### LINKED ISSUE #4249: No session rotation on login

serge-rider:


server/bundles/io.cloudbeaver.service.auth/src/io/cloudbeaver/service/auth/impl/WebServiceAuthImpl.java
```diff
@@ -67,14 +67,22 @@ public class WebServiceAuthImpl implements DBWServiceAuth {
 
     @Override
     public WebAuthStatus authLogin(
-        @NotNull WebSession webSession,
+        @NotNull HttpServletRequest httpRequest,
+        @NotNull WebSession inputWebSession,
```
serge-rider, line 71: What is inputWebSession?
Is it OLD session for the same cookie?
Currently name is confusing as you use inputWebSession in other functions with different meaning.

HocKu7: authLogin knows nothing about the rotating session process. It has only current web session. I think that naming is ok for this method

PR #4332, opened 2026-05-11
[CB] Add shortcut for the cell context menu in data and metadata editors

sergeyteleshev:
closes https://github.com/dbeaver/pro/issues/8424


webapp/packages/core-ui/src/ContextMenu/ContextMenu.tsx
```diff
@@ -75,6 +77,8 @@ export const ContextMenu = observer<IContextMenuNewProps>(function ContextMenuIn
     }
   }, [showAtPosition, menu]);
 
+  menu.setAutoFocusOnShow(autoFocusOnShow);
```
sergeyteleshev: if we trigger menu on shortcut we dont press `MenuButton` component and menu is not available to navigate inside it via keyboard shortcuts by default
passing prop `autoFocusOnShow` to `Menu` component doesn't work: seems like it requires pressing manually `MenuButton`
I have researched and nothing is stoles this focus from the Menu items popover during render of our `ContextMenu`, so the only working solution is to use this line of code

webapp/packages/core-view/src/View/parseHotkey.test.ts
```diff
@@ -0,0 +1,222 @@
...
+    expect(isBindingPressed(event as any, binding)).toBe(true);
+  });
+
+  it('returns true when mod binding matches metaKey event', () => {
+    const binding: IKeyBinding = { id: 'test', keys: 'mod+s' };
+    const event = makeEvent({ metaKey: true, key: 's' });
+    expect(isBindingPressed(event as any, binding)).toBe(true);
+  });
+
+  it('returns false when shift modifier is missing', () => {
```
SychevAndrey, line 193: I don't see a test where we have an extra modifier. Let's say we have a binding `mod+/ `and I press `shift+mod+/`. I believe in that case we should not trigger the binding action, let's add a test for it.

devnaumov (dismissed):

webapp/packages/core-view/src/View/parseHotkey.test.ts
sergeyteleshev: nice catch!

SychevAndrey (approved):

PR #4341, opened 2026-05-14
dbeaver/cloudbeaver#4138 feat: support inline editing of multi-line v…

### LINKED ISSUE #4138: The interface may be misleading and cause update problems
### Description

https://github.com/user-attachments/assets/1145e678-814c-4865-a2d2-abc44409b824

As you can see in the video, if a string contains special characters like \n, \r, etc..., copying and pasting the text from the box does not produce the same result if you click on the box or select the text.
In addition, if we modify the data from the interface box, the data is updated without taking into account the characters that are hidden and overwrites them.

### Steps to reproduce

- Create a table containing a string with \n, \r etc...
- Copy paste its content by clicking on the box and selecting the text to see the difference
- Edit cell content

### Expected/Desired Behavior

The copy and paste should be constant and reflect the reality of what is contained in the box.
A modification should not be able to overwrite the characters that are hidden.

### CloudBeaver Version

25.3.4

### Additional context

_No response_

SychevAndrey:
…alues in data grid cells

- TextEditor: switch <input> to <textarea rows=1> so newlines are preserved when editing. Suppress the trailing insertLineBreak from the Enter that opens the editor (would otherwise wipe the auto-selected value).
- ResultSetFormatAction.truncateText: substitute \n, \r, \r\n, \t with visible glyphs (↵, ␍, →) and collapse other control chars to a space in a single pass.

Perf note: the new replace should be faster than the prior split('') + map + join. No intermediate array allocations, and the callback runs only on matched chars (which are sparse in display text). \p{Cc} also expands coverage from charCode < 32 to include DEL and C1 controls (U+007F, U+0080–U+009F). (closer to Dbeaver behavior)

### DIFF
--- a/webapp/common-react/@dbeaver/react-data-grid/src/editors/TextEditor.module.css
+++ b/webapp/common-react/@dbeaver/react-data-grid/src/editors/TextEditor.module.css
@@ -1,5 +1,6 @@
 :global(.rdg-cell) .editor {
   appearance: none;
+  resize: none;
 
   box-sizing: border-box;
   min-height: initial;
--- a/webapp/common-react/@dbeaver/react-data-grid/src/editors/TextEditor.tsx
+++ b/webapp/common-react/@dbeaver/react-data-grid/src/editors/TextEditor.tsx
@@ -1,6 +1,6 @@
 /*
  * CloudBeaver - Cloud Database Manager
- * Copyright (C) 2020-2025 DBeaver Corp and others
+ * Copyright (C) 2020-2026 DBeaver Corp and others
  *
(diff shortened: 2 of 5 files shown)


SychevAndrey:
https://github.com/user-attachments/assets/fa34c6cd-d3f7-4b1a-b93d-15d9455d532a

sergeyteleshev (dismissed):

devnaumov (dismissed):

devnaumov (approved):

sergeyteleshev (approved):

PR #4350, opened 2026-05-19
dbeaver/pro#9131 use visual keys for copy & paste

devnaumov:
closes [9131](https://github.com/dbeaver/pro/issues/9131)

Added visual keys in the view action. They are the same as before but now include pinned logic. Pinned visual keys should not affect anything except copy and paste for now. sql generation and other processes should continue using the usual keys.

### DIFF
--- a/webapp/packages/plugin-data-viewer/src/DatabaseDataModel/Actions/Grid/GridViewAction.ts
+++ b/webapp/packages/plugin-data-viewer/src/DatabaseDataModel/Actions/Grid/GridViewAction.ts
@@ -73,12 +73,29 @@ export class GridViewAction<
     return this.columnsOrder.map(index => ({ index }));
   }
 
+  get visualColumnKeys(): IGridColumnKey[] {
+    return this.columnKeys.slice().sort((a, b) => {
+      const aPinned = this.isColumnPinned(a);
+      const bPinned = this.isColumnPinned(b);
+
+      if (aPinned === bPinned) {
+        return 0;
+      }
+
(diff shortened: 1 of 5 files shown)


webapp/packages/plugin-data-viewer/src/DatabaseDataModel/Actions/Grid/GridViewAction.ts
```diff
@@ -73,6 +73,19 @@ export class GridViewAction<
     return this.columnsOrder.map(index => ({ index }));
   }
 
+  get visualColumnKeys(): IGridColumnKey[] {
+    return this.columnKeys.slice().sort((a, b) => {
```
SychevAndrey, line 77: Let's say we have cols 1,2,3,4,5,6.

We pin 4 an 5.

What we get from columnKeys and what will be the visualColumnKeys result?

Then we unpin and pin 5th first, then 4th.

What will be the result? Seems like it will be the same to the first one.

SychevAndrey (approved):

webapp/packages/plugin-data-viewer/src/DatabaseDataModel/Actions/Grid/GridViewAction.ts
devnaumov: When we pin columns, their order depends on their position in the table. So if we pin column 3 and then column 1, the order will be 3, 1 (current devel behavior). We will likely change this later

sergeyteleshev (approved):

PR #4354, opened 2026-05-20
dbeaver/pro#8486 feat: add DataReadonlyReason API for data sources

SychevAndrey:
Previously TableStatusIndicator relied on model.isReadonly() which lost reasons. This caused the "Read-only connection" tooltip to show in grouping panel even on editable connections — the connection wasn't read-only, the presentation was.

Each data source now declares its specific reason via getReadonlyReason(). UI maps the typed reason to an accurate message

### DIFF
--- a/webapp/packages/plugin-data-viewer/src/DatabaseDataModel/IDatabaseDataSource.ts
+++ b/webapp/packages/plugin-data-viewer/src/DatabaseDataModel/IDatabaseDataSource.ts
@@ -43,6 +43,12 @@ export enum DatabaseDataAccessMode {
   Default,
   Readonly,
 }
+export enum DataReadonlyReason {
+  /** The data source does not support editing (grouping panel). */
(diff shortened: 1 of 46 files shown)


sergeyteleshev (dismissed):

webapp/packages/plugin-data-viewer/src/DatabaseDataModel/IDatabaseDataSource.ts
```diff
@@ -43,6 +43,12 @@ export enum DatabaseDataAccessMode {
   Default,
   Readonly,
 }
+export enum DataReadonlyReason {
```
devnaumov: Let’s try using `DatabaseDataFeature` to determine the model’s read-only reason (Grouping). This will allow us to keep the logic inside TableStatusIndicator, which is the only place where it’s needed.

devnaumov (approved):

PR #4355, opened 2026-05-20
dbeaver/pro#9095 load active connections before checking execution co…

devnaumov:
closes [9095](https://github.com/dbeaver/pro/issues/9095)

Some of the data sources return an execution context only if a connection exists at the resource level (datasets, resources, file system, etc.). If no one in the app has loaded the connection, the execution context will always be undefined.

```
 get executionContext(): IConnectionExecutionContextInfo | undefined {
    if (
      !this.state.executionContext ||
      !this.connectionInfoResource.has(createConnectionParam(this.state.executionContext.projectId, this.state.executionContext.connectionId))
    ) {
      return undefined;
    }
    return this.state.executionContext;
  }
```

Since datasets are restored at the SQL tab level, I moved the load function up so that we load the connection before checking the execution context. This ensures that the restore process works as expected without side effects from the app.

The only downside is that if every sql tab lacks an execution context, we will load the active project connections. However, this scenario is very unlikely as it would require the UI to be in the VIEWER role (and there datasets have connections anyway), where only the restore behavior load connections (we dont have top bar actions, navigation tree and etc)

### DIFF
--- a/webapp/packages/plugin-sql-editor-navigation-tab/src/SqlEditorTabService.ts
+++ b/webapp/packages/plugin-sql-editor-navigation-tab/src/SqlEditorTabService.ts
@@ -314,11 +314,11 @@ export class SqlEditorTabService extends Bootstrap {
     }
 
     const dataSource = this.sqlDataSourceService.create(tab.handlerState, tab.handlerState.datasourceKey);
+
+    await this.connectionInfoResource.load(ConnectionInfoActiveProjectKey);
     const executionContext = dataSource.executionContext;
 
     if (executionContext) {
-      await this.connectionInfoResource.load(ConnectionInfoActiveProjectKey);
-
       const contextConnection = createConnectionParam(executionContext.projectId, executionContext.connectionId);
 
       if (!this.connectionInfoResource.has(contextConnection)) {


webapp/packages/plugin-sql-editor-navigation-tab/src/SqlEditorTabService.ts
```diff
@@ -314,11 +314,11 @@ export class SqlEditorTabService extends Bootstrap {
     }
 
     const dataSource = this.sqlDataSourceService.create(tab.handlerState, tab.handlerState.datasourceKey);
+
+    await this.connectionInfoResource.load(ConnectionInfoActiveProjectKey);
```
sergeyteleshev, line 318: maybe we can try more local solution here?

just load the connection info with this key where the current issue appears:
[https://github.com/dbeaver/cloudbeaver/blob/2404d1543cfdd448798ca4197beeebe7a21896[…]ebapp/packages/plugin-sql-editor/src/SqlEditorConnectionBar.tsx](https://github.com/dbeaver/cloudbeaver/blob/2404d1543cfdd448798ca4197beeebe7a2189622/webapp/packages/plugin-sql-editor/src/SqlEditorConnectionBar.tsx#L41-L45)

sergeyteleshev (approved):

PR #4358, opened 2026-05-21
[CB] Menu is displayed under opened forms in the Administration page

sergeyteleshev:
closes https://github.com/dbeaver/pro/issues/9362

The main issue was a conflict between the portal dialog and z-index, so I removed the custom handling and let AriaKit manage it entirely.

The notifications were also updated because QA requested handling for a case where notifications remained clickable after a modal dialog was shown, regardless of our previous attempts to prevent it.

While making these changes, I discovered another bug related to the switch thumbs. They were absolutely positioned and used z-index, which caused them to appear above modal dialogs. I fixed this by keeping them within the layout where they are rendered.

https://github.com/user-attachments/assets/a51d51d9-e010-4227-abd1-2bbae4f97103
https://github.com/user-attachments/assets/7554ceb4-c04b-440b-8e92-d49ce8e56c73

### DIFF
--- a/server/bundles/io.cloudbeaver.model/src/io/cloudbeaver/WebSessionProjectImpl.java
+++ b/server/bundles/io.cloudbeaver.model/src/io/cloudbeaver/WebSessionProjectImpl.java
@@ -1,6 +1,6 @@
 /*
  * DBeaver - Universal Database Manager
- * Copyright (C) 2010-2025 DBeaver Corp and others
+ * Copyright (C) 2010-2026 DBeaver Corp and others
  *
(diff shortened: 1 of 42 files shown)


SychevAndrey (dismissed):

devnaumov (dismissed):

SychevAndrey (changes requested):
We can't use the same set of persisted elements for the common dialog. SlideDialog is a bit special.

SychevAndrey (dismissed):

devnaumov (dismissed):

devnaumov (approved):

SychevAndrey (approved):

PR #4359, opened 2026-05-22
dbeaver/pro#8684 [CB] feat: add autoscroll when dragging selection

SychevAndrey:
https://github.com/user-attachments/assets/4a7414ab-0747-4cc5-95c1-5167e6db6ddb

### DIFF
--- a/webapp/packages/plugin-data-spreadsheet-new/src/DataGrid/useGridAutoScroll.ts
+++ b/webapp/packages/plugin-data-spreadsheet-new/src/DataGrid/useGridAutoScroll.ts
@@ -5,7 +5,7 @@
  * Licensed under the Apache License, Version 2.0.
  * you may not use this file except in compliance with the License.
  */
-import { useEffect, useRef, useState } from 'react';
+import { useEffect, useState } from 'react';
 import { getCellLookupPoint } from './helpers/getCellLookupPoint.js';
 import { getEdgeSpeed } from './helpers/getEdgeSpeed.js';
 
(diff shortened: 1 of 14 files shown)


webapp/packages/plugin-data-spreadsheet-new/src/DataGrid/useGridAutoScroll.ts
```diff
@@ -0,0 +1,111 @@
...
+export interface IGridAutoScroll {
+  update: (container: HTMLElement, cursor: IMousePosition) => void;
+  stop: () => void;
+}
+
+interface IGridAutoScrollController extends IGridAutoScroll {
+  setOnScroll: (onScroll: (cellLookupPoint: IMousePosition) => void) => void;
+}
+
+function createGridAutoScrollController(): IGridAutoScrollController {
```
devnaumov, line 26: It seems react data grid already has this functionality built in, did you check their api? It should be possible to use their logic instead of reimplementing it ourselves.
https://comcast.github.io/react-data-grid/#/AllFeatures

webapp/packages/plugin-data-spreadsheet-new/src/DataGrid/useGridAutoScroll.ts
```diff
@@ -0,0 +1,111 @@
...
+  }
+
+  function setOnScroll(nextOnScroll: (cellLookupPoint: IMousePosition) => void): void {
+    onScroll = nextOnScroll;
+  }
+
+  return { update, stop, setOnScroll };
+}
+
+export function useGridAutoScroll(onScroll: (cellLookupPoint: IMousePosition) => void): IGridAutoScroll {
```
devnaumov, line 100: I think we should do the following. On cell select, run scrollToCell or scrollToRow from react data grid api, thats should be enough

devnaumov (changes requested):

PR #4362, opened 2026-05-26
dbeaver/pro#8684 feat: unlock autoscroll when dragging in data grid

SychevAndrey:


webapp/packages/plugin-data-spreadsheet-new/src/DataGrid/DataGridTable.css
```diff
@@ -51,6 +70,11 @@
     & .rdg-cell-custom-deleted {
       background-color: var(--theme-delete-status) !important;
     }
+
+    & .rdg-editor-container textarea::selection {
```
sergeyteleshev, line 75: today it is textarea, tomorrow it can be text input again
so if someone change it in `TextEditor.tsx` and forget we have these styles here we will have a regression. so I suggest to replace `textarea` with unique classname

sergeyteleshev (approved):

devnaumov (approved):

PR #4373, opened 2026-06-01
[CB] Migrate Switch component to ui-kit

sergeyteleshev:
closes https://github.com/dbeaver/pro/issues/8780


webapp/common-react/@dbeaver/ui-kit/src/Switch/Switch.tsx
```diff
@@ -0,0 +1,44 @@
...
+ */
+import clsx from 'clsx';
+import './Switch.css';
+
+export interface SwitchBaseProps extends Omit<React.InputHTMLAttributes<HTMLInputElement>, 'type'> {
+  className?: string;
+  children?: React.ReactNode;
+}
+
+export function SwitchBase({ checked, disabled, children, className, id, onChange, ...inputRest }: SwitchBaseProps): React.ReactElement {
```
SychevAndrey: This component is hard to use accessibly by default. The text passed as children is not connected to the input, so many usages will create a switch without an accessible name. Look at storybook examples. This PR adds a11y connection in core-blocks layer, but it should be a part of a ui-kit.

Please make labeling built-in, for example by rendering a label element tied to the input, or by providing a clear label slot/prop that wires aria-labelledby automatically.

webapp/packages/storybook/src/stories/components/Switch/switch.stories.tsx
```diff
@@ -0,0 +1,158 @@
...
+          '--dbv-kit-switch-track-radius': '8px',
+          '--dbv-kit-switch-thumb-size': '12px',
+          '--dbv-kit-switch-thumb-offset': '2px',
+          '--dbv-kit-switch-checked-offset': '14px',
+        } as React.CSSProperties
+      }
+    >
+      <Switch>Dense unchecked</Switch>
+      <Switch checked>Dense checked</Switch>
+      <Switch disabled>Dense disabled</Switch>
```
SychevAndrey, line 69: SwitchBase is currently mostly style-driven customization. Consumers cannot change internal composition (track/thumb/content) in a clean way. Could we expose compositional parts (for example Root, Input, Track, Thumb, Label)?

webapp/common-react/@dbeaver/ui-kit/src/Switch/Switch.tsx
```diff
@@ -0,0 +1,44 @@
...
+          'dbv-kit-switch__control--disabled': disabled,
+        })}
+      >
+        <div className="dbv-kit-switch__track" />
+        <input
+          {...inputRest}
+          type="checkbox"
+          id={id}
+          role="switch"
+          aria-checked={checked}
```
SychevAndrey: aria-checked is taken from checked prop directly. In uncontrolled usage this can become inconsistent. Look at Checkbox as an example

SychevAndrey (changes requested):
Good job! Let's make it even better by improving a11y and using Composition pattern

webapp/common-react/@dbeaver/ui-kit/src/Switch/Switch.tsx
sergeyteleshev: added tied to the input label

webapp/packages/storybook/src/stories/components/Switch/switch.stories.tsx
sergeyteleshev: I'm not sure we need that. What is the use case?

SychevAndrey: It was not critical in CE, but we definitely need that in UI-library that is supposed to be used in different products. We don't want to refactor a common UI component when we have the use case

webapp/common-react/@dbeaver/ui-kit/src/Switch/Switch.tsx
```diff
@@ -0,0 +1,77 @@
...
+  return (
+    <Switch.Label className={clsx('dbv-kit-switch', className)}>
+      <Switch.Control
+        className={clsx({
+          'dbv-kit-switch__control--checked': checked,
+          'dbv-kit-switch__control--disabled': disabled,
+        })}
+      >
+        <Switch.Track />
+        <Switch.Input
```
SychevAndrey: I like the new shape of a component, however here is an issue with Control component. You set classnames for Control component that affects other component like Thumb via CSS selectors:

```
.dbv-kit-switch__control--checked .dbv-kit-switch__thumb {
    margin-left: var(--dbv-kit-switch-checked-offset);
  }
```

So if I decide to create my own version of a Switch, I will need to now which class to set and it's not stated anywhere in component interface or whatever. We do not have a good example in our ui-kit because usually we reuse ariakit structure like:

```
<ComponentProvider>
    <ComponentAtom />
    <ComponentChild />
</ComponentProvider>
```

In this structure you should have the ability to use a provider, set properties and see how component changes it's behaviour without any details of implementation. For example `@dbeaver/ui-kit/src/Dialog/Dialog.tsx` or `https://ariakit.com/components/radio`

And using this approach we could also generate an Id and use to link input and label and make them freely positioned without wrapping the input to the label by default.

SychevAndrey (changes requested):

SychevAndrey (dismissed):

devnaumov (dismissed):

SychevAndrey (approved):

PR #4381, opened 2026-06-04
dbeaver/pro#8333 fix: Eliminate React 19 error in useEditorDefaultExtensions

SychevAndrey:
replace useRef with useState for managing editor options state


sergeyteleshev (dismissed):

webapp/packages/plugin-codemirror6/src/useEditorDefaultExtensions.ts
```diff
@@ -119,16 +119,18 @@ const DEFAULT_EXTENSIONS_COMPARTMENT = new Compartment();
...
+    options,
+    result: createExtensions(options),
+  }));
 
-  if (isOptionsChanged || extensions.current === null) {
-    previousOptions.current = options;
-    extensions.current = createExtensions(options);
+  if (!isObjectsEqual(options, state.options)) {
+    const result = createExtensions(options);
+    setState({ options, result });
```
devnaumov: this will cause extra rerenders, its better to comment such code parts as expected than fix it this way

devnaumov (changes requested):

webapp/packages/plugin-codemirror6/src/useEditorDefaultExtensions.ts
SychevAndrey: how many? how will it affect the product?

devnaumov: Props are changed -> rerender, then if options are different, you are setting state that rerenders one more time and its just a bad react practice. You are executing state update from component body to sync state from props, it might cause infinite rerenders and other problems. I think that the prev approach is cleaner and less buggy. If we know what we are doing, we dont need to fix it. I suggest to disable this eslint rule as this pattern is commonly used across react ecosystem, what do you think?

devnaumov (changes requested):

PR #4386, opened 2026-06-05
[CB] Create project form with options tab

sergeyteleshev:
closes https://github.com/dbeaver/pro/issues/9525


webapp/packages/plugin-project-info/src/ProjectInfoForm/Options/ProjectInfoFormOptions.tsx
```diff
@@ -0,0 +1,50 @@
...
+          </InputField>
+          <InputField name="id" value={project?.id ?? ''} readOnly>
+            {translate('plugin_project_info_form_options_field_id')}
+          </InputField>
+          {project?.description && (
+            <Textarea name="description" value={project.description} readOnly>
+              {translate('plugin_project_info_form_options_field_description')}
+            </Textarea>
+          )}
+          <InputField name="shared" value={project ? translate(project.shared ? 'ui_yes' : 'ui_no') : ''} readOnly>
```
devnaumov: Probable we should use checkboxes to show this info?

webapp/packages/plugin-project-info/src/ProjectInfoForm/Options/ProjectInfoFormOptionsPart.ts
```diff
@@ -0,0 +1,44 @@
...
+ */
+
+import type { IExecutionContextProvider } from '@cloudbeaver/core-executor';
+import type { ProjectInfo, ProjectInfoResource } from '@cloudbeaver/core-projects';
+import { FormPart, type IFormState } from '@cloudbeaver/core-ui';
+
+import type { IProjectInfoFormState } from '../IProjectInfoFormState.js';
+
+export interface IProjectInfoFormOptionsState {
+  projectInfo: ProjectInfo | null;
```
devnaumov: Probably we should just list properties here, why do we need extra level of projectInfo?

webapp/packages/plugin-project-info/src/ProjectInfoTreeMenuBootstrap.ts
```diff
@@ -0,0 +1,51 @@
...
+
+import { Bootstrap, injectable } from '@cloudbeaver/core-di';
+import { DATA_CONTEXT_NAV_NODE } from '@cloudbeaver/core-navigation-tree';
+import { ActionService, MenuService } from '@cloudbeaver/core-view';
+
+import { ACTION_PROJECT_INFO } from './actions/ACTION_PROJECT_INFO.js';
+import { ProjectInfoFormService } from './ProjectInfoForm/ProjectInfoFormService.js';
+
+@injectable(() => [ActionService, MenuService, ProjectInfoFormService])
+export class ProjectInfoTreeMenuBootstrap extends Bootstrap {
```
devnaumov: Please rename it to ProjectInfoBootstrap

webapp/packages/plugin-project-info/src/ProjectInfoTreeMenuBootstrap.ts
```diff
@@ -0,0 +1,51 @@
...
+      getItems: (context, items) => [...items, ACTION_PROJECT_INFO],
+    });
+
+    this.actionService.addHandler({
+      id: 'action-project-info-handler',
+      actions: [ACTION_PROJECT_INFO],
+      contexts: [DATA_CONTEXT_NAV_NODE],
+      isActionApplicable: context => {
+        const node = context.get(DATA_CONTEXT_NAV_NODE);
+        return !!node?.projectId;
```
devnaumov: please use isProjectNode fn also, it should be applied to the project node iitself

devnaumov (changes requested):

webapp/packages/plugin-project-info/src/ProjectInfoTreeMenuBootstrap.ts
sergeyteleshev: I did that on purpose. The feature is harder to find if it applies only for project node

webapp/packages/plugin-project-info/src/ProjectInfoForm/Options/ProjectInfoFormOptions.tsx
sergeyteleshev: Screenshot 2026-06-08 at 10 20 01

Screenshot 2026-06-08 at 10 20 19

webapp/packages/plugin-project-info/src/ProjectInfoTreeMenuBootstrap.ts
devnaumov: We want to keep navigation tree menu light, also we groomed it as action on the project node level so its fine if it will be only there

devnaumov (dismissed):

devnaumov (approved):

PR #4388, opened 2026-06-05
dbeaver/pro#9120 cloudbeaver conf generator

yagudin10:


apps/config-generator/pom.xml
```diff
@@ -0,0 +1,37 @@
...
+        <maven.compiler.source>21</maven.compiler.source>
+        <maven.compiler.target>21</maven.compiler.target>
+        <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
+    </properties>
+
+    <dependencies>
+        <dependency>
+            <groupId>com.google.code.gson</groupId>
+            <artifactId>gson</artifactId>
+            <version>2.13.2</version>
```
serge-rider, line 21: Use deps from dbeaver-commmon

config/template/cloudbeaver-base.conf
```diff
@@ -0,0 +1,271 @@
+{
```
serge-rider, line 1: Lets keep template somewhere in /config, not in tool sources

serge-rider (changes requested):

PR #4405, opened 2026-06-17
dbeaver/pro#9588 remove top margin

devnaumov:
closes [9588](https://github.com/dbeaver/pro/issues/9588)


webapp/common-react/@dbeaver/react-data-grid/src/formatters/BooleanFormatter.tsx
```diff
@@ -31,7 +31,7 @@ export function BooleanFormatter({ value, className, onClick, onKeyDown, focusab
       onKeyDown={handleKeyDown}
       onClick={onClick}
     >
-      {value === null ? <NullFormatter /> : <CheckboxIndicator size="small" checked={value} />}
+      {value === null ? <NullFormatter /> : <CheckboxIndicator className="tw:mt-0!" size="small" checked={value} />}
```
sergeyteleshev: checkbox automatically adjusts to the center in every part of the app fine, but something went wrong here:
`margin-top: calc(var(--dbv-kit-checkbox-height) * (calc(1 - var(--dbv-kit-checkbox-check-size)) / 2))`

maybe delete `tw:items-center` for the container and fix `NullFormatter` styles if something is broken for it?

sergeyteleshev (dismissed):

webapp/common-react/@dbeaver/react-data-grid/src/formatters/BooleanFormatter.tsx
devnaumov: There should not be margin top on checkbox level and on components in general, margins should be applied in the place of use by parent components of via classname. Otherwise it will break layout

sergeyteleshev (approved):

PR #4406, opened 2026-06-17
dbeaver/pro#8796 [CB] create a separate plugin for the execution plan

SychevAndrey:
also added a linter rule for interface names


webapp/packages/plugin-sql-editor-execution-plan/src/SqlExecutionPlanResultBootstrap.ts
```diff
@@ -0,0 +1,32 @@
...
+    private readonly sqlResultTabsService: SqlResultTabsService,
+    private readonly sqlExecutionPlanService: SqlExecutionPlanService,
+  ) {
+    super();
+  }
+
+  override register(): void {
+    this.sqlResultTabsService.resultPanels.add(SqlExecutionPlanPanel, 0, ({ tabId }) => !this.sqlExecutionPlanService.has(tabId));
+
+    this.sqlResultTabsService.onResultTabClose.addHandler(({ state, tabId }) => {
```
devnaumov, line 24: its better to add handlers on the constructor level

webapp/packages/plugin-sql-editor/src/ISqlEditorTabState.ts
```diff
@@ -45,15 +45,6 @@ export const SQL_EDITOR_RESULT_TAB_SCHEMA = schema.object({
 
 export type ISqlEditorResultTab = schema.infer<typeof SQL_EDITOR_RESULT_TAB_SCHEMA>;
 
-export const EXECUTION_PLAN_TAB_SCHEMA = schema.object({
```
devnaumov, line 48: could you clarify how handleTabRestore is working now in terms of executionPlanTabs

devnaumov (dismissed):

webapp/packages/plugin-sql-editor/src/ISqlEditorTabState.ts
SychevAndrey: Before we used to restore tabs and then clear the state. That's why handleTabRestore looks weird. But now it happens on it's own because we keep tabs only in the in-memory Map. Anyways, result is the same: tabs don't restore after reload

sergeyteleshev (dismissed):

devnaumov (approved):

PR #4412, opened 2026-06-19
dbeaver/pro#9652 refactor ref api

HocKu7:
Closes https://github.com/dbeaver/pro/issues/9652

CleanShot 2026-06-24 at 23 09 13@2x


serge-rider (dismissed):

server/bundles/io.cloudbeaver.server/src/io/cloudbeaver/service/sql/WebSQLDataFilter.java
```diff
@@ -92,10 +94,15 @@ public String getWhere() {
         return where;
     }
 
+    public boolean isAnyConstraint() {
```
serge-rider, line 100: What is this? Add javadoc pls

server/bundles/io.cloudbeaver.server/schema/service.sql.graphqls
```diff
@@ -68,19 +68,37 @@ input SQLDataFilter {
...
+    sourceColumnIndex: Int!
+    "Column name in the source result set"
+    sourceColumnName: String!
+    "Column name in the target entity"
+    targetColumnName: String!
+}
+
 type SQLResultReference @since(version: "26.1.1") {
+    "True for reverse references (incoming FKs)"
+    isReference: Boolean!
```
serge-rider: isReference -> reference

server/bundles/io.cloudbeaver.server/schema/service.sql.graphqls
```diff
@@ -68,19 +68,37 @@ input SQLDataFilter {
...
 
     constraints: [ SQLDataFilterConstraint ]
+    "When true constraints are combined with OR instead of AND"
+    anyConstraint: Boolean
     where: String
     orderBy: String
 }
 
+"Pair (source-row column index, target-entity column name) used to build a navigation filter"
+type SQLReferenceColumnMapping @since(version: "26.1.1") {
```
serge-rider: 26.1.2

serge-rider (changes requested):

webapp/packages/plugin-data-viewer-references/src/DataViewerReferencesBootstrap.ts
```diff
@@ -0,0 +1,45 @@
...
+
+  override register(): void {
+    this.dataPresentationService.add({
+      id: 'references-presentation',
+      type: DataPresentationType.toolsPanel,
+      title: 'references',
+      icon: '/icons/plugin_data_viewer_result_set_grouping_m.svg',
+      dataFormat: ResultDataFormat.Resultset,
+      hidden: (dataFormat, model, resultIndex) => {
+        const source = model.source as any;
```
SychevAndrey: please don't use `any`

webapp/packages/plugin-data-viewer-references/src/DataViewerReferencesPresentation.tsx
```diff
@@ -0,0 +1,122 @@
...
+} from '@cloudbeaver/plugin-data-viewer';
+
+import type { IDataViewerReferencesPresentationState } from './IDataViewerReferencesState.js';
+import { useReferencesDataModel } from './useReferencesDataModel.js';
+
+export const DataViewerReferencesPresentation: DataPresentationComponent = observer(function DataViewerReferencesPresentation({
+  model: unknownModel,
+  resultIndex,
+}) {
+  const originalModel = unknownModel as any;
```
SychevAndrey: if it's unknown, maybe better to use `unknown` type?

webapp/packages/plugin-data-viewer-references/src/DataViewerReferencesPresentation.tsx
```diff
@@ -0,0 +1,122 @@
...
+
+  const model = useReferencesDataModel(originalModel, resultIndex, state);
+  const associations = referencesAction.associations;
+  const defaultAssociation = associations[0];
+
+  if (!state.association && defaultAssociation) {
+    state.association = defaultAssociation.associationName;
+  }
+
+  //@TODO Find a better way to check whether model is ready to be used
```
SychevAndrey: is there a separate task or will be fixed here?

webapp/packages/plugin-data-viewer-references/src/useReferencesDataModel.ts
```diff
@@ -0,0 +1,208 @@
...
+
+              const constraints: SqlDataFilterConstraint[] = [];
+
+              for (const row of rows) {
+                const rowValue = data.getRowValue(row);
+
+                if (rowValue) {
+                  const targetValue = sourceColumn ? rowValue[sourceColumn.position] : undefined;
+
+                  if (targetValue && operation) {
```
SychevAndrey: since target value is just any value it can also be 0 or false. then contraint will not be pushed.

webapp/packages/plugin-data-viewer-references/src/useReferencesDataModel.ts
```diff
@@ -0,0 +1,208 @@
...
+                    constraints.push({
+                      attributeName: currentReference.isReference ? targetMapping?.targetColumnName : targetColumn?.name,
```
SychevAndrey: maybe we could move it to some helper with a good naming/comment? because it's hard to understand what is going on here.

webapp/packages/plugin-data-viewer-references/src/useReferencesDataModel.ts
```diff
@@ -0,0 +1,208 @@
...
+        prevStateRef.current = { association, sourceResultId, activeRows };
+
+        if (association && sourceResultId) {
+          const executionContext = sourceModel.source.executionContext;
+          model.source.setExecutionContext(executionContext).setSupportedDataFormats(connectionInfo?.supportedDataFormats ?? []);
+          const context = executionContext?.context;
+
+          if (context) {
+            const connectionKey = createConnectionParam(context.projectId, context.connectionId);
+            const references = referencesAction.associations;
```
SychevAndrey: let's stick to one naming. in `webapp/packages/plugin-data-viewer-references/src/DataViewerReferencesPresentation.tsx `it's called _associations_.

webapp/packages/plugin-data-viewer-references/src/useReferencesDataModel.ts
```diff
@@ -0,0 +1,208 @@
...
+      state.modelId = model.id;
+
+      model.setAccess(DatabaseDataAccessMode.Readonly).setCountGain(dataViewerSettingsService.getDefaultRowsCount()).setSlice(0);
+
+      return {
+        source,
+        model,
+        dispose() {
+          this.model.dispose();
+          tableViewerStorageService.remove(this.model.id);
```
SychevAndrey, line 101: it's hard to track what is going on with state.modelId and this.model.id.

We mutate state and we use different ids for dispose? is it for purpose? Shouldn't they be the same?

SychevAndrey (changes requested):

webapp/packages/plugin-data-viewer-references/src/DataViewerReferencesPresentation.tsx
```diff
@@ -0,0 +1,122 @@
...
+
+  const currentAssociation = associations.find(a => a.associationName === state.association);
+
+  if (!currentAssociation) {
+    return <TextPlaceholder>{translate('plugin_data_viewer_references_no_reference')}</TextPlaceholder>;
+  }
+
+  function openAssociation() {
+    if (!currentAssociation?.targetNodePath) {
+      throw new Error('Target node path is not defined');
```
sergeyteleshev: this things like `throw new Error` on layout level usually breaks a whole layout. did you test this case? what will happen? I suppose like notification or StatusMessage is more than enough in most cases

webapp/packages/plugin-data-viewer-references/src/DataViewerReferencesPresentation.tsx
```diff
@@ -0,0 +1,122 @@
...
+    <CaptureViewScope>
+      <div className="tw:flex tw:flex-col tw:h-full tw:gap-2 tw:bg-(--theme-secondary)">
+        <div className="tw:flex tw:gap-2 tw:items-center">
+          <Select
+            className="tw:flex-1"
+            state={state}
+            name="association"
+            items={associations.map(a => a.associationName)}
+            valueSelector={a => {
+              const assotiation = referencesAction.getAssociation(a);
```
sergeyteleshev: typo

webapp/packages/plugin-data-viewer-references/src/DataViewerReferencesPresentation.tsx
```diff
@@ -0,0 +1,122 @@
...
+
+  return (
+    <CaptureViewScope>
+      <div className="tw:flex tw:flex-col tw:h-full tw:gap-2 tw:bg-(--theme-secondary)">
+        <div className="tw:flex tw:gap-2 tw:items-center">
+          <Select
+            className="tw:flex-1"
+            state={state}
+            name="association"
+            items={associations.map(a => a.associationName)}
```
sergeyteleshev: is it guaranteed that association names are unique? if so no issues here. otherwise it is collisions may appear and the whole feature can have a bug

webapp/packages/plugin-data-viewer-references/src/DataViewerReferencesPresentation.tsx
```diff
@@ -0,0 +1,122 @@
...
+              }
+
+              return `${assotiation.targetEntityName} (${assotiation.associationName})`;
+            }}
+          />
+          {currentAssociation.targetNodePath && (
+            <Button size="small" variant="secondary" onClick={openAssociation}>
+              {translate('ui_open')}
+            </Button>
+          )}
```
sergeyteleshev, line 105: lets always show the button so the interface don't jump
but disable it if no target node path or something is loading

webapp/packages/plugin-data-viewer-references/src/DataViewerReferencesDataSource.ts
```diff
@@ -0,0 +1,25 @@
...
+
+export class DataViewerReferencesDataSource extends ContainerDataSource<IDataContainerOptions> {
+  constructor(
+    serviceProvider: IServiceProvider,
+    connectionExecutionContextService: ConnectionExecutionContextService,
+    graphQLService: GraphQLService,
+    asyncTaskInfoService: AsyncTaskInfoService,
+  ) {
+    super(serviceProvider, graphQLService, asyncTaskInfoService, connectionExecutionContextService);
+    this.setFeature(DatabaseDataFeature.References);
```
sergeyteleshev, line 23: i can see that it is not reused any where (only few places to show/hide the feature itself). this means reference panel table cells probably has no actions can be applied via context menu or shortcuts

is this expected?

since it is `ResultSet` I would also try to add `References` in the implementations of `ResultSet` with `allowedFeatures` arrays. and check if features works correct here. if something is broken we can remove it or comment + add the ticket to fix it

the point is to add a functionality from the start what we can afford, without going too deep into it right now

webapp/packages/plugin-data-viewer-references/src/LocaleService.ts
```diff
@@ -0,0 +1,30 @@
...
+  override register(): void {
+    this.localizationService.addProvider(this.provider.bind(this));
+  }
+
+  private async provider(locale: string) {
+    switch (locale) {
+      case 'ru':
+        return (await import('./locales/ru.js')).default;
+      default:
+        return (await import('./locales/en.js')).default;
```
sergeyteleshev, line 27: few more locales would be nice

webapp/packages/plugin-data-viewer-references/src/useReferencesDataModel.ts
```diff
@@ -0,0 +1,208 @@
...
+        },
+      };
+    },
```
sergeyteleshev, line 104: this is hard to read

i think we can split it to 2 functions:

`retrieveModelFromCache()`
`constructModel()`

webapp/packages/plugin-data-viewer-references/src/useReferencesDataModel.ts
```diff
@@ -0,0 +1,208 @@
...
+          const context = executionContext?.context;
+
+          if (context) {
+            const connectionKey = createConnectionParam(context.projectId, context.connectionId);
+            const references = referencesAction.associations;
+            const currentReference = references.find(r => r.associationName === association);
+
+            if (currentReference?.targetNodePath) {
+              const sourceColumn = data.columns.find(c => currentReference.columnMapping.find(m => m.sourceColumnName === c.name));
+              const targetColumn = data.columns.find(c => currentReference.columnMapping.find(m => m.targetColumnName === c.name));
```
sergeyteleshev: this is gonna be super slow for big tables
pls use `Set` to make `O(1)` for inner for cycle (`find`)

webapp/packages/plugin-data-viewer-references/src/useReferencesDataModel.ts
```diff
@@ -0,0 +1,208 @@
...
+            .setSupportedDataFormats([])
+            .clearError()
+            .setResults([]);
+        }
+      },
+      { fireImmediately: true, equals: isObjectsEqual, delay: 300 },
+    );
+
+    return sub;
+  }, [state, sourceModel, sourceResultIndex]);
```
sergeyteleshev, line 211: all in all hard to read and understand. some prettifying would be nice

webapp/packages/plugin-data-viewer-request-query-viewer/src/DataViewerRequestQueryViewerBootstrap.ts
```diff
@@ -29,7 +29,10 @@ export class DataViewerRequestQueryViewerBootstrap extends Bootstrap {
     this.tableHeaderService.tableHeaderPlaceholder.add(
       DataViewerRequestQueryViewer,
       Number.MIN_SAFE_INTEGER,
-      props => !isResultSetDataSource(props.model.source) || this.dataViewerRequestQueryViewerSettingsService.disabled,
+      props =>
+        !isResultSetDataSource(props.model.source) ||
+        this.dataViewerRequestQueryViewerSettingsService.disabled ||
+        props.model.source.hasFeature(DatabaseDataFeature.References),
```
sergeyteleshev, line 35: thats a bit weird we only hide `references` panel
maybe we should hide all `ContainerDataSources` here?

webapp/packages/plugin-data-viewer-result-set-grouping/src/GroupingDataSource.ts
```diff
@@ -53,6 +53,7 @@ export class GroupingDataSource extends QueryDataSource<IDataGroupingOptions> {
         limit,
         constraints: options.constraints,
         where: options.whereFilter || undefined,
+        anyConstraint: options.anyConstraint,
```
sergeyteleshev, line 56: why we have this field? does backend can understand it from `constraints` field?

webapp/packages/plugin-data-viewer/src/DatabaseDataModel/Actions/ResultSet/ResultSetReferencesAction.ts
```diff
@@ -0,0 +1,64 @@
...
+    super(source as unknown as IDatabaseDataSource<unknown, IDatabaseResultSet>, result as IDatabaseResultSet);
+    this.associations = [];
+
+    makeObservable(this, {
+      associations: observable.ref,
+    });
+
+    setTimeout(() => {
+      this.loadAssociations();
+    }, 0);
```
sergeyteleshev, line 44: smells like a workaround and potential bugs. why we have it here? is there a way to move it to a more proper place?

webapp/packages/plugin-data-viewer/src/DatabaseDataModel/Actions/ResultSet/ResultSetReferencesAction.ts
```diff
@@ -0,0 +1,64 @@
...
+    const { associations } = await this.graphQLService.sdk.getSqlResultAssociations({
+      resultsId: result.id,
+      projectId: result.projectId,
+      connectionId: result.connectionId,
+      contextId: result.contextId,
+    });
+
+    this.associations = associations;
+    return this.associations;
+  }
```
sergeyteleshev: seems like it is an `action`

webapp/packages/plugin-data-viewer/src/DatabaseDataModel/IDatabaseDataOptions.ts
```diff
@@ -16,5 +16,6 @@ export interface IDatabaseDataOptions {
   whereFilter: string;
   /** A complex object that can represent filters and sorting options of the result set */
   constraints: SqlDataFilterConstraint[];
+  anyConstraint?: boolean;
```
sergeyteleshev, line 19: not sure we need it

webapp/packages/plugin-data-viewer/src/DatabaseDataModel/IDatabaseDataSource.ts
```diff
@@ -60,6 +60,8 @@ export enum DatabaseDataFeature {
   QueryResult = 'QueryResult',
   /** Data grouping and aggregation functionality. Used for GROUP BY operations. */
   Grouping = 'Grouping',
+  /** Displays incoming and outgoing table references for the selected table. */
+  References = 'References',
```
sergeyteleshev, line 64: if we can just reuse `ResultSet` it would be cool

sergeyteleshev (changes requested):

webapp/packages/plugin-data-viewer-references/src/DataViewerReferencesPresentation.tsx
devnaumov: Separate

webapp/packages/plugin-data-viewer-references/src/useReferencesDataModel.ts
devnaumov: Yes, we are syncing model id this way. The logic is the same with groping panel

webapp/packages/plugin-data-viewer-references/src/DataViewerReferencesPresentation.tsx
devnaumov: This case is never going to happen as we render open button only if currentAssociation is presented

webapp/packages/plugin-data-viewer-references/src/DataViewerReferencesPresentation.tsx
devnaumov: It would be strange, user will see a button but would not be able to click it. TargetNodePath should be provided as we are fetching result based on that

webapp/packages/plugin-data-viewer-references/src/DataViewerReferencesDataSource.ts
devnaumov: I dont follow the logic, we set feature here so we dont need to check instance of source, the same way we do for grouping panel

webapp/packages/plugin-data-viewer-references/src/useReferencesDataModel.ts
devnaumov: Logic is different now

webapp/packages/plugin-data-viewer-request-query-viewer/src/DataViewerRequestQueryViewerBootstrap.ts
devnaumov: Dont quite understand, request query feature works well with container data source we just hide it from references panel so we dont have header so it will look minimal as in dbeaver

webapp/packages/plugin-data-viewer-result-set-grouping/src/GroupingDataSource.ts
devnaumov: They cant, this flag means whether we need to glue constaints with AND or OR operator

webapp/packages/plugin-data-viewer/src/DatabaseDataModel/Actions/ResultSet/ResultSetReferencesAction.ts
devnaumov: Its should be loaded when we first access references action (contructor is executed), the same way we do in content action where we load quotas

webapp/packages/plugin-data-viewer/src/DatabaseDataModel/Actions/ResultSet/ResultSetReferencesAction.ts
devnaumov: Action is useful if you changing more than one observable properties so mobx can batch them in one update, in this case its only one so no difference

webapp/packages/plugin-data-viewer/src/DatabaseDataModel/IDatabaseDataOptions.ts
devnaumov: This flag existed on the backend before this feature, its how we work with constaints, you can combine them with AND or OR in terms of sql

webapp/packages/plugin-data-viewer/src/DatabaseDataModel/IDatabaseDataSource.ts
devnaumov: This feature was implemented so we dont check instance of source. No difference from grouping panel

webapp/packages/plugin-data-viewer-references/src/DataViewerReferencesPresentation.tsx
sergeyteleshev: I'm speaking from experience here. All of these cases we had this `throw new Error()` we had and which are never supposed to happen usually happens. Then we create a ticket for that and fix the issue with replacing this error with:
1. early return
2. log or notification or status message

I personally fixed at least 2-3 bugs like this during this year. So all I'm saying is after few iterations with this feature, this may happen. Right now we can fix it quite cheap since we noticed it pretty early

webapp/packages/plugin-data-viewer-references/src/DataViewerReferencesPresentation.tsx
sergeyteleshev: okay. my concern here is so user can't press it if something is loading and that there are not blinking interface
if you think its fine I'm good with that

webapp/packages/plugin-data-viewer-references/src/DataViewerReferencesDataSource.ts
sergeyteleshev: okay lets leave `DatabaseDataFeature.References` it is fine and expected
what about context menu and shortcuts? can we add something so its not empty? like generate sql or sorting if it is possible

webapp/packages/plugin-data-viewer-references/src/DataViewerReferencesPresentation.tsx
SychevAndrey: please link a task

webapp/packages/plugin-data-viewer-references/src/DataViewerReferencesDataSource.ts
devnaumov: We have sorting, I checked grouping panel and we have the same actions there. Probably we should create a ticket and add more actions to grouping panel and references panel

SychevAndrey (dismissed):

webapp/packages/plugin-data-viewer-references/src/DataViewerReferencesDataSource.ts
sergeyteleshev: okay lets go with the ticket then

sergeyteleshev (dismissed):

yagudin10 (dismissed):

sergeyteleshev (dismissed):

sergeyteleshev (approved):

PR #4419, opened 2026-06-22
dbeaver/pro#8899 use same model for updating result set

yagudin10:


server/bundles/io.cloudbeaver.server/src/io/cloudbeaver/service/sql/WebSQLResultsRow.java
```diff
@@ -53,4 +58,33 @@ public Map<String, Object> getUpdateValues() {
     public Map<String, Object> getMetaData() {
         return metaData;
     }
+
+    @Override
+    public int getRowNumber() {
+        return 0;
```
serge-rider: Why always 0? it has toi have row order number

server/bundles/io.cloudbeaver.server/src/io/cloudbeaver/service/sql/resultset/WebSQLDataStatementInfo.java
```diff
@@ -0,0 +1,38 @@
...
+
+    @NotNull
+    private final Object[] finalRow;
+
+    public WebSQLDataStatementInfo(@NotNull DBSEntity entity, @NotNull Object[] finalRow) {
+        super(entity);
+        this.finalRow = finalRow;
+    }
+
+    @NotNull
```
serge-rider, line 33: What is finalRow? You use it everywhere now and there no comments or explanations

serge-rider (changes requested):

PR #4426, opened 2026-06-24
dbeaver/pro#9612 api for connection types

yagudin10:


server/bundles/io.cloudbeaver.server/src/io/cloudbeaver/service/core/impl/WebServiceCore.java
```diff
@@ -321,6 +322,15 @@ public WebPropertyInfo[] getDriverProperties(
 
     }
 
+    @NotNull
+    @Override
+    public List<DBPConnectionType> getConnectionTypes(@NotNull WebSession webSession, @Nullable String id) {
+        return DataSourceProviderRegistry.getInstance().getConnectionTypes().stream()
+            .map(ct -> id == null || id.equals(ct.getId()) ? ct : null)
```
serge-rider, line 335: Use just filter instead of map + filter

server/bundles/io.cloudbeaver.server/schema/service.core.graphqls
```diff
@@ -633,6 +635,19 @@ type LogEntry {
...
+    "Unique identifier of the connection type"
+    id: ID!
+    "Human-readable name of the connection type"
+    name: String!
+    "Description of the connection type"
+    description: String
+    "Color of the connection type (for light theme)"
+    color: String!
+    "Alternative color of the connection type (for dark theme)"
+    alternativeColor: String
```
serge-rider: Lets add other boolean ct parameters here too (autocommit, confirmExecute, etc)

serge-rider (changes requested):

PR #4427, opened 2026-06-25
dbeaver/pro#9586 adjust validation

devnaumov:
closes [9586](https://github.com/dbeaver/pro/issues/9586)


webapp/packages/plugin-data-viewer/src/DatabaseDataModel/Actions/ResultSet/ResultSetDataContentAction.ts
```diff
@@ -121,7 +121,8 @@ export class ResultSetDataContentAction extends DatabaseDataAction<any, IDatabas
       return false;
     }
 
-    return (content.contentLength ?? 0) > limit;
+    const loadedBlob = this.retrieveBlobFromCache(holder.key);
```
sergeyteleshev: I would add this in the early return above. it is more clear what its about

if blob or text is already in the cache - boom - it is not truncated then

sergeyteleshev (dismissed):

sergeyteleshev (dismissed):

SychevAndrey (dismissed):

PR #4439, opened 2026-06-30
dbeaver/pro#4425 [CB] Confirmation for unsaved changes in administration

SychevAndrey:
Shows a unified **Save / Don't save / Cancel** dialog before you leave any admin page or close a create/edit form, so changes are never lost silently. It's built on a shared `confirmUnsavedChanges` helper and a `useUnsavedChanges` hook

To stop the admin drawer from getting out of sync when you press **Cancel**, `TabsState` gets an optional `controlledSelection` mode, where the selected tab follows the actual route instead of moving on click.

> controlledSelection is a temporary, scoped flag. Ideally currentTabId should always work this way (a proper controlled component), so the next step is to make that the default and remove the flag

image

https://github.com/user-attachments/assets/af490f9d-c72b-4f3d-8368-798e42553b7d


webapp/packages/core-ui/src/Screens/AppScreen/useUnsavedChanges.ts
```diff
@@ -0,0 +1,54 @@
...
+import { type IUnsavedChangesProvider, useObjectRef } from '@cloudbeaver/core-blocks';
+import { useService } from '@cloudbeaver/core-di';
+
+import { UnsavedChangesService } from './UnsavedChangesService.js';
+
+/**
+ * Registers the page form with the global unsaved-changes guard for as long as the component is mounted.
+ * Leaving the page (route transition) prompts a Save / Don't save / Cancel dialog when the provider is changed.
+ */
+export function useUnsavedChanges(provider: IUnsavedChangesProvider): void {
```
devnaumov: Please do not introduce new hooks and services that are responsible for showing notification about unsaved changes.

For cases such as driver creation, Identity Providers, Secret Management and and other places where we use options panel, use the same approach we use for options panel in connection and other dialogs. Add handler to the closeTask and check everything there. As general rule, use only handlers for resolving unsaved changes issues, prevent closing by interrupting context. 

All other cases should be resolved by using canDeActivate handler. Seems like we have regression in terms of unsaved changes notification when we are leaving server configuration page, you can find and fix this problem first

devnaumov (changes requested):

SychevAndrey:
Why I kept **confirmUnsavedChanges** 
It is not a new pattern. It still calls commonDialogService.open(ConfirmationDialog, …) and is invoked from the same canDeActivate / closeTask handlers, interrupting the context on cancel. It only removes the boilerplate inside those handlers.

1. It removes real duplication. The same code was copy-pasted across many places (Server Config, Domain Manager,  edit panels etc.). Now it becomes like that:

```
if (!(await confirmUnsavedChanges(commonDialogService, formState))) {
  ExecutorInterrupter.interrupt(contexts);
}
```

2. It fixes inconsistencies in translation between dialogs and improves UX

- titles/messages/buttons vary: `_ui_save_reminder+ui_are_you_sure_` vs `core_blocks_confirmation_dialog_title+administration_drivers_driver_unsaved_changes, confirm button ui_yes vs ui_close vs ui_processing_ok`
- most are two-way (**OK/Cancel**) where **OK** silently discards changes — there's no explicit **Save**; and the message is ambiguous
- the isSaving guard exists in some copies (Server Config, Domain Manager) and is missing in the edit panels. The helper encodes one correct behavior in one place

3. Hides the implementation. DialogueStateResult.Resolved, extraAction and other things that might be tricky for a developer to think about every time they want to add a confirmation dialog.  It returns a plain boolean, so the caller keeps full control of the interrupt. Enables save-loader, navigate only after.

4. One shape for three kinds of state. IUnsavedChangesProvider lets FormState, IEditableSettingsSource (Settings/Permissions), and other services be handled identically. 

5. Reasonable defaults, still customizable. Defaults for title/message/buttons live in one place; any call can override title/subTitle/message.

**Trade-offs** 
One extra abstraction.
Risk of becoming a really complex wrapper with many props if we keep adding options. (we can control it)
onConfirm prop added to ConfirmationDialog, so affects all those dialogs.

So, overall, I believe it's a good helper for our case and a nice feature for our users.

webapp/packages/core-blocks/src/CommonDialog/ConfirmationDialog.tsx
```diff
@@ -37,6 +37,10 @@ export interface ConfirmationDialogPayload {
   confirmActionText?: TLocalizationToken;
   cancelActionText?: TLocalizationToken;
   extraActionText?: TLocalizationToken;
+  /**
+   * Async action run when the confirm button is pressed. While it runs the confirm button shows a loader and is disabled. The dialog resolves only if it returns `true`; otherwise it is rejected (no confirm).
```
sergeyteleshev: fix linter pls in this file

webapp/packages/core-blocks/src/CommonDialog/ConfirmationDialog.tsx
```diff
@@ -37,6 +37,10 @@ export interface ConfirmationDialogPayload {
...
+   */
+  onConfirm?: () => Promise<boolean>;
```
sergeyteleshev, line 45: we should add onReject helper as well
plus it can be async or non-async as well for both callbacks

webapp/packages/core-ui/src/Tabs/TabsState.tsx
```diff
@@ -109,15 +109,16 @@ export const TabsState = observer(function TabsState<T = Record<string, any>>({
       selected,
       store,
       tabList,
+      currentTabId,
     },
   );
 
-  useEffect(() => {
-    if (isNotNullDefined(currentTabId)) {
+  useLayoutEffect(() => {
```
sergeyteleshev, line 116: Could you please clarify why have you changed it to `useLayoutEffect`?
I cannot see that we work directly with DOM here to compute something before repaint
`useLayoutEffect` is more heavy for computations so if there is no visual bugs/blinks I would suggest to revert it back to `useEffect`

webapp/packages/plugin-settings-administration/src/SettingsAdministrationPluginBootstrap.ts
```diff
@@ -23,8 +33,42 @@ export class SettingsAdministrationPluginBootstrap extends Bootstrap {
...
+
+    return confirmUnsavedChanges(this.commonDialogService, {
+      get isChanged() {
+        return edited.some(source => source.isEdited());
+      },
+      save: async () => {
+        try {
+          for (const source of edited) {
+            await source.save();
+          }
```
sergeyteleshev, line 59: I think we have 2 issues here:

1. If 1 resource is failed to save. It is gonna throw an error and thats it. The rest sources is not saved, but could've been 
2. We save sources sequentially which takes longer time to save it all

I suggest to use `Promise.all` or `Promise.allSettled` API to resolve both issues

webapp/packages/core-blocks/src/CommonDialog/ConfirmationDialog.tsx
SychevAndrey: on Reject usually we don't do async operations, and we have control in calling code because we handle rejection outside. I don't think we need to add method just in case. Let's wait until we need it.

webapp/packages/core-ui/src/Tabs/TabsState.tsx
SychevAndrey: This code fixes the regression that we got from migration to ariakit. selectedId is used to get updated during the render. Moving it to useEffect made ariakit unsync with the state

webapp/packages/plugin-settings-administration/src/SettingsAdministrationPluginBootstrap.ts
SychevAndrey: We definitely don't want to save something partially. We either save all or nothing. So, Promise.all is what we need here.

SychevAndrey: Actually, this is a problem with Promise.all. If we are saving all sources at the same time, there might be races between them. I believe for await here is the safest option.

sergeyteleshev (approved):

PR #4454, opened 2026-07-07
dbeaver/pro#9614 add predefined param for connection type

devnaumov:
closes [9614](https://github.com/dbeaver/pro/issues/9614)


webapp/packages/core-app/src/Body.tsx
```diff
@@ -36,6 +37,7 @@ export const Body = observer(function Body() {
 
   // TODO: must be loaded in place where it is used
   useResource(Body, ProjectInfoResource, CachedMapAllKey, { silent: true });
+  useResource(Body, ConnectionTypeResource, CachedMapAllKey, { silent: true });
```
sergeyteleshev: lets load in place where it is required cause now the tracks are hot and we can do it quite easy
from my perspective inside Connection Tree should be enough, what do you think? cause this tree like is always gonna be shown despite everything and the feature is strongly sticked to it

sergeyteleshev: also in administration part when we load connection types tab

webapp/packages/core-connections/src/ConnectionType/ConnectionTypeResource.ts
```diff
@@ -0,0 +1,99 @@
...
+import { SessionResource } from '@cloudbeaver/core-root';
+import { GraphQLService, type ConnectionType as ConnectionTypeFragment } from '@cloudbeaver/core-sdk';
+
+export type ConnectionType = ConnectionTypeFragment;
+
+export const NEW_CONNECTION_TYPE_SYMBOL = Symbol('new-connection-type');
+export type NewConnectionType = ConnectionType & { [NEW_CONNECTION_TYPE_SYMBOL]: boolean; timestamp: number };
+
+@injectable(() => [GraphQLService, SessionResource])
+export class ConnectionTypeResource extends CachedMapResource<string, ConnectionType> {
```
sergeyteleshev, line 23: I think we should outdate `ConnectionInfoResource` if this resource is outdated to keep things fresh

webapp/packages/core-connections/src/ConnectionType/ConnectionTypeResource.ts
```diff
@@ -0,0 +1,99 @@
...
+  if (isNewConnectionType(b)) {
+    return 1;
+  }
+
+  if (isNewConnectionType(a)) {
+    return -1;
+  }
+
+  return 0;
+}
```
sergeyteleshev, line 124: what about `predefined` flag? do we want to consider it here as well?

webapp/packages/core-connections/src/ConnectionType/useConnectionTypeColor.ts
```diff
@@ -0,0 +1,22 @@
...
+
+export function useConnectionTypeColor(connectionKey: IConnectionInfoParams | undefined): string | undefined {
+  const connectionTypeService = useService(ConnectionTypeService);
+
+  if (!connectionKey) {
+    return;
+  }
+
+  return connectionTypeService.getConnectionTypeColor(connectionKey);
+}
```
sergeyteleshev, line 22: maybe here we also can use useResource for `ConnectionInfoResource` & `ConnectionTypeResource` to be sure all loaded and no cases when we try to calc color but it is not preloaded with resources

webapp/packages/core-ui/src/Tabs/Tab/Tab.module.css
```diff
@@ -28,7 +28,7 @@
   }
 
   &:not(:global([aria-selected='true'])) {
-    background-color: transparent !important;
+    background-color: transparent;
```
sergeyteleshev, line 31: autotests required then to see if no visual regression

webapp/packages/plugin-object-viewer/src/ObjectPropertiesPage/ObjectFolders.tsx
```diff
@@ -115,13 +115,15 @@ export const ObjectFolders = observer<IProps>(function ObjectFolders({ tab }) {
   return folders.length > 0 ? (
     <TabsState currentTabId={folderId} orientation="vertical" localState={innerTabState} lazy onChange={openFolder}>
       <SContext registry={objectFoldersRegistry}>
-        <TabList aria-label="Object folders" vertical>
-          {folders.map(folderId => (
-            <FolderTabRenderer key={folderId} nodeId={nodeId} folderId={folderId} parents={parents} />
-          ))}
-        </TabList>
+        <div>
```
sergeyteleshev, line 118: why do we need `div` here?

webapp/packages/core-connections/src/ConnectionType/ConnectionTypeService.ts
```diff
@@ -0,0 +1,56 @@
...
+
+import { injectable } from '@cloudbeaver/core-di';
+import { ThemeService } from '@cloudbeaver/core-theming';
+
+import { ConnectionTypeResource, mapColorValue, PREDEFINED_UNSET_COLOR } from './ConnectionTypeResource.js';
+import { ConnectionInfoResource } from '../ConnectionInfoResource.js';
+import type { IConnectionInfoParams } from '../CONNECTION_INFO_PARAM_SCHEMA.js';
+
+@injectable(() => [ConnectionTypeResource, ConnectionInfoResource, ThemeService])
+export class ConnectionTypeService {
```
sergeyteleshev, line 17: you have 2 ways to calculate colors now:
1. This `service`
2. `useConnectionTypeColor` hook

both has similar interface. you need a `connectionKey` to keep things going
both cases uses hook and the service for `tsx` components

seems like you can just use the hook and pass to it something like:
`{connectionId?: string, projectId?: string}` - and handle edge case if nothing is provided inside the hook to avoid these `if/else` clauses you have for `service` case of implementation

sergeyteleshev: so smells like we can just remove this `service` and move all of its logic inside the `hook`

webapp/packages/core-connections/package.json
```diff
@@ -37,6 +37,7 @@
     "@cloudbeaver/core-root": "workspace:*",
     "@cloudbeaver/core-sdk": "workspace:*",
     "@cloudbeaver/core-settings": "workspace:*",
+    "@cloudbeaver/core-theming": "workspace:*",
```
sergeyteleshev, line 40: pls create PR for `widgets` repo

webapp/packages/core-app/src/Body.tsx
devnaumov: The connection tree might be hidden for certain roles in TE. Later we will show types for every tree that connected to the connection, for example ai chat custom scope tree and etc. Connection types are strongly connected to the whole app as projects and we want them to be ready as soon as possible. As we will not use this resource in place where we resolve type color, it will be just `useResource(Component, ConnectionTypeResource, CachedMapAllKey)`; in like 10 places that is confusing as well.

sergeyteleshev: okay then adding it to [useConnectionTypeColor.ts](https://github.com/dbeaver/cloudbeaver/pull/4454/files/060a7087dc9cc33a23631f85a29c3c474dc056bb#diff-094d8abda969404e5fc19e7064fb85d33dda0d7e6651318623735a265d5e956e) should be enough, I guess

webapp/packages/core-connections/src/ConnectionType/ConnectionTypeResource.ts
devnaumov: This function only compares by the new flag, general sorting happens inside compare fn as we do in other resources

webapp/packages/plugin-object-viewer/src/ObjectPropertiesPage/ObjectFolders.tsx
devnaumov: Because we need background color to be applied around tabs and not on the tabs itself

webapp/packages/core-connections/src/ConnectionType/ConnectionTypeService.ts
devnaumov: We cant, there might be cases where you need this color in some service or not in react context in general. When all your logic inside hook, you can only use it inside components. Also in some places we dont use hook (for example when you cant use it cause you cant use hooks conditionally), there we get color from service. So the service is source of truth here

webapp/packages/core-connections/src/ConnectionType/useConnectionTypeColor.ts
devnaumov: Its better to now use useResource hook inside hooks that might be used inside nodes or cells as its very perfomance heavy

webapp/packages/core-app/src/Body.tsx
devnaumov: Its better to now use useResource hook inside hooks that might be used inside nodes or cells as its very perfomance heavy. In our case we use it for navigation tree

webapp/packages/core-connections/src/ConnectionType/ConnectionTypeResource.ts
```diff
@@ -68,6 +76,31 @@ export class ConnectionTypeResource extends CachedMapResource<string, Connection
...
   }
+
+  private handleTypeChanges(key: ResourceKey<string>) {
+    const types = this.get(key);
+
+    if (!types) {
+      return;
+    }
+
+    const typeIds: string[] = [];
```
sergeyteleshev, line 87: use `Set` for O(1) search

sergeyteleshev (dismissed):

webapp/packages/plugin-connections/src/PluginBootstrap.ts
```diff
@@ -1,15 +1,22 @@
...
-@injectable()
+@injectable(() => [ConnectionTypeResource])
 export class PluginBootstrap extends Bootstrap {
-  constructor() {
+  constructor(private readonly connectionTypeResource: ConnectionTypeResource) {
     super();
   }
+
+  override async load(): Promise<void> {
+    await this.connectionTypeResource.load(CachedMapAllKey);
```
SychevAndrey: is it really have to block the loading? why not to do it lazy? plus we don't handle errors here

webapp/packages/core-connections/src/ConnectionType/ConnectionTypeResource.ts
```diff
@@ -0,0 +1,132 @@
...
+    this.onItemDelete.addHandler(this.handleTypeChanges.bind(this));
+    this.onItemUpdate.addHandler(this.handleTypeChanges.bind(this));
+    this.onDataOutdated.addHandler(this.handleTypeChanges.bind(this));
+
+    this.sync(
+      sessionResource,
+      () => {},
+      () => CachedMapAllKey,
+    );
+  }
```
SychevAndrey, line 40: do we really need it?
all connection should already be synced with the app, so we need to only sync with connections themselves

webapp/packages/plugin-sql-editor/src/SqlEditor/SqlEditor.tsx
```diff
@@ -60,6 +63,13 @@ export const SqlEditor = observer<ISqlEditorProps>(function SqlEditor({ state, c
     }
   }, [isEditorEmpty]);
 
+  let typeColor: string | undefined;
+
+  const context = data.model.dataSource?.executionContext;
+  if (context) {
+    typeColor = connectionTypeService.getConnectionTypeColor(createConnectionParam(context.projectId, context.connectionId));
```
SychevAndrey: why not webapp/packages/core-connections/src/ConnectionType/useConnectionTypeColor.ts ?

webapp/packages/plugin-sql-editor-navigation-tab/src/SqlEditorTab.tsx
```diff
@@ -62,8 +65,21 @@ export const SqlEditorTab: TabHandlerTabComponent<ISqlEditorTabState> = observer
 
   const tooltip = useObjectInfoTooltip(connection?.name, executionContext?.defaultCatalog, executionContext?.defaultSchema, project?.name);
 
+  let typeColor: string | undefined;
+
+  if (executionContext) {
+    typeColor = connectionTypeService.getConnectionTypeColor(createConnectionParam(executionContext.projectId, executionContext.connectionId));
```
SychevAndrey: I believe we should use the hook in tsx components since it's introduced in this PR

webapp/packages/plugin-datasource-context-switch/src/ConnectionSchemaManager/ConnectionSelector/ConnectionIcon.tsx
```diff
@@ -1,32 +1,39 @@
...
   const { isVisible, setRef } = useVisible();
 
   const deferredIsVisible = useDeferredValue(isVisible);
+  const connectionTypeService = useService(ConnectionTypeService);
 
   if (!isNotNullDefined(connectionKey)) {
     return null;
   }
 
+  const typeColor = connectionTypeService.getConnectionTypeColor(connectionKey);
```
SychevAndrey: useConnectionTypeColor

webapp/packages/core-connections/src/ConnectionType/ConnectionTypeService.ts
```diff
@@ -0,0 +1,56 @@
...
+      return;
+    }
+
+    let color = connectionType.colorLight;
+
+    if (connectionType.colorDark && theme === 'dark') {
+      color = connectionType.colorDark;
+    }
+
+    color = color ?? connectionType.colorLight;
```
SychevAndrey: what is this fallback for? 

color is already colorLight if we don't have colorDark

SychevAndrey (dismissed):

webapp/packages/plugin-connections/src/PluginBootstrap.ts
devnaumov: Dont understand the question. We need to make sure connection types are loaded as soon as possible

webapp/packages/core-connections/src/ConnectionType/ConnectionTypeResource.ts
devnaumov: If sessionResource is outdated, we want connection types to be marked as outdated as well

SychevAndrey: why do we want it? Session resource has fields like language, why do we need to mark all connection types as outdated when we change it?

sergeyteleshev (approved):

PR #4455, opened 2026-07-08
dbeaver/pro#9000 adds API token settings

sergeyteleshev:
closes https://github.com/dbeaver/pro/issues/9000


webapp/packages/core-ui/src/Form/FormPart.ts
```diff
@@ -200,6 +200,12 @@ export abstract class FormPart<TPartState extends object, TFormState = any> impl
     this.state = state;
   }
 
+  extendSchema<TExtension extends schema.ZodRawShape>(extension: TExtension): void {
```
devnaumov: I think its better to use zod passthrough to not strip unrelated fields from state

sergeyteleshev: agreed to `LooseObject` due to its local solution and not deprecated APi comparing to `passthrough`

devnaumov (approved):

PR #4464, opened 2026-07-15
dbeaver/pro#9638 [CB] refresh is needed to see replaced connectionfolder in another instance

SychevAndrey:
There are two parts. The first commit is just fixing stale reference and the second one is an improvement that we might want to have or not (need to discuss)


webapp/packages/plugin-connections/src/NavNodes/ConnectionNavNodeService.ts
```diff
@@ -162,6 +162,10 @@ export class ConnectionNavNodeService {
 
         const folderId = node?.parentId; // current parent
 
+        if (folderId && folderId !== parentId && this.navTreeResource.has(parentId)) {
+          this.navTreeResource.moveToNode(connectionInfo.nodePath, parentId);
```
devnaumov: We should handle this on the CbDatasourceUpdated level as it affects only this branch of logic. One session case is already handled

webapp/packages/core-connections/src/ConnectionInfoResource.ts
```diff
@@ -214,18 +215,18 @@ export class ConnectionInfoResource extends CachedMapResource<IConnectionInfoPar
...
             .requestSynchronization('connection', connection.map(connection => connection?.name).join('\n'))
             .then(state => {
               if (state) {
-                this.markOutdated(key);
+                this.updateFromEvent(key);
               }
             });
         } else {
-          this.markOutdated(key);
+          this.updateFromEvent(key);
```
devnaumov, line 228: Please try something like this


```
for (const k of key) {
            const currentConnection = this.get(k);
            const newConnection = await this.refresh(k);

            if (currentConnection?.nodePath !== newConnection?.nodePath) {
              if (currentConnection?.nodePath) {
                await navNodeInfoResource.loadNodeParents(currentConnection.nodePath);
                const parent = navNodeInfoResource.getParent(currentConnection.nodePath);
                this.navTreeResource.markOutdated(parent);
              }

              if (newConnection.nodePath) {
                await navNodeInfoResource.loadNodeParents(newConnection.nodePath);
                const parent = navNodeInfoResource.getParent(newConnection.nodePath);
                this.navTreeResource.markOutdated(parent);
              }
            }
          }
```

SychevAndrey: Went with your version, just changes a couple of things.

- Removed **loadNodeParents** for the old path: after a move the old path is gone on the server, so it'd just fail. The old node is already in the tree cache, so **getParent** is enough.
- Added try catch for the case when connection is deleted

Still not handled: the cold-cache case where the connection has no **nodePath**. Then we don't know the old location.

devnaumov (approved):

sergeyteleshev (approved):

PR #4473, opened 2026-07-20
dbeaver/pro#9877 UI fixes

SychevAndrey:


webapp/packages/core-blocks/src/ObjectPropertyInfo/ObjectPropertyInfoForm/RenderField.tsx
```diff
@@ -175,6 +175,7 @@ export const RenderField = observer<RenderFieldProps>(function RenderField({
           readOnly={readonly}
           description={property.hint}
           className={className}
+          portal={true}
```
devnaumov: Could break the layout when the selector is shown inside a form and the page is scrolled. The selector may overlap surrounding elements. Also please eslint rule, just porta={true} -> portal

SychevAndrey: But it is also breaking without it. So, what are we going to do?

devnaumov: @sergeyteleshev could you please check the case with broken layout that you ve fixed several month ago?

sergeyteleshev: These cases looks fine to me. I'm gonna write to the ticket to check these cases for QA

Screenshot 2026-07-22 at 10 39 38

Screenshot 2026-07-22 at 10 41 51

sergeyteleshev (approved):

devnaumov (approved):

PR #4476, opened 2026-07-20
dbeaver/pro#9659 add dnd scroll container

devnaumov:
closes [9659](https://github.com/dbeaver/pro/issues/9659)

https://github.com/user-attachments/assets/bded8ef4-c803-4ec5-8306-6257cf74460b

### DIFF
--- a/webapp/packages/plugin-navigation-tree/src/NavigationTree/ElementsTree/ElementsTree.tsx
+++ b/webapp/packages/plugin-navigation-tree/src/NavigationTree/ElementsTree/ElementsTree.tsx
@@ -26,6 +26,7 @@ import {
 import { useService } from '@cloudbeaver/core-di';
 import { EventContext, EventStopPropagationFlag } from '@cloudbeaver/core-events';
 import { EObjectFeature, type NavNode, NavNodeInfoResource, NavTreeResource, ROOT_NODE_PATH } from '@cloudbeaver/core-navigation-tree';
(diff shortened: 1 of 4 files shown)


webapp/packages/plugin-navigation-tree/src/NavigationTree/ElementsTree/ElementsTree.tsx
```diff
@@ -172,7 +173,7 @@ export const ElementsTree = observer(
     return (
       <>
         <ElementsTreeTools tree={tree} settingsElements={settingsElements} />
-        <div ref={treeMergedRef} className={s(styles, { treeBox: true })}>
+        <DNDScrollContainer ref={treeMergedRef} className={s(styles, { treeBox: true })} isDragging={!!dndBox.state.context}>
```
sergeyteleshev, line 176: The best solution here is to invent something tricky with html/css so it triggers scroll. Cause it already works fine, but this zone is too small
I can see why you want JS to handle that. But feel like there is a native html/css solution, it's just tricky to find

sergeyteleshev (approved):

webapp/packages/plugin-navigation-tree/src/NavigationTree/ElementsTree/ElementsTree.tsx
devnaumov: I couldn't find any native solution. Everyone mentions that the native solution has a very small scrollable area. If you know of any solutions, feel free to suggest them.

PR #4506, opened 2026-07-29
dbeaver/pro#8821 feat: add import settings step

SychevAndrey:


webapp/packages/core-sdk/src/queries/connections/getConnectionImportDriverConfiguration.gql
```diff
@@ -0,0 +1,14 @@
+query getConnectionImportDriverConfiguration($projectId: ID, $connectionId: ID!) {
+  connections: userConnections(projectId: $projectId, id: $connectionId) {
+    id
```
sergeyteleshev, line 3: `projectId`?

webapp/packages/plugin-data-import/src/DataImportDriverConfigurationResource.ts
```diff
@@ -0,0 +1,73 @@
...
+  ConnectionInfoResource,
+  type IConnectionInfoParams,
+  isConnectionInfoParamEqual,
+} from '@cloudbeaver/core-connections';
+import { injectable } from '@cloudbeaver/core-di';
+import { CachedMapResource, isResourceAlias, type ResourceKey, ResourceKeyUtils } from '@cloudbeaver/core-resource';
+import { type GetConnectionImportDriverConfigurationQuery, GraphQLService } from '@cloudbeaver/core-sdk';
+import { schemaValidationError } from '@cloudbeaver/core-utils';
+
+export type IDataImportDriverConfiguration = GetConnectionImportDriverConfigurationQuery['connections'][number]['driverConfiguration'];
```
sergeyteleshev: pls create graghql fragments on your own and use it here

webapp/packages/plugin-data-import/src/DataImportDialog/useDataImportDialog.ts
```diff
@@ -25,17 +27,42 @@ const DEFAULT_STATE_GETTER: () => IDataImportDialogState = () => ({
...
+  if (configuration.supportedInsertReplaceMethods) {
+    settings.onDuplicateKeyMethod = undefined;
+  }
+
+  if (configuration.supportsBulkLoad) {
+    settings.useBulkLoad = false;
+  }
+
+  return settings;
+}
```
sergeyteleshev, line 51: it would be nice to have 1 interface with settings so don't need to map it at all

webapp/packages/plugin-data-import/src/DataImportDialog/useDataImportDialog.ts
```diff
@@ -25,17 +27,42 @@ const DEFAULT_STATE_GETTER: () => IDataImportDialogState = () => ({
...
+
+export function useDataImportDialog(initialState?: IDataImportDialogState): IDialog {
+  return useObservableRef<IDialog>(
     () => ({
       state: initialState ?? DEFAULT_STATE_GETTER(),
       stepBack() {
-        if (this.state.step === EDataImportDialogStep.File) {
+        if (this.state.step === EDataImportDialogStep.Settings) {
+          this.state.step = EDataImportDialogStep.File;
+        } else if (this.state.step === EDataImportDialogStep.File) {
```
sergeyteleshev, line 60: what if we introduce this kind of data structure?
`{steps: [Processor, File, Settings], currentStep: 0, goNext(){ this.currentStep++ }, goBack(){this.currentStep--}}`

or something like this so we have like defined determinated steps and can safely iterate it without fearing that somewhere we can miss-step the flow

webapp/packages/plugin-data-import/src/DataImportDialog/DataImportDialog.tsx
```diff
@@ -33,21 +40,51 @@ export const DataImportDialog: DialogComponent<IDataImportDialogPayload, IDataIm
...
   const translate = useTranslate();
   const dialog = useDataImportDialog(payload.initialState);
+  const driverConfigurationResource = useResource(
+    DataImportDialog,
+    DataImportDriverConfigurationResource,
+    createConnectionParam(payload.projectId, payload.connectionId),
+    { silent: true },
+  );
+
+  const driverConfiguration = driverConfigurationResource.tryGetData ?? null;
```
sergeyteleshev, line 49: is there a reason for `tryGetData` here? `data` usually triggers suspense mechanism. is it not needed here?

webapp/packages/plugin-data-import/src/DataImportDialog/DataImportDialog.tsx
```diff
@@ -33,21 +40,51 @@ export const DataImportDialog: DialogComponent<IDataImportDialogPayload, IDataIm
...
-  if (dialog.state.step === EDataImportDialogStep.File && dialog.state.selectedProcessor) {
+  if (dialog.state.step !== EDataImportDialogStep.Processor && dialog.state.selectedProcessor) {
     title += ` (${dialog.state.selectedProcessor.name ?? dialog.state.selectedProcessor.id})`;
     icon = dialog.state.selectedProcessor.icon ?? icon;
   }
 
+  function importData() {
+    if (dialog.state.file && dialog.state.selectedProcessor) {
+      resolveDialog({ file: dialog.state.file, processorId: dialog.state.selectedProcessor.id, settings: dialog.state.settings });
+    }
```
sergeyteleshev, line 67: is it something in the interface notifying user that there are no file or there is no selected processor? maybe not a bad idea to notify what happened if something is off the scenario

webapp/packages/plugin-data-import/src/DataImportDialog/useDataImportDialog.ts
SychevAndrey: we can create a ticket for investigation, I guess

webapp/packages/plugin-data-import/src/DataImportDialog/DataImportDialog.tsx
SychevAndrey: how is it possible?

webapp/packages/plugin-data-import/src/DataImportDialog/DataImportDialog.tsx
SychevAndrey: Yes. Suspense would block the whole dialog on a load. We don't want it, Processor and File steps can render without it. loading is shown on the button via isLoading(), and on error we just skip the Settings step.

webapp/packages/plugin-data-import/src/DataImportDialog/DataImportDialog.tsx
SychevAndrey: So the user can't trigger import without both, so nothing to notify about here

webapp/packages/plugin-data-import/src/DataImportDialog/DataImportDialog.tsx
```diff
@@ -1,28 +1,35 @@
...
 
 export interface IDataImportDialogResult {
   file: File;
   processorId: string;
+  settings: DataTransferImportSettings;
 }
 
 export interface IDataImportDialogPayload {
   tableName: string;
+  projectId: string;
```
devnaumov: connectionKey: IConnectionInfoParams?

webapp/packages/plugin-data-import/src/DataImportDialog/DataImportDialog.tsx
```diff
@@ -33,21 +40,51 @@ export const DataImportDialog: DialogComponent<IDataImportDialogPayload, IDataIm
 }) {
   const translate = useTranslate();
   const dialog = useDataImportDialog(payload.initialState);
+  const driverConfigurationResource = useResource(
+    DataImportDialog,
+    DataImportDriverConfigurationResource,
+    createConnectionParam(payload.projectId, payload.connectionId),
+    { silent: true },
```
devnaumov, line 46: why silent?

webapp/packages/plugin-data-import/src/DataImportDialog/DataImportDialog.tsx
devnaumov, line 49: do we have specific behaviour here? why tryGetData instead of .data?

webapp/packages/plugin-data-import/src/DataImportDialog/useDataImportDialog.ts
devnaumov: Yea, but as long as we only have two steps and a few forms using them, it's fine. However, as the logic becomes more complex, we should reconsider.

webapp/packages/plugin-data-import/src/DataImportDialog/DataImportDialog.tsx
SychevAndrey: see the answer above

webapp/packages/plugin-data-import/src/DataImportDialog/DataImportDialog.tsx
SychevAndrey: I believe that those setting are kinda optional, so if something goes wrong, we can just skip it. If we show an error or smth, we should think about an action the user can do. Retry to get the driver configuration or what? I would let them go and try import with default settings. And if there is a real problem, they will se an error there

webapp/packages/plugin-data-import/src/DataImportDialog/useDataImportDialog.ts
sergeyteleshev: i got it why this happens:

because getting data and saving data has different interfaces on backend:
`dataTransferImportDataIntoResults` - saving
`IDataImportDriverConfiguration`- getting in resource

webapp/packages/plugin-data-import/src/DataImportDialog/DataImportDialog.tsx
devnaumov: Okay, but its better to default to no options for useResource and .data for getting data from resource

devnaumov (dismissed):

webapp/packages/plugin-data-import/src/DataImportDialog/ImportSettingsForm.tsx
```diff
@@ -0,0 +1,72 @@
...
+
+export const ImportSettingsForm = observer<Props>(function ImportSettingsForm({ settings, driverConfiguration }) {
+  const translate = useTranslate();
+  const { supportedInsertReplaceMethods, supportsBulkLoad, supportsTransactions } = driverConfiguration;
+
+  useEffect(() => {
+    if (settings.useBulkLoad) {
+      settings.onDuplicateKeyMethod = undefined;
+    }
+  }, [settings.useBulkLoad]);
```
sergeyteleshev, line 29: could you explain the case pls?
from my perspective we don't modify `useBulkLoad` manually via interface. so it arrives from the resource like that, right? and then we modify `onDuplicateKeyMethod`. isn't it supposed to happen on backend side then? since both of these settings arrives from back in 1 method? correct me if i'm wrong

webapp/packages/plugin-data-import/src/locales/en.ts
```diff
@@ -4,4 +4,21 @@ export default [
...
+    'Execute the import within a database transaction. If the import fails, all changes can be rolled back',
+  ],
+
+  [
+    'plugin_data_import_settings_on_duplicate_key_title',
+    'Choose how to handle rows that already exist in the target table based on a primary or unique key',
+  ],
+  ['plugin_data_import_settings_on_duplicate_key_placeholder', 'None'],
+  ['plugin_data_import_settings_open_new_connection', 'Open new connection'],
+  ['plugin_data_import_settings_open_new_connection_title', 'Open a new connection after the import is complete'],
```
sergeyteleshev: `is completeD`?

webapp/packages/plugin-data-import/src/DataImportDialog/ImportSettingsForm.tsx
SychevAndrey: we modify both

webapp/packages/plugin-data-import/src/locales/en.ts
SychevAndrey: image

sergeyteleshev (dismissed):

webapp/packages/plugin-data-import/src/DataImportDialog/ImportSettingsForm.tsx
sergeyteleshev: onChange + mutation then would be nice so we avoid `useEffect` pattern

devnaumov (dismissed):

devnaumov (approved):

sergeyteleshev (approved):

PR #4507, opened 2026-07-29
4551 generate full ddl for postgresql metaobjects

sergeyteleshev:
closes https://github.com/dbeaver/pro/issues/4551

`metadataGetNodeDDL` & `sqlGenerateEntityQuery` - these queries basically do the same, so backend introduced `asyncSqlGenerateEntityQuery` (which allows to get heavy queries as well) and we need to use only it so we don't have any code & logic duplications

also I have introduced new plugin `core-sql-generator` and moved the resource there so we don't have to import `plugin-sql-generator` into `plugin-ddl-viewer`. these 2 plugins already have enough deps with `sql-editor` and `data-editor`. by that we reduce probability of a circular deps plugins import in the future


webapp/packages/plugin-ddl-viewer/src/DdlViewer/DDLViewerTabPanel.tsx
```diff
@@ -42,16 +57,49 @@ export const DDLViewerTabPanel: NavNodeTransformViewComponent = observer(functio
...
+      notificationService.logException(error, 'plugin_ddl_viewer_full_ddl_error_title');
+    } finally {
+      setDdlLoading(false);
+    }
+  }
+
+  useEffect(() => {
+    if (ddlGenerator) {
+      loadDdl(isFullDdl);
+    }
```
SychevAndrey: looks like it should be a separate resource or at least a part of SqlGeneratorsResource.

SychevAndrey (changes requested):

webapp/packages/plugin-ddl-viewer/src/DdlViewer/DDLViewerTabPanel.tsx
sergeyteleshev: nice catch! moved it to a separate resource so we can reuse its all features we need in this component

SychevAndrey (dismissed):

webapp/packages/core-sql-generator/src/module.ts
```diff
@@ -0,0 +1,22 @@
...
+ * Copyright (C) 2020-2026 DBeaver Corp and others
+ *
+ * Licensed under the Apache License, Version 2.0.
+ * you may not use this file except in compliance with the License.
+ */
+import { Dependency, ModuleRegistry, proxy } from '@cloudbeaver/core-di';
+import { SqlEntityQueryResource } from './SqlEntityQueryResource.js';
+import { SqlGeneratorsResource } from './SqlGeneratorsResource.js';
+
+export default ModuleRegistry.add({
```
devnaumov: Please make it just plugin, dont prefix it with core

webapp/packages/plugin-sql-generator/src/SqlGenerators/GeneratedSqlDialog.tsx
```diff
@@ -149,6 +158,16 @@ export const GeneratedSqlDialog = observer<DialogComponentProps<Payload>>(functi
                 label={translate('app_shared_sql_generators_compact_sql')}
                 onChange={value => state.handleOptionChange('compactSql', value)}
               />
+              {hasFullDdl && (
+                <Checkbox
+                  id="show-full-ddl"
+                  state={state}
```
devnaumov: state and name can be omitted as you update state value using onChange, other fields too

webapp/packages/plugin-ddl-viewer/src/DdlViewer/DDLQueryStateService.ts
```diff
@@ -0,0 +1,26 @@
...
+ * Copyright (C) 2020-2026 DBeaver Corp and others
+ *
+ * Licensed under the Apache License, Version 2.0.
+ * you may not use this file except in compliance with the License.
+ */
+import { injectable } from '@cloudbeaver/core-di';
+import { MetadataMap } from '@cloudbeaver/core-utils';
+
+@injectable()
+export class DDLQueryStateService {
```
devnaumov: Please move this logic to the footer service, you can keep only one method, toggleFullDdl

devnaumov (dismissed):

PR #4512, opened 2026-07-30
dbeaver/pro#10062 syncs cell selection on right click to avoid selection collision

sergeyteleshev:
closes https://github.com/dbeaver/pro/issues/10062

syncs focus to the right-clicked cell in CellRenderer.openContextMenu before opening the context menu, via a new focusCell method on the selection context. This guarantees the app's focus always matches the clicked cell, regardless of whether react-data-grid emitted a position-change event


sergeyteleshev:
closed in favor of https://github.com/dbeaver/cloudbeaver/pull/4513

PR #4518, opened 2026-07-31
dbeaver/pro#9999 api for ai engine icon

yagudin10:


webapp/packages/plugin-ai-administration/src/AIProfiles/AIProfilesTable.tsx
```diff
@@ -74,7 +74,17 @@ export const AIProfilesTable = observer<Props>(function AIProfilesTable({ profil
...
+
+      if (engine?.icon) {
+        return (
+          <div title={title} className="tw:flex tw:justify-center">
+            <IconOrImage icon={engine.icon} />
+          </div>
+        );
+      }
+
+      return <span title={title}>{title}</span>;
```
sergeyteleshev, line 96: lets show and icon and the engine name. users cannot know each engine by only icon. right now they have to hover too see the name and we can easily avoid this action since there is enough space in the table

webapp/packages/plugin-ai/src/module.ts
```diff
@@ -0,0 +1,19 @@
...
+ * Licensed under the Apache License, Version 2.0.
+ * you may not use this file except in compliance with the License.
+ */
+
+import { Dependency, ModuleRegistry, proxy } from '@cloudbeaver/core-di';
+
+import { AiEnginesResource } from './AiEnginesResource.js';
+
+export default ModuleRegistry.add({
+  name: '@cloudbeaver/plugin-ai',
```
sergeyteleshev, line 14: plugin-ai is too common name. in the future we may want to place everything to it instead of decomposing  ai-features to the plugins, I guess

`plugin-ai-engines`?

sergeyteleshev (dismissed):

webapp/packages/plugin-ai/src/module.ts
devnaumov: We have plugin-ai-administration, so it seems like plugin-ai would be a suitable public plugin name here, at least for now

sergeyteleshev (approved):

PR #4525, opened 2026-08-04
dbeaver/pro#8926 feat: add js helpers for export as image

SychevAndrey:


devnaumov (dismissed):

webapp/common-typescript/@dbeaver/js-helpers/src/isImageBroken.ts
```diff
@@ -0,0 +1,23 @@
+/*
+ * CloudBeaver - Cloud Database Manager
+ * Copyright (C) 2020-2026 DBeaver Corp and others
+ *
+ * Licensed under the Apache License, Version 2.0.
+ * you may not use this file except in compliance with the License.
+ */
+
+export function isImageBroken(imageUrl: string): Promise<boolean> {
```
sergeyteleshev, line 9: you have moved some helpers from `core-utils` here but didn't remove it there
so pls remove it and change implementations imports

webapp/common-typescript/@dbeaver/js-helpers/src/download.ts
```diff
@@ -0,0 +1,32 @@
+/*
+ * CloudBeaver - Cloud Database Manager
+ * Copyright (C) 2020-2026 DBeaver Corp and others
+ *
+ * Licensed under the Apache License, Version 2.0.
+ * you may not use this file except in compliance with the License.
+ */
+
+export function download(content: Blob | string, fileName = '', blank?: boolean): void {
```
sergeyteleshev, line 9: this one as well

webapp/common-typescript/@dbeaver/js-helpers/src/withTimestamp.ts
```diff
@@ -0,0 +1,14 @@
+/*
+ * CloudBeaver - Cloud Database Manager
+ * Copyright (C) 2020-2026 DBeaver Corp and others
+ *
+ * Licensed under the Apache License, Version 2.0.
+ * you may not use this file except in compliance with the License.
+ */
+
+export function withTimestamp(value: string): string {
```
sergeyteleshev, line 9: and this one

webapp/common-typescript/@dbeaver/js-helpers/src/exportDomImage.ts
```diff
@@ -0,0 +1,91 @@
...
+}
+
+function resolveSize(node: Element, config?: IDomImageExportConfig): { width: number; height: number } {
+  if (config?.width !== undefined && config.height !== undefined) {
+    return { width: config.width, height: config.height };
+  }
+
+  const rect = node.getBoundingClientRect();
+
+  return { width: config?.width ?? rect.width, height: config?.height ?? rect.height };
```
sergeyteleshev: what if `width` is `undefined` but height is not. in that case we can get result like this:
`{width: rect.width, height: config.height}` - is that accurate to use it? or better to make it consistent just return `rect` there instead if some of the `config` properties are `undefined`?

webapp/common-typescript/@dbeaver/js-helpers/src/fitToPixelBudget.ts
```diff
@@ -0,0 +1,34 @@
...
+  maxSide?: number;
+  maxPixels?: number;
+}
+
+export function fitToPixelBudget(width: number, height: number, budget?: IPixelBudget): number {
+  const { scale = 1, maxSide = DEFAULT_MAX_IMAGE_SIDE, maxPixels = DEFAULT_MAX_IMAGE_PIXELS } = budget ?? {};
+
+  if (width <= 0 || height <= 0) {
+    return scale;
+  }
```
sergeyteleshev, line 29: could you pls clarify how `width` or `height` can be lower that zero?

sergeyteleshev (changes requested):

webapp/common-typescript/@dbeaver/js-helpers/src/fitToPixelBudget.ts
SychevAndrey: it's a number, you can pass any number below zero. we can take a modulo or just set Math.max(0, width) or just check like here

webapp/common-typescript/@dbeaver/js-helpers/src/exportDomImage.ts
SychevAndrey: I agree. We could keep the current implementation but only if we keep aspect-ration of the original rect with provided width/height. In current implementation it's a bit strange

sergeyteleshev (approved):

devnaumov (approved):

PR #4532, opened 2026-08-06
dbeaver/pro#8752 fix: change loaders check

SychevAndrey:
A loader that has already finished can no longer add items, so it must not keep a submenu alive

### DIFF
--- a/webapp/packages/core-ui/src/ContextMenu/SubMenuElement.tsx
+++ b/webapp/packages/core-ui/src/ContextMenu/SubMenuElement.tsx
@@ -76,7 +76,7 @@ export const SubMenuElement = observer<ISubMenuElementProps>(function SubMenuEle
     }
 
     const hasVisibleItems = subMenuData.items.some(item => !item.hidden);
-    const canLazyLoadItems = subMenuData.loaders.length > 0;
+    const canLazyLoadItems = subMenuData.loaders.some(loader => !loader.isLoaded());
 
     return !hasVisibleItems && !canLazyLoadItems;
   });


webapp/packages/core-ui/src/ContextMenu/SubMenuElement.tsx
```diff
@@ -76,7 +76,7 @@ export const SubMenuElement = observer<ISubMenuElementProps>(function SubMenuEle
     }
 
     const hasVisibleItems = subMenuData.items.some(item => !item.hidden);
-    const canLazyLoadItems = subMenuData.loaders.length > 0;
+    const canLazyLoadItems = subMenuData.loaders.some(loader => !loader.isLoaded());
```
sergeyteleshev, line 79: This is a tricky solution because some menus or submenus needs to be rendered first, then it triggers loaders, it gets the data, and then using its data `isHidden`, `isDisabled` to show/hide/disable some items of the menu

If we don't render it at all, it wont trigger loading and the next flow of defining wether menu item is supposed to be shown/hidden/disabled etc. So as a result menu won't be drawn at all

Pls verify that behavior that submenus are not broken. As far as I remember this thing wasn't fixed yet

sergeyteleshev: if my concerns are correct, it will be better to go with local tools menu solution for this ticket

sergeyteleshev: also changes are affects all of the submenus

devnaumov, line 79: Yea, we cant accept this fix, should investigate further

devnaumov (changes requested):

webapp/packages/core-ui/src/ContextMenu/SubMenuElement.tsx
SychevAndrey: https://github.com/dbeaver/cloudbeaver/pull/3993/changes 
the bug was introduced here
When something is hidden like 
`ACTION_RESOURCE_MANAGER_SCRIPTS — isHidden: () => !resourceManagerScriptsService.enabled`
 we still have `loaders.length > 0` and **canLazyLoadItems** never goes **false**, and the button renders
 
 with the fix, menu hides if loaders resolved to nothing (**canLazyLoadItems** is **false** when everything is loaded and we only look at **hasVisibleItems**)

PR #4537, opened 2026-08-07
dbeaver/pro#9982 adds position prop to InputAutocompletionMenu for accurate placement

sergeyteleshev:
closes https://github.com/dbeaver/pro/issues/9982


devnaumov (dismissed):

webapp/packages/core-blocks/src/FormControls/useInputAutocomplete.ts
```diff
@@ -199,14 +208,14 @@ export const useInputAutocomplete = (
...
     document.body.removeChild(span);
 
-    state.position = {
+    state.setPosition({
       x: spanRect.width + letterWidth,
       y: spanRect.height + CONTEXT_INPUT_OFFSET_Y,
-    };
-  }, [state.inputValue]);
+    });
+  }, [state.inputValue, state.selectionStart]);
```
sergeyteleshev, line 218: on `esc` we reset proposals:
https://github.com/dbeaver/cloudbeaver/blob/b00443f8b29bb7316a71ddf5343ef1a26e4e6de5/webapp/packages/core-blocks/src/FormControls/useInputAutocomplete.ts#L151-L153

but the text input for `where filter` still has value and if we type something new then quickly remove what we just typed the input remains the same and no re-computations happened for proposals positioning in `useEffect`

so I just synced this behavior if we reset proposals we need to recompute proposals position as well

sergeyteleshev: **Full bug case:** 
2. Enter something
3. Press ESC
4. The input field has focus, click to remove focus
5. Then click back and within 300ms you must type one character and then delete it
6. See that tooltips are drawn at coordinates 0,0

devnaumov (approved):

PR #4538, opened 2026-08-07
dbeaver/pro#10177 pass defaults to main properties

devnaumov:
closes [10177](https://github.com/dbeaver/pro/issues/10177)

Root cause: formState was published before initialization completed. The mounted form triggered useAutoLoad(), racing with setDriverId() and overwriting driver defaults via setInitialState().

Fix: Initialize ConnectionFormState and ConnectionFormOptionsPart locally, load driver defaults, apply config, then assign the completed state to this.formState. getDefaults() now also preserves the current Options state.


webapp/packages/plugin-connections/src/ConnectionForm/Options/ConnectionFormOptionsPart.ts
```diff
@@ -340,6 +338,36 @@ export class ConnectionFormOptionsPart extends FormPart<IConnectionFormOptionsSt
...
+      config.driverId = this.state.driverId;
+    }
+
+    if (this.state.authModelId) {
+      const authProperties = await this.getConnectionAuthModelProperties(this.state.authModelId);
+      config.credentials = getObjectPropertyDefaults(authProperties);
+    }
+
+    if (config.driverId) {
+      const driver = await this.dbDriverResource.load(config.driverId, ['includeMainProperties']);
```
sergeyteleshev, line 355: includes is pattern we refused to support a while ago. it would be nice to create new resource `DBDriverMainPropertiesResource` and reuse it here

webapp/packages/plugin-connections/src/ConnectionForm/Options/ConnectionFormOptionsPart.ts
```diff
@@ -340,6 +338,36 @@ export class ConnectionFormOptionsPart extends FormPart<IConnectionFormOptionsSt
...
+
+      if (config.mainPropertyValues) {
```
sergeyteleshev, line 357: we get config `defaultStateGetter`
in this helper `mainPropertyValues` is always an object

so this check is always true
probably there is a miss-match with the `zod` type here:
https://github.com/dbeaver/cloudbeaver/blob/8bbf4c2f8b2d25188e9ce3490d298ce7de979222/webapp/packages/plugin-connections/src/ConnectionForm/CONNECTION_CONFIG_SCHEMA.ts#L26

webapp/packages/plugin-connections/src/PublicConnectionForm/PublicConnectionFormService.ts
```diff
@@ -91,7 +91,10 @@ export class PublicConnectionFormService {
       connectionId: config.connectionId,
     }).setMode(config.connectionId ? FormMode.Edit : FormMode.Create);
 
-    await this.optionsPart?.load();
+    // Don't call load in create mode. There we rely on the defaults set here, and useAutoLoad will load the part on mount.
+    if (this.formState.mode === FormMode.Edit) {
+      await this.optionsPart?.load();
+    }
```
sergeyteleshev: Something is off here

Previously we waited for part being `loaded` and set `initialState`. And only then we went to setting `driverId` and all attached config to it. And it was pretty straightforward and imperative for both form modes: `create` and `edit`

Right now it smells like race condition. What if in `create` mode we resolve `setDriverId()` before the `optionsPart` is actually `loaded` and set `initialState`?

sergeyteleshev: What was the original issue here? Maybe we can resolve it differently?

webapp/packages/plugin-connections/src/ConnectionForm/Options/ConnectionFormOptionsPart.ts
devnaumov: Not scope of the ticket, code is already exist in this part

webapp/packages/plugin-connections/src/ConnectionForm/Options/ConnectionFormOptionsPart.ts
devnaumov: yes, types are from zod schema, there mainPropertyValues is optional

sergeyteleshev (approved):

PR #4543, opened 2026-08-11
dbeaver/pro#10205 removes nanoid CVE

sergeyteleshev:
closes https://github.com/dbeaver/pro/issues/10205

### DIFF
--- a/webapp/package.json
+++ b/webapp/package.json
@@ -22,7 +22,8 @@
     "js-yaml": ">=5.2.2",
     "brace-expansion": ">=5.0.9",
     "axios": ">=1.18.0",
-    "postcss": ">=8.5.18"
+    "postcss": ">=8.5.18",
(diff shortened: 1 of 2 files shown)


webapp/package.json
```diff
@@ -22,7 +22,9 @@
     "js-yaml": ">=5.2.2",
     "brace-expansion": ">=5.0.9",
     "axios": ">=1.18.0",
-    "postcss": ">=8.5.18"
+    "postcss": ">=8.5.18",
+    "nanoid": ">=3.3.17",
+    "less": ">=4.7.0"
```
sergeyteleshev, line 27: deps chain is like this:
`vite` -> `less` -> `image-size`

`image-size` had CVE and it is readonly repo now, not supported by author anymore. so I've updated `less` to version where they removed this `image-size` dep

devnaumov (approved):

PR #4545, opened 2026-08-11
dbeaver/pro#10137 fixes max height for ariakit groups

sergeyteleshev:
closes https://github.com/dbeaver/pro/issues/10137


SychevAndrey (changes requested):
I believe we should use --popover-avaliable-height here.

Image

Please try, maybe __group change is redundant and we can have this style only on __popover level

SychevAndrey (approved):

PR #4565, opened 2026-08-17
dbeaver/pro#6505 adds min value for object property info

sergeyteleshev:
closes https://github.com/dbeaver/pro/issues/6505


server/bundles/io.cloudbeaver.server/schema/service.core.graphqls
```diff
@@ -25,6 +25,8 @@ type ObjectPropertyInfo {
     validValues: [ Object ]
     "Default property value"
     defaultValue: Object
+    "Minimum allowed value for numeric properties"
+    minValue: Int @since(version: "26.2.0")
```
devnaumov: I think it's better to add a new constraints field. Once we have more constraints, maxValue, minLength for text inputs, etc, the flat list will grow too large, so it's better to keep them all here.

```
type PropertyConstraints {
    min: Float
    max: Float
    step: Float
    minLength: Int
    maxLength: Int
}

# в ObjectPropertyInfo:
constraints: PropertyConstraints @since(version: "26.2.0")
```

sergeyteleshev: i personally like the idea and wanted to do so from the start. but decided to start with simple implementation which is required right now. and maybe later extend it with something like you suggest. if backend team don't mind I would implement that

yagudin10: sounds good to me

SychevAndrey (dismissed):

yagudin10 (dismissed):

yagudin10 (approved):

HocKu7 (approved):

PR #4586, opened 2026-08-21
dbeaver/cloudbeaver#4564 change db user password

### LINKED ISSUE #4564: Introduce CloudBeaver password change UI
**Is your feature request related to a problem? Please describe.**

CloudBeaver exposes no UI to change the password of the connected database user. Two consequences:

1. Expired DB credentials (Oracle `ORA-28001`, PostgreSQL `SQLSTATE 28P01` with `password expired`) block login. The user has no in-app path to set a new password and depends on an administrator.
2. Routine DB-side password rotation forces a switch to DBeaver desktop or a native client (`sqlplus`, `psql`).

**Describe the solution you'd like**

Add a per-connection "Change database password" action, mirroring DBeaver desktop:

1. Context-menu entry on a connection: `Security > Change database password`.
2. Dialog collects current password, new password, and repeat. Server applies the configured password policy.
3. Backend GraphQL mutation invokes the per-dialect `SQLDialectChangePasswordHandler` reachable from the embedded DBeaver core bundles.
4. Expired-password login variant applies only to Oracle (`ORA-28001`), MySQL (`ER_MUST_CHANGE_PASSWORD` 1820), and SQL Server (error 18488). The login dialog offers an inline "set new password" flow using the driver's native change-at-login mechanism for those three engines. Postgres-family (PostgreSQL, Greenplum, CockroachDB), Exasol, and Vertica have no equivalent driver-level flow. On those engines, users change the password from the normal "Change database password" action while credentials remain valid.
5. Gate the action behind three AND-combined conditions, each mapped to an existing CloudBeaver policy dimension. The action appears only when all three are true:
   - Server flag `dbUserPasswordChangeEnabled` or similar, likely added to `CBAppConfig` next to `publicCredentialsSaveEnabled` / `enableReverseProxyAuth`. Default `false`.
   - Team permission `PERMISSION_DB_USER_PASSWORD_CHANGE` or similar, likely added to `DBWConstants`, enforced via `@WebAction(requirePermissions = ...)` on the mutation.
   - Per-connection field `configuration.dbUserPasswordChangeAllowed` or similar on the data-source entry. Default `false`.
6. The mutation writes an audit-log entry unconditionally, independent of the gates.

Compatibility of the change was verified against the databases DBeaver desktop already supports for this action: Oracle, PostgreSQL, MySQL, SQL Server, Greenplum, Cockroach, Exasol, Vertica.

**Describe alternatives you've considered**

- Redirect users to DBeaver desktop or a DBA. This forces context switch and undermines CloudBeaver as a single-tab SQL entry point.
- Admin-driven password reset outside the tool. This exposes user credentials to admins and does not resolve the expired-password login failure.

**Additional context**

- Per-dialect `SQLDialectChangePasswordHandler` classes live in DBeaver core. CloudBeaver server bundles already declare `Require-Bundle` on `org.jkiss.dbeaver.*`, so the handlers are reachable without new dependencies.
- The existing `ChangePassword.tsx` under `webapp/packages/plugin-user-profile/src/UserProfileForm/UserAuthenticationPart/` targets the CloudBeaver-local user (via `UserInfoResource.updateLocalPassword`) and is not applicable to the connected database user.
- Related, closed: #361 (CloudBeaver-local user, different scope).
- The three gate dimensions cover mixed deployments where the same user connects to PAM-owned, Vault-owned, Kerberos `IDENTIFIED EXTERNALLY`, CMU `IDENTIFIED GLOBALLY`, and personally-owned accounts in one CloudBeaver instance. Each of those connection classes has different rules for who may change the password, so a single-axis gate (flag-only or permission-only) does not suffice.
- Per-database credential-lifecycle notes:
  - Oracle: `IDENTIFIED BY new REPLACE old` is REQUIRED when the caller's own password is expired.
  - MySQL: `disconnectOnExpiredPasswords=false` MUST be set on the JDBC URL to enable Connector/J sandbox mode for the expired-login variant.
  - SQL Server: the expired-login variant MUST use the Microsoft JDBC driver `SQLServerConnection.setChangePassword` API.
  - Exasol: expiry uses a server-side grace period (`GRACE_DAYS`). Password change during grace is the only supported path. No driver handshake.
  - Vertica: the DEFAULT profile carries `PASSWORD_LIFE_TIME = UNLIMITED`. Expiry applies only when a non-default profile is assigned.
  - PostgreSQL, Greenplum, CockroachDB: expiry is `VALID UNTIL <timestamp>` per role. Expired credentials fail authentication with no server-mediated recovery. Users MUST rotate before `VALID UNTIL` elapses.

TobyTheHutt:
## Summary

Relates to #4564

Adds a `Security > Change database password` context-menu action on connections. The action opens a modal dialog and calls a new GraphQL mutation that invokes the DBeaver core `DBAUserPasswordManager` adapter of the connected driver.

## Design

- Server flag `dbUserPasswordChangeEnabled` gates the mutation server-wide. Default is `false`. Persistence goes through `CBAppConfig`, `AdminServerConfig`, `CBServerConfigurationController`, `CBServerConfigurationMapper`, and `WebServerConfig`. Admin toggle lives in Configuration Wizard > Security.
- Mutation `changeConnectionUserPassword(projectId: ID, connectionId: ID!, oldPassword: String!, newPassword: String!): Boolean!`. Guard: `@WebProjectAction(requireProjectPermissions = {PERMISSION_PROJECT_DATASOURCES_EDIT})`. Both password parameters carry `@WebParameterSecure` to keep them out of logs.
- On DB rejection the mutation raises `DBWebException` wrapping the driver's exception. Callers receive the original SQL state, message, and cause chain.
- Frontend menu appears when the server flag is on and `connection.canEdit` is true. Backend re-checks the project permission on every call.
- Audit events fire on the `cb_security_audit` topic with kinds `ATTEMPTED`, `SUCCEEDED`, `FAILED`, and `GATE_REJECTED`. Payload carries `sessionId`, `userId`, `projectId`, `connectionId`, `driverId`, `kind`, `reasonCode`, `errorClass`. No plaintext passwords.

Screenshots:

<details>

PW-Reset Flag:

Screenshot 2026-08-21 102959

PW-Change navigation:

Screenshot 2026-08-21 103043

PW-Change dialog:

Screenshot 2026-08-21 103120

PW-Change confirmation:
Screenshot 2026-08-21 103136

</details>

## Engine coverage

Dispatch runs through the DBeaver core `DBAUserPasswordManager` adapter. Verified against Oracle 21c XE, PostgreSQL 16, and SQL Server 2022. Engines whose driver ships an adapter (Greenplum, CockroachDB, Exasol, Vertica) inherit support without further code changes. Drivers without an adapter (H2, MySQL) return `GATE_REJECTED reasonCode=DIALECT_UNSUPPORTED` in the audit trail and `This driver does not support password change from CloudBeaver.` in the notification.

## Out of scope
- Expired-password login variant (Oracle `ORA-28001`, MySQL `1820`, SQL Server `18488`).
- MySQL adapter (upstream DBeaver contribution).
- Locale translations beyond English.

## Test plan

- [ ] Admin toggles `Enable database user password change` in Configuration Wizard or Server Administration > Security.
- [ ] With flag on: right-click a configured and logged-on PostgreSQL/Oracle/MSSQL connection > `Security > Change database password`. Submit `oldpw`, `newpw`, `newpw`. Reconnect with `newpw` succeeds. Reconnect with `oldpw` returns `SQLSTATE 28P01` or technology equivalent.
- [ ] With flag off: the `Security` submenu is absent.
- [ ] Non-admin user without `datasources-edit` on the project sees no menu. Direct GraphQL mutation returns a permission-denied response.
- [ ] Wrong-old-password on Oracle: notification body contains `ORA-28008: invalid old password`. On SQL Server: `Msg 15116`.
- [ ] H2 connection: notification `This driver does not support password change from CloudBeaver.`. Audit line `kind=GATE_REJECTED reasonCode=DIALECT_UNSUPPORTED`.
- [ ] `docker logs cloudbeaver-dev | grep cb_security_audit` shows one `ATTEMPTED` + one `SUCCEEDED` per successful change, `ATTEMPTED` + `FAILED` with `errorClass` on driver rejection, and no plaintext passwords.


LonwoLonwo:
Hello @TobyTheHutt 

Thanks for the contribution.
We will take a look and get back to you with the answer and comments soon.

server/bundles/io.cloudbeaver.server/src/io/cloudbeaver/service/core/DBWServiceCore.java
```diff
@@ -170,6 +170,15 @@ WebConnectionInfo testConnection(
         @NotNull Map<String, Object> connectionConfig
     ) throws DBWebException;
 
+    @WebProjectAction(requireProjectPermissions = {RMConstants.PERMISSION_PROJECT_DATASOURCES_EDIT})
```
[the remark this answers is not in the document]
TobyTheHutt: Yes it does contradict the issue. I started implementing the shape and realized that it doesn't make sense to have a per-team pw-reset block if the user can just proceed to change their password through an SQL prompt. Reasons to block PW-Resets are for shared-connections which use service accounts (when the user doesn't have edit permission on the connection object), or when an external authentication mechanism like RADIUS or OIDC are in-use.

If the user has permission to run statements, they usually have permission to change their passwords on the DB technologies known to me.

The only narrow use-case in which this change would expose an undesired credential change is, if the database was set up with `PERMISSION_PROJECT_DATASOURCES_EDIT` or an analogue setting, while having no SQL execution rights. This would allow the user to change the password through the GUI while they couldn't through an SQL worksheet. However, this is extremely specific and the only rationale I could think of to have per-team PW-update settings.

I did generally not treat the issue as a hard policy or contract and deviated for these reasons. But I can change it, if that's desired.

server/bundles/io.cloudbeaver.server/src/io/cloudbeaver/service/core/impl/WebServiceCore.java
```diff
@@ -641,6 +681,203 @@ private DataSourceDescriptor getDataSourceDescriptor(
...
+    public boolean changeConnectionUserPassword(
+        @NotNull WebSession webSession,
+        @Nullable String projectId,
+        @NotNull String connectionId,
+        @NotNull String oldPassword,
+        @NotNull String newPassword
+    ) throws DBWebException {
+        DBPDataSourceContainer container = null;
+        try {
+            requireServerFlagEnabled(webSession, projectId, connectionId);
```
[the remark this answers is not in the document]
TobyTheHutt: This is identical to the old `@WebAction` shape and not something I introduced. I can fix this with my PR, if that's desired. Framework-level permission denials appear in the server request log.

server/bundles/io.cloudbeaver.server/src/io/cloudbeaver/service/core/impl/WebServiceCore.java
```diff
@@ -641,6 +681,203 @@ private DataSourceDescriptor getDataSourceDescriptor(
...
+        @NotNull String connectionId,
+        @NotNull String newPassword
+    ) throws DBWebException {
+        if (newPassword.length() > MAX_PASSWORD_LENGTH) {
+            emitPasswordChangeAudit(webSession, container, projectId, connectionId,
+                WSSecurityAuditEvent.Kind.GATE_REJECTED, "PASSWORD_TOO_LONG", null);
+            throw new DBWebException(
+                "Password exceeds the maximum length of " + MAX_PASSWORD_LENGTH + " characters.");
+        }
+        if (containsBlockedPasswordChar(newPassword)) {
```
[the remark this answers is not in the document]
TobyTheHutt: `PasswordPolicyConfig` governs CloudBeaver user account passwords, not DB user passwords. DB user password policy is the DB's concern. In this PR, I surface DB policy rejections verbatim via DBWebException. Applying the CB policy here would enforce the wrong layer.

TobyTheHutt:
I looked at the Copilot feedback and commented or reacted to it based on my personal judgement. I'll provide a follow-up commit after the human maintainer review and further instructions.

This PR should not directly close the linked issue. It only covers the DB user password change. It does not yet cover the password-change prompt upon an expired DB user password, which was the original intent of the linked issue.

If any rephrasing or re-specification of the issue is desired, let me know and I'll be on it.

webapp/packages/plugin-connections/src/ContextMenu/ConnectionMenuBootstrap.ts
```diff
@@ -220,6 +225,42 @@ export class ConnectionMenuBootstrap extends Bootstrap {
         }
       },
     });
+
+    // UX-only gate. Backend re-checks PERMISSION_PROJECT_DATASOURCES_EDIT on the mutation.
+    this.menuService.addCreator({
```
devnaumov: First of all, thanks for the contribution. There’s one important thing to fix on the client side.

Please do not create a new MENU_CONNECTION_SECURITY menu. Register the new action under connection-management, alongside ACTION_CONNECTION_CHANGE_CREDENTIALS, ACTION_CONNECTION_DISCONNECT, etc.

For the new action, check whether dbUserPasswordChangeEnabled is enabled and whether connection?.canEdit is true inside the isHidden handler for this specific action.

```
this.actionService.addHandler({
      id: 'connection-management',
      actions: [ACTION_CONNECTION_CHANGE_DB_PASSWORD, ACTION_CONNECTION_CHANGE_CREDENTIALS, ACTION_CONNECTION_DISCONNECT, ACTION_CONNECTION_DISCONNECT_ALL],
... 
```

webapp/packages/plugin-connections/src/ContextMenu/ConnectionMenuBootstrap.ts
```diff
@@ -220,6 +225,42 @@ export class ConnectionMenuBootstrap extends Bootstrap {
...
+        return [...items, MENU_CONNECTION_SECURITY];
+      },
+    });
+
+    this.menuService.addCreator({
+      menus: [MENU_CONNECTION_SECURITY],
+      getItems: (context, items) => [...items, ACTION_CONNECTION_CHANGE_DB_PASSWORD],
+    });
+
+    this.actionService.addHandler({
```
devnaumov: This can be removed. Please see the previous comment.

webapp/packages/plugin-connections/src/locales/en.ts
```diff
@@ -1,4 +1,15 @@
 export default [
+  ['plugin_connections_menu_security', 'Security'],
```
devnaumov: Please move the new localization block to the end of the array.

devnaumov (changes requested):

server/bundles/io.cloudbeaver.server/src/io/cloudbeaver/server/events/WSSecurityAuditEvent.java
```diff
@@ -0,0 +1,99 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.cloudbeaver.server.events;
+
+import org.jkiss.code.NotNull;
+import org.jkiss.code.Nullable;
+import org.jkiss.dbeaver.model.websocket.event.WSAbstractEvent;
+
+public class WSSecurityAuditEvent extends WSAbstractEvent {
```
yagudin10, line 23: Do we need that class? I think it would be better just write logs as usual by using Log.getLog(Class.class)

server/bundles/io.cloudbeaver.server/src/io/cloudbeaver/server/events/WSSecurityAuditEventHandler.java
```diff
@@ -0,0 +1,44 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.cloudbeaver.server.events;
+
+import org.jkiss.code.NotNull;
+import org.jkiss.dbeaver.Log;
+import org.jkiss.dbeaver.model.websocket.WSEventHandler;
+
+public class WSSecurityAuditEventHandler implements WSEventHandler<WSSecurityAuditEvent> {
```
yagudin10, line 23: Don't think that we need it.

server/bundles/io.cloudbeaver.server/src/io/cloudbeaver/service/core/impl/WebServiceCore.java
```diff
@@ -84,6 +88,42 @@ public class WebServiceCore implements DBWServiceCore {
 
     private static final Log log = Log.getLog(WebServiceCore.class);
 
+    private static final int MAX_PASSWORD_LENGTH = 128;
+    // ASCII space (0x20) is the first printable character. Anything below is a control character.
+    private static final int FIRST_PRINTABLE_ASCII = 0x20;
+
+    private static boolean containsBlockedPasswordChar(String password) {
```
yagudin10, line 95: Not sure we need this kind of validation. Moreover, there are also non-printable characters above 0x20.

server/bundles/io.cloudbeaver.server/src/io/cloudbeaver/service/core/impl/WebServiceCore.java
```diff
@@ -84,6 +88,42 @@ public class WebServiceCore implements DBWServiceCore {
...
+        for (int i = 0; i < password.length(); i++) {
+            if (password.charAt(i) < FIRST_PRINTABLE_ASCII) {
+                return true;
+            }
+        }
+        return false;
+    }
+
+    private void emitPasswordChangeAudit(
```
yagudin10, line 104: Delete that method

server/bundles/io.cloudbeaver.server/src/io/cloudbeaver/service/core/impl/WebServiceCore.java
```diff
@@ -641,6 +681,203 @@ private DataSourceDescriptor getDataSourceDescriptor(
...
+            return;
+        }
+        emitPasswordChangeAudit(webSession, container, projectId, connectionId,
+            WSSecurityAuditEvent.Kind.FAILED, "PERSIST_FAILED", failureClass);
+        throw new DBWebException(
+            "Database password was changed but CloudBeaver failed to persist the new credential. "
+                + "The connection will require re-entry of the new password.");
+    }
+
+    private void safeDisconnect(@NotNull WebSession webSession, @NotNull DBPDataSourceContainer container) {
```
yagudin10: Use WebDataSourceUtils#disconnectDataSource

server/bundles/io.cloudbeaver.server/src/io/cloudbeaver/service/core/impl/WebServiceCore.java
```diff
@@ -641,6 +681,203 @@ private DataSourceDescriptor getDataSourceDescriptor(
...
+    ) throws DBWebException {
+        DBPDataSourceContainer container = null;
+        try {
+            requireServerFlagEnabled(webSession, projectId, connectionId);
+            container = resolveContainer(webSession, projectId, connectionId);
+            validateNewPassword(webSession, container, projectId, connectionId, newPassword);
+            ensureConnected(webSession, container, projectId, connectionId);
+            DBAUserPasswordManager manager = resolveUserPasswordManager(webSession, container, projectId, connectionId);
+            String userName = resolveUserName(webSession, container, projectId, connectionId);
+            applyPasswordChange(webSession, container, projectId, connectionId, manager, userName, oldPassword, newPassword);
```
yagudin10, line 700: Not sure we need to use an approach like this. Usually, we try to reuse code that already exists in the DBeaver repository. For example, in this case, it would be better to use DBAuthUtils#promptAndChangePasswordForCurrentUser with some adjustments.

server/bundles/io.cloudbeaver.service.admin/schema/service.admin.graphqls
```diff
@@ -205,6 +205,8 @@ input ServerConfigInput {
     publicCredentialsSaveEnabled: Boolean
     "Whether saving credentials is allowed"
     adminCredentialsSaveEnabled: Boolean
+    "Whether the change-DB-password mutation is enabled server-wide"
+    dbUserPasswordChangeEnabled: Boolean
```
yagudin10: Add `since` annotation

server/bundles/io.cloudbeaver.server.ce/src/io/cloudbeaver/server/CBServerConfigurationMapper.java
```diff
@@ -63,6 +63,7 @@ protected void populateConfigurations(@NotNull I input, @NotNull C serverConfig,
         appConfig.setSupportsCustomConnections(input.isCustomConnectionsEnabled());
         appConfig.setPublicCredentialsSaveEnabled(input.isPublicCredentialsSaveEnabled());
         appConfig.setAdminCredentialsSaveEnabled(input.isAdminCredentialsSaveEnabled());
+        appConfig.setDbUserPasswordChangeEnabled(input.isDbUserPasswordChangeEnabled());
```
yagudin10, line 66: Also we need to add the configuration parameter to `config/template/cloudbeaver-base.conf`

yagudin10 (changes requested):

server/bundles/io.cloudbeaver.server/src/io/cloudbeaver/service/core/impl/WebServiceCore.java
```diff
@@ -641,6 +681,203 @@ private DataSourceDescriptor getDataSourceDescriptor(
...
+            persistNewPassword(webSession, container, projectId, connectionId, newPassword);
+            safeDisconnect(webSession, container);
```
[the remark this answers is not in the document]
TobyTheHutt: Looked into this. Since it's a very unlikely scenario and requires a new container lock and, in my eyes, meaningfully extends the scope of this PR. If maintainers don't agree and want me to commit a fix within this PR, let me know. If desired, I'll propose a fix in a follow-up PR for this.

server/bundles/io.cloudbeaver.server/src/io/cloudbeaver/service/core/impl/WebServiceCore.java
```diff
@@ -641,6 +681,203 @@ private DataSourceDescriptor getDataSourceDescriptor(
         return testDataSource;
     }
 
+    @Override
+    public boolean changeConnectionUserPassword(
```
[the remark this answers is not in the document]
TobyTheHutt: Checked this. The The `persistSecrets` path requires a mock secret controller and I'd have to meaningfully extend the scope of this PR to implement this. The existing tests are already deterministic without a live DB and suffice in my opinion. Let me know if any of the maintainers disagree and want a fix in this PR.

server/bundles/io.cloudbeaver.server/src/io/cloudbeaver/server/events/WSSecurityAuditEvent.java
TobyTheHutt: The `WSSecurityAuditEventHandler` writes a structured log line and prints machine readable fields like `kind`, `reasonCode`, `connectionId`, `driverId` or `errorClass`, rather than just plaintext. It also adds the capability for future WebSocket subscriptions and ensures consistend vocabulary with other CB EventBus consumers.

`log.info` does not provide this, which is why I decided for this approach and continue to recommend it. Let me know if you concur of whether you want me to fall back to `log.info` anyway.

server/bundles/io.cloudbeaver.server/src/io/cloudbeaver/server/events/WSSecurityAuditEventHandler.java
TobyTheHutt: See https://github.com/dbeaver/cloudbeaver/pull/4586#discussion_r3862059741. Please confirm you really don't want it, before I just fall back.

server/bundles/io.cloudbeaver.server/src/io/cloudbeaver/service/core/impl/WebServiceCore.java
TobyTheHutt: This is a potential security concern because some DBs parse those control sequences. There are indeed other non-printable characters above these values, but they are not parsed the same way. Especially NUL (0x00) and characters that could cause JDBC driver string truncation or or SQL injection via control sequences are of concern here.

The risk is low and I can drop it. I just decided to add it because it's a low-cost hardening. Please confirm how I should proceed here.

server/bundles/io.cloudbeaver.server/src/io/cloudbeaver/service/core/impl/WebServiceCore.java
TobyTheHutt: Relates to https://github.com/dbeaver/cloudbeaver/pull/4586#discussion_r3862059741. This is the helper that dispatches `WSSecurityAuditEvent`. I'll act upon your comment depending on how the other discussions are resolved.

server/bundles/io.cloudbeaver.server/src/io/cloudbeaver/service/core/impl/WebServiceCore.java
TobyTheHutt: I'd have to add it to the classpath, which is feasible. Please confirm that this is what you meant by "some adjustments" and I'll be on it.

server/bundles/io.cloudbeaver.service.admin/schema/service.admin.graphqls
TobyTheHutt: Will fix with https://github.com/dbeaver/cloudbeaver/pull/4586#discussion_r3841608794.

TobyTheHutt:
@yagudin10, @devnaumov  I closed some conversations and added some comments to others, in which I need clarifications or a confirmation. Please check my actions and review my comments. I'll wait for your feedback and prepare the other code changes in the meantime.

TobyTheHutt:
@yagudin10 committed the feedback where I was confident about what change is required. Update me on the open comments and I'll do the rest.

server/bundles/io.cloudbeaver.server/src/io/cloudbeaver/server/events/WSSecurityAuditEventHandler.java
yagudin10: @TobyTheHutt Let's remove it.

server/bundles/io.cloudbeaver.server/src/io/cloudbeaver/service/core/impl/WebServiceCore.java
yagudin10: Is it possible to use these characters when using GraphQL API?

server/bundles/io.cloudbeaver.server/src/io/cloudbeaver/service/core/impl/WebServiceCore.java
yagudin10: @TobyTheHutt Let's remove it.

server/bundles/io.cloudbeaver.server/src/io/cloudbeaver/service/core/impl/WebServiceCore.java
yagudin10: Thought about it, and I think that your code it good rn. Changes in DBeaver repo could cause some bugs in the both products. But I think that we need to extract all the methods that use `audit` code.

server/bundles/io.cloudbeaver.server/src/io/cloudbeaver/service/core/impl/WebServiceCore.java
```diff
@@ -84,6 +88,42 @@ public class WebServiceCore implements DBWServiceCore {
 
     private static final Log log = Log.getLog(WebServiceCore.class);
 
+    private static final int MAX_PASSWORD_LENGTH = 128;
```
yagudin10, line 91: Why 128?
Let's remove this limit.

webapp/packages/plugin-administration/src/locales/en.ts
```diff
@@ -76,6 +76,8 @@ export default [
   ],
   ['administration_configuration_wizard_configuration_security_public_credentials', 'Save users credentials'],
   ['administration_configuration_wizard_configuration_security_public_credentials_description', 'Allows to save credentials for non-admin users'],
+  ['administration_configuration_wizard_configuration_security_db_user_password_change', 'Enable database user password change'],
+  ['administration_configuration_wizard_configuration_security_db_user_password_change_description', 'Users with edit rights on a connection can change the database user password via the connection context menu.'],
```
devnaumov, line 80: Please do not mention specific ui place and specific rights here.
something like The database user password can be changed for an existing connection

server/bundles/io.cloudbeaver.server/src/io/cloudbeaver/service/core/impl/WebServiceCore.java
devnaumov: @TobyTheHutt

devnaumov: @TobyTheHutt Could you please check the latest comments? We’re very close to merging it.

server/bundles/io.cloudbeaver.server/src/io/cloudbeaver/service/core/impl/WebServiceCore.java
TobyTheHutt: Just read the GraphQL spec and it restricts those characters natively. I'll remove it. Thanks for the hint!

webapp/packages/plugin-administration/src/locales/en.ts
TobyTheHutt: Done

server/bundles/io.cloudbeaver.server/src/io/cloudbeaver/service/core/impl/WebServiceCore.java
TobyTheHutt: Done.

server/bundles/io.cloudbeaver.server/src/io/cloudbeaver/service/core/impl/WebServiceCore.java
TobyTheHutt: Done.

server/bundles/io.cloudbeaver.server/src/io/cloudbeaver/server/events/WSSecurityAuditEventHandler.java
TobyTheHutt: Done.

server/bundles/io.cloudbeaver.server/src/io/cloudbeaver/server/events/WSSecurityAuditEvent.java
TobyTheHutt: Removed.

TobyTheHutt:
@yagudin10 @devnaumov 
Sorry for the late response. I've addressed the open threads and hope the implementation matches what you've expected. Let me know if you want any further changes.

webapp/packages/plugin-connections/src/locales/en.ts
```diff
@@ -47,4 +47,15 @@ export default [
   ['plugin_connections_connection_connection_time', 'Connection time: {arg:time} \n'],
 
   ['plugin_connections_connection_type_label', 'Connection type'],
+
+  ['plugin_connections_menu_security', 'Security'],
```
devnaumov: plugin_connections_menu_security is not in use anymore

devnaumov (dismissed):

yagudin10 (dismissed):

webapp/packages/plugin-connections/src/locales/en.ts
TobyTheHutt: Good catch, thanks!

yagudin10 (approved):

devnaumov (approved):

PR #4649, opened 2026-09-09
dbeaver/cloudbeaver#4648 Fix polling fallback after WebSocket closes

### LINKED ISSUE #4648: 26.2.0: Long-polling fallback does not start after a clean WebSocket close
### Description

CloudBeaver 26.2.0 introduced a regression to the long polling fallback behavior.

Previously, if the WebSocket either errored or closed cleanly, the app would fall back to long polling. In 26.2.0,  the fallback occurs after a ws error, but not when the ws is closed cleanly, even when the close was not initiated by an `unsubscribe` event and therefore an intentional client shutdown.

The observed request sequence is:
 - `/api/ws` upgrades successfully with HTTP 101
 -  The WebSocket closes roughly two seconds later.
  - No `/api/events` long-poll requests are made afterward.
  - Asynchronous task completion messages are never received.
  - Queries remain indefinitely on “Loading…”

This was observed when updating from the 26.1.4 to 26.2.0, with no other deployment configuration change.

### Steps to reproduce

A publicly accessible deployment reproducing this condition is not currently available. The issue can be reproduced whenever the WebSocket connection is accepted and subsequently closed cleanly by an intermediary:

  1. Run CloudBeaver 26.2.0 behind a reverse proxy or service mesh that accepts the WebSocket upgrade but then closes the connection cleanly.
  2. Open CloudBeaver and connect to a database.
  3. Open the SQL editor and execute a query such as `SELECT 1`.
  4. In the browser developer tools, observe `/api/ws` upgrade with status 101 and then close.
  5. Observe that CloudBeaver does not begin making `/api/events` requests.
  6. Observe that the query remains on “Loading…” indefinitely.

At the transport level, the same behavior can be reproduced by completing the WebSocket subject without first unsubscribing it. Long polling is not activated from the completion handler.

### Expected/Desired Behavior

When a WebSocket closes without an intentional client unsubscribe, CloudBeaver should switch to long polling and continue receiving asynchronous task events.

Expected flow request:

`101 WebSocket → clean close → GET /api/events → async-task completion → result grid`

### CloudBeaver Version

26.2.0

### Additional context

_No response_

ThomasSwinicki:
Closes #4648

## Summary

RxJS reports a clean WebSocket close through `complete`, bypassing the existing error-based fallback. `TransportSubject` therefore completed instead of switching to long polling.

This change:

  - Switches to long polling when the WebSocket completes.
  - Preserves intentional teardown behavior.
  - Adds regression tests for transport fallback and lifecycle behavior.

## Testing

  - Added focused `TransportSubject` unit tests.
  - Full frontend suite passed: 748 passed, 27 skipped.
  - Production frontend bundle completed successfully.

### DIFF
--- a/webapp/packages/core-root/src/TransportSubject.test.ts
+++ b/webapp/packages/core-root/src/TransportSubject.test.ts
@@ -7,7 +7,7 @@
  */
 
 import { Subject } from 'rxjs';
-import { webSocket, type WebSocketSubjectConfig } from 'rxjs/webSocket';
+import { webSocket } from 'rxjs/webSocket';
 import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest';
 
 import type { EnvironmentService } from '@cloudbeaver/core-sdk';
@@ -30,23 +30,14 @@ interface ITestEvent {
 describe('TransportSubject', () => {
   let websocketSubject: Subject<ITestEvent>;
   let pollingSubject: Subject<ITestEvent>;
-  let websocketConfig: WebSocketSubjectConfig<ITestEvent>;
-  let startPolling: () => void;
 
   beforeEach(() => {
     (globalThis as any)._ROOT_URI_ = '{ROOT_URI}';
     websocketSubject = new Subject();
     pollingSubject = new Subject();
 
-    vi.mocked(webSocket).mockImplementation(config => {
-      websocketConfig = config as WebSocketSubjectConfig<ITestEvent>;
-      return websocketSubject as never;
-    });
-
-    vi.mocked(longPolling).mockImplementation(options => {
-      startPolling = () => options.startObserver?.next();
-      return pollingSubject as never;
-    });
+    vi.mocked(webSocket).mockReturnValue(websocketSubject as never);
+    vi.mocked(longPolling).mockReturnValue(pollingSubject as never);
 
     vi.spyOn(console, 'warn').mockImplementation(() => undefined);
   });
--- a/webapp/packages/core-root/src/TransportSubject.test.ts
+++ b/webapp/packages/core-root/src/TransportSubject.test.ts
@@ -23,14 +23,14 @@ vi.mock('./longPolling.js', () => ({
   longPolling: vi.fn(),
 }));
 
-interface TestEvent {
+interface ITestEvent {
   id: string;
 }
 
 describe('TransportSubject', () => {
-  let websocketSubject: Subject<TestEvent>;
-  let pollingSubject: Subject<TestEvent>;
-  let websocketConfig: WebSocketSubjectConfig<TestEvent>;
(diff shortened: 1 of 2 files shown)


webapp/packages/core-root/src/TransportSubject.ts
```diff
@@ -32,7 +32,7 @@ export class TransportSubject<T> extends Subject<T> {
     this.output = new Subject();
     this.ready = new Subject();
 
-    this.ready$ = this.ready.pipe(take(1), shareReplay(1));
+    this.ready$ = this.ready.pipe(shareReplay(1));
```
devnaumov: @ThomasSwinicki  First of all, thank you for the contribution!

One thing needs to be fixed, I think we should keep take(1) here. ready$ currently represents the initial transport readiness, and its only consumer calls SessionEventSource.onInit.execute(). Removing take(1) would run onInit again when polling starts and on subsequent polling restarts, causing unnecessary reinitialization and resending of the active-project state.

The fallback itself does not require this change because the server-side event filters are session-scoped and survive a transport switch. The clean-close regression can be fixed independently by switching to polling from the WebSocket complete handler.

I suggest restoring take(1) and adjusting or removing the test that expects a readiness event for every activated transport. Keeping this change focused on the complete handler also minimizes the risk of behavioral regressions.

webapp/packages/core-root/src/TransportSubject.test.ts
```diff
@@ -0,0 +1,139 @@
...
+
+vi.mock('rxjs/webSocket', () => ({
+  webSocket: vi.fn(),
+}));
+
+vi.mock('./longPolling.js', () => ({
+  longPolling: vi.fn(),
+}));
+
+interface TestEvent {
```
devnaumov: Please add I prefix to this interface

devnaumov (changes requested):

webapp/packages/core-root/src/TransportSubject.ts
ThomasSwinicki: No problem! Thanks for catching that. I've restored `take(1)` and removed the test expecting another readiness notification when polling starts.

webapp/packages/core-root/src/TransportSubject.test.ts
```diff
@@ -0,0 +1,139 @@
...
+      startPolling = () => options.startObserver?.next();
+      return pollingSubject as never;
+    });
+
+    vi.spyOn(console, 'warn').mockImplementation(() => undefined);
+  });
+
+  afterEach(() => {
+    vi.restoreAllMocks();
+  });
```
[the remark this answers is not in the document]
ThomasSwinicki: Included in feedback commit.

devnaumov (approved):

sergeyteleshev (approved):

PR #4657, opened 2026-09-14
dbeaver/pro#10012 delegate suspense to tab panels

devnaumov:
closes [10012](https://github.com/dbeaver/pro/issues/10012)

Suspense was replaced with controlled loading={!formState} because the issue is related to asynchronous form-state initialization rather than component loading.

Previously, PublicConnectionForm could be loaded while formState was still null. Since no child component was rendered or suspended at that point, the Suspense fallback was not triggered, leaving an empty area.

ConnectionForm is imported directly because PublicConnectionForm is already lazy-loaded at a higher level;

Now:
loading explicitly represents form-state readiness;
the loader remains visible until formState is created;
lazy-loaded tabs retain their own Suspense boundaries inside TabPanel.


webapp/packages/plugin-connections/src/PublicConnectionForm/PublicConnectionForm.tsx
```diff
@@ -30,8 +31,8 @@ export const PublicConnectionForm: React.FC = observer(function PublicConnection
 
   return (
     <ColoredContainer className={s(style, { loader: true })}>
-      <Loader className={s(style, { loader: true })} suspense>
-        {service.formState && <ConnectionFormLoader formState={service.formState} onSave={save} onCancel={close} />}
+      <Loader className={s(style, { loader: true })} loading={!formState}>
```
sergeyteleshev, line 34: there is a case if something went wrong during form state creation and it would throw an error. in that case loading will be infinite

we need a determined way to get loading state

sergeyteleshev (approved):

PR #4673, opened 2026-09-18
dbeaver/pro#10736 add multisource presentation

devnaumov:
closes [10736](https://github.com/dbeaver/pro/issues/10736)


webapp/packages/core-connections/src/ConnectionInfoResource.ts
```diff
@@ -408,6 +408,10 @@ export class ConnectionInfoResource extends CachedMapResource<IConnectionInfoPar
       config,
     });
 
+    if (config.hidden) {
+      return connection;
+    }
```
sergeyteleshev: why dont we store hidden connection in the resource? i think we should store it but dont show in UI if it is hidden. by doing so we can operate the data the way we want and its better then not having it at all

webapp/packages/core-connections/src/ConnectionInfoResource.ts
```diff
@@ -507,16 +511,21 @@ export class ConnectionInfoResource extends CachedMapResource<IConnectionInfoPar
   }
 
   async update(key: IConnectionInfoParams, config: ConnectionConfig): Promise<DatabaseConnection> {
-    await this.performUpdate(key, [], async () => {
+    const connection = await this.performUpdate(key, [], async () => {
```
sergeyteleshev: same here

PR #4680, opened 2026-09-21
dbeaver/pro#10717 Add combobox groups and pinned footer items

sergeyteleshev:
Closes dbeaver/pro#10717

## Summary
- Add Ariakit-backed combobox groups and separator boundaries, following the Select pattern.
- Add pinned footer items that remain visible while filtering and scrolling.
- Keep selection handling shared between regular and footer items.

## Validation
- Core blocks and UI-kit TypeScript build passed.
- Core blocks ESLint, changed-file Prettier, and diff checks passed.
- The core-blocks suite passed before the newly added test file was removed at the requester's request (152 passed, 17 skipped). No new test file is included.
- UI-kit lint reports two existing interface-naming violations.
- Browser visual and screen-reader verification pending.


webapp/packages/core-blocks/src/FormControls/Combobox.tsx
```diff
@@ -178,7 +194,7 @@ export const Combobox: IComboboxType = observer(function Combobox({
   const comboboxStore = useComboboxStore({
     value: displayValue,
     setValue: setInputValue,
-    selectedValue: selectedValue,
+    selectedValue: String(selectedKey ?? ''),
```
sergeyteleshev, line 193: we need it because network profiles has collisions in naming. for instance global and project profile can have the same name 'new-profile'. in this case we will select profiles wrongly. thats why we need to store unique keys instead of visible values

webapp/packages/core-blocks/src/FormControls/Combobox.tsx
```diff
@@ -75,6 +80,7 @@ export const Combobox: IComboboxType = observer(function Combobox({
   name,
   state,
   items,
+  footerItems,
```
sergeyteleshev, line 83: I have supported the behavior we adopted for Select. so we also have dividers and visually all looks the same as it was for profile selector

devnaumov (approved):

